PR #3, opened 2017-10-27
whitelist stinger transitions

andycreeth:
It seems like transition point doesn't work, but it doesn't seem to work in latest OBS either.


derrod:
Stinger transitions work fine in latest obs master.

andycreeth:
@derrod Is that released?  I might just be misunderstanding what the transition point is supposed to be doing.

andycreeth:
I also noticed we don't have crossfade or audio monitoring, so I think we're a fair bit behind OBS master for this.  I will look into that.

andycreeth:
Yeah I totally wasn't understanding how the transition point works.  It does seem to be working properly.  We just need to update our obs-studio build so we can get the crossfade and monitoring.

PR #6, opened 2017-10-27
Remove audio monitoring from the blacklist

eddyStreamlabs:


andycreeth:
Cool will merge this when new node-obs goes in.

PR #11, opened 2017-10-30
rename source

holiber:


app/components/windows/NameSource.vue
```diff
@@ -1,6 +1,6 @@
 <template>
 <modal-layout
-  title="Name Source"
+  :title="options.rename ? 'Reame Source' : 'NameSource'"
```
andycreeth: - Should be "Rename" instead of "Reame"
- Add a space between "Name" and "Source"

holiber: thanks

PR #13, opened 2017-10-30
use app urls for module resolution, move sources into directory

andycreeth:
This PR started because I wanted to move the sources service from a single file to a directory.  I realized that our use of relative paths for everything makes this very annoying.  So I changed our webpack and tsconfig to allow us to load files relative to the `app` directory instead.  This means we can move files around in the directory structure without breaking imports.  We should do this from now on instead of using relative paths.

There are a number of ways to achieve this effect.  The way I have it set up here seems to preserve all intellisense, click-to-definition etc in VSCode.  @holiber Can you please check in your IDE to make sure this change doesn't impact your workflow?

### DIFF
--- a/app/services/sources/index.ts
+++ b/app/services/sources/index.ts
@@ -6,16 +6,16 @@ import {
   getPropertiesFormData,
   setPropertiesFormData,
   setupSourceDefaults
-} from '../components/shared/forms/Input';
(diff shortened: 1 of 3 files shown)


andycreeth:
Will resolve merge conflicts.

PR #14, opened 2017-10-30
Reset obs context

eddyStreamlabs:
Reset the audio and video context after importing from obs to applied the imported settings.

### DIFF
--- a/app/services/obs-importer.ts
+++ b/app/services/obs-importer.ts
@@ -233,7 +233,6 @@ export class ObsImporterService extends Service {
         const appData = electron.remote.app.getPath('userData');
         const currentFilePath = path.join(appData, file);
 
-        fs.createReadStream(obsFilePath).pipe(fs.createWriteStream(currentFilePath));
         const readData = fs.readFileSync(obsFilePath);
         fs.writeFileSync(currentFilePath, readData);
       }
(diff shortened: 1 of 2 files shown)


app/services/obs-importer.ts
```diff
@@ -230,6 +234,8 @@ export class ObsImporterService extends Service {
         const currentFilePath = path.join(appData, file);
 
         fs.createReadStream(obsFilePath).pipe(fs.createWriteStream(currentFilePath));
```
andycreeth: You should remove the old async code.  I'm actually surprised this isn't causing any errors.

app/services/scenes-transitions.ts
```diff
@@ -89,16 +89,22 @@ export class ScenesTransitionsService extends StatefulService<ISceneTransitionsS
 
   setType(type: string) {
     const oldTransition = this.getCurrentTransition() as obs.ITransition;
-    const newTransition = obs.TransitionFactory.create(type, 'Global Transition');
 
-    obs.Global.setOutputSource(0, newTransition);
+    const transition = TRANSITION_TYPES.find(transition => {
```
andycreeth: What's up with this diff.  I thought we merged this earlier.

PR #18, opened 2017-10-31
Add Properties Manager

andycreeth:


app/services/sources/properties-managers/properties-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+  setPropertyFormData(property: TCustomProperty |
+    input.IFormInput<input.TObsValue> |
+    input.IListInput<input.TObsValue>): void;
+  settings: Dictionary<any>;
+}
+
+
+export interface ICustomProperty<TValue> extends input.IFormInput<TValue> {
+  isCustom: true;
+}
```
holiber: Are you sure that here can be something different than `TObsValue` ?

andycreeth: 👍 I restricted `TValue` so it must be a `TObsValue`

PR #33, opened 2017-11-06
Move to backtrace.io

computerquip-streamlabs:
Backtrace has a better interface and more feature (such as better analytics, explicit support for custom attributes, and various others). It also already includes Windows and Electron symbol servers and we can add our own as well. 

Note that yarn.lock has been updated to reflect the bugsplat packaged removed and the backtrace-node package added.

### DIFF
--- a/app/app.ts
+++ b/app/app.ts
@@ -19,17 +19,30 @@ const { ipcRenderer, remote } = electron;
 const slobsVersion = remote.process.env.SLOBS_VERSION;
 
 if (remote.process.env.NODE_ENV === 'production') {
-  const bugsplat = require('bugsplat')('slobs', 'slobs-renderer', slobsVersion);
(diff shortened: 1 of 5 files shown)


app/app.ts
```diff
@@ -19,17 +19,30 @@ const { ipcRenderer, remote } = electron;
 const slobsVersion = remote.process.env.SLOBS_VERSION;
 
 if (remote.process.env.NODE_ENV === 'production') {
-  const bugsplat = require('bugsplat')('slobs', 'slobs-renderer', slobsVersion);
-  window.onerror = (messageOrEvent, source, lineno, colno, error) => bugsplat.post(error);
+  const bt = require('backtrace-node');
```
andycreeth: This is being compiled into our webpack bundle.  It's probably fine, but for something like this it's probably best to just require it dynamically at runtime.  To do that, you will need to add it to the list of externals in our webpack config.

PR #34, opened 2017-11-07
Loading config fail case

eddyStreamlabs:
- Create / save a backup file after each successful load
 - Handle sync and async exceptions during the loading
 - Load the backup file if an error occurred during the loading
 - If no backup file exists yet, load the default configuration

### DIFF
--- a/app/services/config-persistence/config.ts
+++ b/app/services/config-persistence/config.ts
@@ -143,9 +143,8 @@ export class ConfigPersistenceService extends PersistentStatefulService<IScenesC
     return new Promise(resolve => {
       const configName = this.state.activeCollection;
 
-      this.scenesService.scenes.forEach(scene => {
-        scene.remove();
-      });
+      this.scenesService.scenes.forEach(scene => scene.remove(true));
+      this.sourcesService.sources.forEach(source => { if (source.type !== 'scene') source.remove(); });
 
       const backConfigFile = this.getConfigFilePath(configName) + '.bak';
 
--- a/app/services/config-persistence/config.ts
+++ b/app/services/config-persistence/config.ts
@@ -65,7 +65,7 @@ export class ConfigPersistenceService extends PersistentStatefulService<IScenesC
     const configsNamesWithoutBackup: string[] = [];
 
     if (configsNames.length) {
-      configsNames.forEach (configName => {
+      configsNames.forEach(configName => {
         if (configName.indexOf('_backup') === -1) configsNamesWithoutBackup.push(configName);
       });
       this.ADD_SCENES_COLLECTIONS(configsNamesWithoutBackup);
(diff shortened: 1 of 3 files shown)


app/services/config-persistence/config.ts
```diff
@@ -62,8 +62,13 @@ export class ConfigPersistenceService extends PersistentStatefulService<IScenesC
     if (!fs.existsSync(this.configFileDirectory)) return;
 
     const configsNames = fs.readdirSync(this.configFileDirectory).map(file => file.replace(/\.[^/.]+$/, ''));
+    const configsNamesWithoutBackup: string[] = [];
+
     if (configsNames.length) {
-      this.ADD_SCENES_COLLECTIONS(configsNames);
+      configsNames.forEach (configName => {
```
andycreeth: No space after `forEach`

app/services/config-persistence/config.ts
```diff
@@ -62,8 +62,13 @@ export class ConfigPersistenceService extends PersistentStatefulService<IScenesC
...
+        if (configName.indexOf('_backup') === -1) configsNamesWithoutBackup.push(configName);
```
andycreeth: I think we should do what OBS does and use a `.json.bak` file extension for backup files.  The way you have this written, any scene collection with `_backup` in the name will be treated as a backup, even if that's just what the user named it.

app/services/config-persistence/config.ts
eddyStreamlabs: Done.

app/services/config-persistence/config.ts
eddyStreamlabs: Done.

app/services/config-persistence/config.ts
```diff
@@ -61,16 +61,11 @@ export class ConfigPersistenceService extends PersistentStatefulService<IScenesC
...
-    const configsNamesWithoutBackup: string[] = [];
-
-    if (configsNames.length) {
-      configsNames.forEach(configName => {
-        if (configName.indexOf('_backup') === -1) configsNamesWithoutBackup.push(configName);
-      });
-      this.ADD_SCENES_COLLECTIONS(configsNamesWithoutBackup);
-    }
+    const configsNames = fs.readdirSync(this.configFileDirectory).filter(fileName => {
+      return fileName.indexOf('.bak') === -1;
```
andycreeth: You should use a regex to make sure the `.bak` is at the end of the string:
```TypeScript
fileName.match(/\.bak$/)
```
This will return an array of match data if it matches, or `null` if it doesn't match.

eddyStreamlabs: Done.

app/services/config-persistence/nodes/sources.ts
```diff
@@ -142,7 +142,7 @@ export class SourcesNode extends Node<ISchema, {}> {
         }
       );
       if (source.audioMixers) {
-        this.audioService.getSource(sourceInfo.id).setMul(sourceInfo.volume);
+        this.audioService.getSource(sourceInfo.id).setMul(sourceInfo.volume || 1);
```
andycreeth: I think this is a destructive change.  If people have their volume set to `0`, it will revert back to full volume.

PR #35, opened 2017-11-07
Blackmagic input source

eddyStreamlabs:
- Whitelist decklink-input type
 - Only get the available white listed type of sources and filters
 - Add description and support list for decklink-input source

### DIFF
--- a/app/services/utils.ts
+++ b/app/services/utils.ts
@@ -80,18 +80,4 @@ export default class Utils {
     }
     return result;
   }
-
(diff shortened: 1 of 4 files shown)


app/services/utils.ts
```diff
@@ -80,4 +80,18 @@ export default class Utils {
     }
     return result;
   }
+
+  static getAvailableWhitelistedType  (obsAvailableTypes: string[], whitelistedTypes:  any[]):  any[] {
```
andycreeth: I don't think you really need to define a function here.  You can replicate this same behavior using this line of code:
```TypeScript
whitelistedTypes.filter(type => obsAvailableTypes.includes(type.value));
```

PR #36, opened 2017-11-08
Implement Streamlabels (Except trains)

andycreeth:
This PR implements streamlabels in the app.

Here are some details:
- Initial fetch of data is done via HTTP request
- Subsequent updates are handled via websocket events
- Custom properties on properties managers have been ditched in favor of custom UI components to handle properties.  This is because OBS property types were too restrictive to implement what we wanted for streamlabels.


app/services/streamlabels/definitions.ts
```diff
@@ -0,0 +1,861 @@
+import { TPlatform } from 'services/platforms';
```
andycreeth: Most of this file is copied from the existing streamlabels app.  I munged the data a bit, removed mixer definitions for now, and defined some interfaces based on the structure of the data.

app/components/windows/AddSource.vue.ts
```diff
@@ -65,11 +67,16 @@ export default class AddSource extends Vue {
...
       this.error = 'That name is already taken';
     } else {
-      const sceneItem = this.scenesService.activeScene.createAndAddSource(
+      const source = this.sourcesService.createSource(
         this.name,
-        this.sourceType
+        this.sourceType,
+        {},
+        {
+          propertiesManager: this.propertiesManager ? this.propertiesManager : void 0
```
holiber: Do we really need explicitly set which property manager we have to use? What will happen if I try to use `Default` property manager for `widget` or `streamlabels` ?

app/services/streamlabels/index.ts
```diff
@@ -0,0 +1,212 @@
...
+        if (e.type === 'streamlabels') {
+          this.setStreamlabelsData(e.message.data);
+        }
+      });
+    });
+  }
+
+
+  log(message: string, ...args: any[]) {
+    console.log(`Streamlabels: ${message}`, ...args);
```
holiber: do we need logs always enabled?

app/services/streamlabels/index.ts
```diff
@@ -0,0 +1,212 @@
...
+      .then(handleErrors)
+      .then(() => true);
+  }
+
+
+  /**
+   * Attempt to load initial data via HTTP instead of waiting
+   * for a socket event
+   */
+  fetchInitialData() {
```
holiber: can you set returning type for this function?

app/services/streamlabels/index.ts
```diff
@@ -0,0 +1,212 @@
...
+    return path.join(this.streamlabelsDirectory, `${filename}.txt`);
+  }
+
+
+  /**
+   * Updates the data in a single streamlabels file
+   * @param filename the filename
+   * @param statname the name of the stat
+   */
+  updateSubscription(subscription: IStreamlabelsSubscription) {
```
holiber: I don't see params above as arguments for this metod

app/services/streamlabels/index.ts
```diff
@@ -0,0 +1,212 @@
...
+    if (!this.data) return;
+
+    console.log('Writing file', subscription.filename, subscription.statname);
```
holiber: should be `this.log()`

app/services/streamlabels/index.ts
```diff
@@ -0,0 +1,212 @@
...
+
+    electron.ipcRenderer.send('streamlabels-writeFile', {
+      path: this.getStreamlabelsPath(subscription.filename),
+      data: this.data[subscription.statname]
+    });
```
holiber: Why we need main process here?

app/services/streamlabels/index.ts
andycreeth: Yeah maybe I should add an ENV var

app/components/windows/AddSource.vue.ts
andycreeth: Yes.  The properties manager determines whether it is streamlabels or not.  For example:
- A "normal" text source: `text_gdiplus` source with `default` manager
- A "Stream Label": `text_gdiplus` source with `streamlabel` manager
- A "normal" browser source: `browser_source` with `default` manager
- A "Widget": `browser_source` with `widget` manager

app/services/streamlabels/index.ts
andycreeth: Oops, I changed the method and forgot to edit the comment

app/services/streamlabels/index.ts
andycreeth: Actually meant to remove this one entirely.

app/services/streamlabels/index.ts
andycreeth: In theory, we shouldn't need the main process at all.  Unfortunately there are a lot of issues doing asynchronous writes from the renderer process.  Synchronous writes work just fine (we do those for config saving and a few other things).  However, asynchronous writes will fail a good portion of the time.  Given the high volume of writes we need to do for streamlabels, I wasn't comfortable doing them synchronously.  Asynchronous writes work fine from the main process, so I am using an asynchronous IPC to do them from there.

As best I can tell, the issue seems to be a bug in electron.  I lost many hours to investigation and it turns out that Karl hit the exact same issue when writing the original streamlabels app (also electron).  When we have time we are going to try to reproduce in a test case and file a bug with electron.

holiber: Ouch

PR #38, opened 2017-11-08
expose API

holiber:


andycreeth:
Hey, I'm going to try to find some time soon to give this a more thorough look over.  Overall I think this is a good direction.

We should talk over as a team whether we want all of these settings exposed and what the defaults should be.  My guess is that we probably don't want to allow listening on anything but the 127.0.0.1 loopback address, mainly to prevent security issues for users who don't know any better.  We may also just want to keep named pipes permanently listening, and then tell streamdeck that they should use named pipes.  I'm envisioning people messing with settings they don't understand and then being confused why their streamdeck suddenly doesn't work anymore.

holiber:
We can show some advanced settings only if application has been started with some additional parameters.

PR #40, opened 2017-11-08
Bump node obs 0129

eddyStreamlabs:


andycreeth:
One of the appveyor builds got stuck.  Doing a rebuild just to make sure this is good.

PR #44, opened 2017-11-13
Streamlabels trains/combos

andycreeth:
Trains are fully client side, so this was a little tough to implement.  Unfortunately we inherited a lot of technical debt from the existing stream labels system.

I think the entire stream labels system needs to be moved entirely client-side and work off a data driven API.  Unfortunately this requires a lot of backend changes and we don't have the resources to tackle that right now.

### DIFF
--- a/app/services/sources/properties-managers/streamlabels-manager.ts
+++ b/app/services/sources/properties-managers/streamlabels-manager.ts
@@ -1,6 +1,6 @@
 import { PropertiesManager } from './properties-manager';
 import { Inject } from 'util/injector';
-import { StreamlabelsService } from 'services/streamlabels';
+import { StreamlabelsService, IStreamlabelSubscription } from 'services/streamlabels';
 import { getDefinitions } from 'services/streamlabels/definitions';
 import { UserService } from 'services/user';
 
@@ -14,7 +14,7 @@ export class StreamlabelsManager extends PropertiesManager {
   @Inject() userService: UserService;
 
   settings: IStreamlabelsManagerSettings;
-  filename: string;
+  subscription: IStreamlabelSubscription;
   blacklist = ['read_from_file', 'file'];
   customUIComponent = 'StreamlabelProperties';
 
@@ -37,8 +37,8 @@ export class StreamlabelsManager extends PropertiesManager {
 
 
   unsubscribe() {
-    if (this.filename) {
-      this.streamlabelsService.unsubscribe(this.filename);
+    if (this.subscription) {
+      this.streamlabelsService.unsubscribe(this.subscription);
     }
   }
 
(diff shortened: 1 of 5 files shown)


app/services/sources/properties-managers/streamlabels-manager.ts
```diff
@@ -14,7 +14,7 @@ export class StreamlabelsManager extends PropertiesManager {
   @Inject() userService: UserService;
 
   settings: IStreamlabelsManagerSettings;
-  filename: string;
+  subscription: IStreamlabelSubscription;
```
holiber: For easier refactoring it's better to use `private` modifier by default for each parameter that we are not going to use in vue components and outside of this service. When you will try to remove or rename private property, IDE will try to find usages only in current file instead of whole project.

app/services/streamlabels/definitions.ts
```diff
@@ -16,10 +16,11 @@ export interface IStreamlabelDefinition {
 }
 
 export interface IStreamlabelSettingsDefinition {
-  format: { tokens: string[]; };
+  format?: { tokens: string[]; };
   item_format?: { tokens: string[]; };
   item_separator?: { tokens: string[]; };
-  includeResubsOption?: boolean;
+  settingsStat?: string;
+  settingsWhitelist?: string[];
```
holiber: Wow, the all fields are optional?

app/services/streamlabels/definitions.ts
```diff
@@ -555,26 +548,109 @@ const youtubeDefinitions: IStreamlabelSet = {
 const twitchDefinitions: IStreamlabelSet = {
   // Trains are currently disabled
```
holiber: Looks like this comment is not actual now

app/services/streamlabels/index.ts
```diff
@@ -137,40 +277,168 @@ export class StreamlabelsService extends Service {
...
+  private outputAllTrains() {
+    Object.keys(this.trains).forEach(train => this.outputTrainInfo(train));
+  }
+
+
+  /**
+   * Outputs all files on a train, except for the clock while the
+   * train is running.
+   */
+  private outputTrainInfo(trainType: string) {
```
holiber: Can we define a type here? Looks like should be something like
```
type TStreamLabelsEventType = 'donation' | 'subscription' etc..
private outputTrainInfo(trainType: TStreamLabelsEventType ) { ...
```

app/services/streamlabels/definitions.ts
andycreeth: Yeah.  It very much depends on the type of streamlabel is.

app/services/streamlabels/definitions.ts
andycreeth: Oops, will remove

PR #47, opened 2017-11-15
styling for the filters window

mbiemiller:
- cleaned up the UI for the filters window


andycreeth:
Oops I didn't see this in time to get it in the 0.7.0 release but we can get it in the next one.

PR #57, opened 2017-11-17
Revive the overlay system

andycreeth:
In this PR:
- Expose overlay import/export in a new tab of settings
- Move overlay save/load to use async/await for code tidiness
- Allow importing of an overlay into a new scene collection named after the overlay file
- Overlays now support exporting/importing streamlabels
- Overlays now support exporting/importing widgets

### DIFF
--- a/app/services/config-persistence/config.ts
+++ b/app/services/config-persistence/config.ts
@@ -236,12 +236,17 @@ export class ConfigPersistenceService extends PersistentStatefulService<IScenesC
 
 
   switchToBlankConfig(configName = DEFAULT_SCENES_COLLECTION_NAME) {
+    this.switchToBlankConfig(configName);
+    this.setUpDefaults();
+  }
+
+
+  switchToEmptyConfig(configName: string) {
     if (this.scenesService.state.scenes.length) {
       throw 'unable to switch to blank config while current config is loaded';
     }
     this.SET_ACTIVE_COLLECTION(configName);
     this.ADD_SCENES_COLLECTIONS([configName]);
-    this.setUpDefaults();
   }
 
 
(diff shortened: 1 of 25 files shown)


app/services/config-persistence/config.ts
```diff
@@ -236,12 +236,17 @@ export class ConfigPersistenceService extends PersistentStatefulService<IScenesC
 
 
   switchToBlankConfig(configName = DEFAULT_SCENES_COLLECTION_NAME) {
+    this.switchToBlankConfig(configName);
```
holiber: Infinity recursion here. Looks like in some cases app will not start

app/services/config-persistence/nodes/overlays/video.ts
```diff
@@ -15,18 +15,26 @@ interface IContext {
...
   schemaVersion = 1;
 
-
-  save(context: IContext) {
+  async save(context: IContext) {
     const filePath = context.sceneItem.getObsInput().settings['local_file'];
+
+    // Don't bother saving media sources that don't have
+    // a file currently loaded
+    if (!filePath) return;
```
holiber: What if I want to save filters for the source and add the file a little bit later?

app/services/config-persistence/config.ts
andycreeth: Yikes, meant to call the new method I created.  Good catch!

app/services/config-persistence/nodes/overlays/video.ts
andycreeth: We don't support that.  The overlay system is meant for exporting finished overlays.  Anything that won't import in a good way for an end user, like a media source without a file, will be ignored.  The main people exporting these files will be our designers.

PR #61, opened 2017-11-20
Exclude unnecessary files in OSN

computerquip-streamlabs:
I think that's everything. Only distribute and the root files are necessary. If I missed anything, let me know.


andycreeth:
Awesome.

PR #65, opened 2017-11-21
replace obs source name with id

holiber:


andycreeth:
It looks like you have a test failure.

PR #69, opened 2017-11-21
fix overlay widget importing

andycreeth:
We weren't importing saved settings on the browser source.  I already packaged up a version on this branch for Eric to distribute to designers, so we just need to make sure this hits staging before next release.

### DIFF
--- a/app/services/scenes-collections/nodes/overlays/widget.ts
+++ b/app/services/scenes-collections/nodes/overlays/widget.ts
@@ -34,6 +34,9 @@ export class WidgetNode extends Node<ISchema, IContext> {
     context.sceneItem.source.replacePropertiesManager('widget', {
       widgetType: this.data.type
     });
+
+    // Make sure we don't override the url setting
+    delete this.data.settings['url'];
     context.sceneItem.getObsInput().update(this.data.settings);
   }
 }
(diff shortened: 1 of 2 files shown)


app/services/scenes-collections/nodes/overlays/widget.ts
```diff
@@ -34,5 +34,9 @@ export class WidgetNode extends Node<ISchema, IContext> {
     context.sceneItem.source.replacePropertiesManager('widget', {
       widgetType: this.data.type
     });
+
+    // Make sure we don't override the url setting
+    delete this.data.settings['url'];
+    context.sceneItem.getObsInput().update(this.data.settings);
```
holiber: I would prefer to use our API methods for that instead `getObsInput().update`

```
context.sceneItem.getSource().updateSettings(this.data.settings);
```
Because it's possible that for example that we will need to generate some event like `onSettingsUpdated`.

andycreeth: Sounds good.  I copied this from another overlay node, and I think the `updateSettings` function didn't exist back then.

andycreeth: We use `getObsInput` in a lot of places, but eventually we may want to clean all of those up and make it a private function.

holiber: I'm not sure that is possible to make it a private source while we `obs-studio-node` methods that requires ObsInput as argument. For example the volmeter:
```
obsVolmeter.attach(source.getObsInput());
```

it will be possible if `obs-studio-node` will use for example name or id:
```
obsVolmeter.attach(source.id);
```

andycreeth: Ah, makes sense.

PR #70, opened 2017-11-22
Don't recreate properties to check if properties are valid

computerquip-streamlabs:


holiber:
Looks like something wrong with filters

computerquip-streamlabs:
I'm going to say it's fine after playing with it a bit. I'm still unsure as to why the previous test failed which worries me but I see no way that's possible since configurable returns a boolean unconditionally (no chance of returning something other than true or false).

PR #78, opened 2017-11-28
Handle path filters in path properties

computerquip-streamlabs:


computerquip-streamlabs:
I noticed the types concerning the different input value classes isn't actually being checked. I'm going to fix that and then resubmit.

PR #80, opened 2017-11-28
Add path filters and some type safety improvements

computerquip-streamlabs:
Path properties weren't taking advantage of file filters since the functionality was missing from obs-studio-node for a bit. In addition, this uses object spread functionality to take advantage of type information and changes the type information require fields that should always be present. Read individual commit messages for more information.

### DIFF
--- a/app/components/shared/forms/Input.ts
+++ b/app/components/shared/forms/Input.ts
@@ -63,12 +63,14 @@ export interface IListOption<TValue> {
 }
 
 export interface IPathInputValue extends IFormInput<string> {
+  defaultPath: string;
   filters: IElectronOpenDialogFilter[];
 }
 
 export interface INumberInputValue extends IFormInput<number> {
-  minVal?: number;
-  maxVal?: number;
+  minVal: number;
+  maxVal: number;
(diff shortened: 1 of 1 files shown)


app/components/shared/forms/Input.ts
```diff
@@ -349,30 +356,41 @@ export function getPropertiesFormData(obsSource: obs.ISource): TFormData {
 
     if (isNumberProperty(obsProp)) {
       if (obsProp.details.type === obs.ENumberType.Slider) {
-        Object.assign(formItem as ISliderInputValue, {
+        (formItem as INumberInputValue) = {
```
andycreeth: Hmm, it looks like you're re-assigning a `const` declaration, which should result in an exception.  Is this working as expected?  Maybe I'm missing something.

computerquip-streamlabs: Herm good point. Seems making formItem non-const also has typescript complain that it's never re-assigned though which is odd.

computerquip-streamlabs: And yes, it appears to work correctly.

PR #83, opened 2017-11-29
Add interactive deploy script, switch to sentry.io for renderer exceptions

andycreeth:


holiber:
In `service-manager.ts` we have a function `executeServiceRequest` that wraps all JSONRPC requests in `try catch` and shows the error via `console.log`. Looks like we have manually send error callstack to Raven in that place.

holiber:
From other side any incorrect usage of API will send errors to Raven

holiber:
Yes. We can, but looks like we have to re-raise it only if it was a request from the child window. We can know that request is from child window if parameter `fetchMutations` exists in request body.

PR #87, opened 2017-11-29
Use backtrace in production only

computerquip-streamlabs:


andycreeth:
That failure is expected and is fixed on staging now.

PR #90, opened 2017-11-30
avoid shutdown errors

andycreeth:


andycreeth:
Excuse the large diff, I ran prettier formatter on the file.

app/services/app/app.ts
```diff
@@ -118,24 +98,30 @@ export class AppService extends StatefulService<IAppState> implements IAppServic
...
   }
 
   /**
    * reset current scene collection and load new one
    */
-  loadConfig(configName?: string, options = { saveCurrent: true }): Promise<void> {
+  loadConfig(
+    configName?: string,
+    options = { saveCurrent: true }
+  ): Promise<void> {
```
holiber: since we started to use `async/await`  code of this function can be simplified

andycreeth:
Hmm that's 2 different test intermittent test failures on app restart tests that have historically been very stable.  I'm going to do some more investigation before I consider this safe to merge.

PR #91, opened 2017-11-30
send user context to sentry

andycreeth:


app/services/app/app.ts
```diff
@@ -56,48 +56,46 @@ export class AppService extends StatefulService<IAppState>
 
   @track('app_start')
   load() {
-    // This is synchronous and can take a really long time for large configs.
-    // Setting a timeout allows the spinner and loading text to be drawn to
-    // the screen before starting on the slow synchronous operation.
-    // TODO: loading should be async
-    setTimeout(() => {
```
andycreeth: This `setTimeout` was completely unnecessary and I think was a relic of refactoring.  The actually useful `setTimeout` is in the `loadConfig` function.

app/app.ts
```diff
@@ -19,10 +19,32 @@ import RavenVue from 'raven-js/plugins/vue';
...
+      'format=minidump&' +
+      'token=e3f92ff3be69381afe2718f94c56da4644567935cc52dec601cf82b3f52a06ce',
+    extra: {
+      version: slobsVersion,
+      processType: 'renderer'
+    }
+  });
+}
+
+if (isProduction || process.env.SLOBS_REPORT_TO_SENTRY) {
```
andycreeth: We still aren't going to sent errors to sentry in development, but keeping this code here allows us to enable reporting to the development project by simply setting an env var.

PR #99, opened 2017-12-01
fix nested scenes

holiber:


test/api/scenes.js
```diff
@@ -125,3 +125,31 @@ test('Scenes events', async t => {
...
 
+
+test('Creating nested scenes', async t => {
+  const client = await getClient();
+
+  const sceneA = await client.request('ScenesService', 'createScene', 'SceneA');
+  const sceneB = await client.request('ScenesService', 'createScene', 'SceneB');
+  const sceneC = await client.request('ScenesService', 'createScene', 'SceneC');
+
+  await await client.request(sceneA.resourceId, 'addSource', sceneB.id);
```
andycreeth: Does this return a promise that resolves with another promise?  I guess I don't really understand why.

holiber: actually this is a mistake. Should be only one "await" here

PR #100, opened 2017-12-04
Implement guest API system

andycreeth:
## The Problem:
We need to expose an API for remotely loaded webpages in guest webviews that allows them to interact with the host application (Streamlabs OBS).  Because the host is extremely privileged (full node integration), we need to be very careful about what API we provide to the remote content.  We want to be able to provide the minimum possible API on a per-webview basis.  This will usually be 1 or 2 specific-purpose functions that the guest content will call.

## Theory
Electron allows us to inject a privileged preload script that can expose some global functions before node integration is removed from the webview and the guest content is loaded.

## Solution
For guest webviews that require access to our app, we require a preload script that sets up a global `streamlabsOBS` proxy object.  The Vue component that owns the webview can expose one or more functions to that webview by passing the webview and a request handler object to the `GuestApiService`, which performs the necessary setup.  When the guest content calls a function on the `streamlabsOBS` global, it is proxied to the request handler.  Any callback functions are replaced with placeholder objects and proxied via IPC events.  Functions on the request handler object must be asynchronous and return a promise.

### DIFF
--- a/app/services-manager.ts
+++ b/app/services-manager.ts
@@ -43,6 +43,7 @@ import { ObserveList } from './util/service-observer';
 import { Subject } from 'rxjs/Subject';
 import { Subscription } from 'rxjs/Subscription';
 import { Observable } from 'rxjs/Observable';
+import { GuestApiService } from 'services/guest-api';
 
 const { ipcRenderer } = electron;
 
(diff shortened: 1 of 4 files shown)


andycreeth:
Yup, my thoughts exactly.  `postMessage` was the first thing I looked for since it's the standard way to do cross origin communication like this.

PR #101, opened 2017-12-05
selectively lock out the UI while loading

andycreeth:
Previously we just replaced almost the entire UI with a spinner while loading.  However, overlays require that we actually be able to show more specific information while loading an overlay.  To enable this, we are moving over to more selectively locking out specific pieces of the UI.  Now, during the loading state we:

- Show a spinner in place of the Editor and Live pages
- Disable navigation in the Top Nav
- Disable toggling streaming and recording in the footer

This new system is possibly slightly more prone to breakage since more parts of the app need to know about and understand the loading state.  However, it is a little less jarring, and is required for overlays.

### DIFF
--- a/app/components/StartStreamingButton.vue
+++ b/app/components/StartStreamingButton.vue
@@ -1,7 +1,11 @@
 <template>
 <button
   class="button button--go-live"
-  :class="{'button--soft-warning': (streamingService.isStreaming && streamingService.stopStreaming), 'button--action': streamingService.startStreaming}"
+  :class="{
+    'button--soft-warning': (streamingService.isStreaming && streamingService.stopStreaming),
+    'button--action': streamingService.startStreaming
+  }"
+  :disabled="disabled"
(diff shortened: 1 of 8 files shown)


app/components/StartStreamingButton.vue
```diff
@@ -1,7 +1,11 @@
 <template>
 <button
   class="button button--go-live"
-  :class="{'button--soft-warning': (streamingService.isStreaming && streamingService.stopStreaming), 'button--action': streamingService.startStreaming}"
+  :class="{
+    'button--soft-warning': (streamingService.isStreaming && streamingService.stopStreaming),
```
holiber: `streamingService.stopStreaming` actually is a method, do you know that is purpose of that in the button class definition?

andycreeth: Huh, you're right.  That's super weird.  I have no idea where this logic came from, but it's pointless.  I will clean it up.

PR #102, opened 2017-12-05
API test part 2

holiber:
This PR includes:

- TypeScript support for tests
- unstable tests with application restart are fixed
- synchronous API calls support in tests
- errors from child window API requests now are sending to Raven
- better API errors handling

example how to run a single test:
`yarn test test-dist/test/scenes.js`


app/services-manager.ts
```diff
@@ -407,8 +438,31 @@ export class ServicesManager extends Service {
...
+   */
+  getResourceScheme(resourceId: string): Dictionary<string> {
+    const resource = this.getResource(resourceId);
+    if (!resource) {
+      this.requestErrors.push(`Resource not found: ${resourceId}`);
+      return null;
+    }
+    const resourceScheme = {};
+
+    Object.keys(resource.__proto__).concat(Object.keys(resource))
```
andycreeth: `__proto__` is considered a non-standard feature, and `Object.getPrototypeOf` is preferred

package.json
```diff
@@ -15,7 +15,7 @@
     "install-plugins": "node_modules/7zip-bin-win/x64/7za.exe x plugins/obs-browser-2987-old.7z -onode-obs/",
     "package": "rm -rf dist && build -w --x64 --em.env=production",
     "release": "yarn install --cwd bin && node bin/release.js",
-    "test": "ava",
+    "test": "cd test && tsc && ava",
```
andycreeth: We're trying to avoid changing directories in scripts like this.  Can we pass the relative path to `tsc` instead?

test/api/scenes.ts
```diff
@@ -0,0 +1,170 @@
...
+  t.is(lastEventData && lastEventData['sceneItemId'], image.sceneItemId);
+
+
+  image.setVisibility(false);
+  await sleep(100);
+  t.is(lastEventData && lastEventData['visible'], false);
+
+  lastEventData = null;
+  image.remove();
+  await sleep(100);
```
andycreeth: When we assert that an element is present on the page, it will search for the element for up to 2 seconds (or whatever time we set) before giving up.  This is nice because if the rendering is happening quickly, the tests don't spend unnecessary amounts of time waiting, but will wait a long time if they need to.

I think that ideally we would do something similar for API event tests.  API events, just like rendered DOM nodes, are asynchronous results that we expect to appear within a reasonable timeframe.  It seems like we should build a promise-based assertion helper that allows you to expect an event, maybe with a custom matcher function provided, and it will resolve as soon as it receives the event, or reject and fail the test if it reaches the timeout.  I think this system would be much more reliable when the system is running slowly, and would run much faster when things are running well.  In its current state, I am a little worried that this test may be unreliable, especially on slow CI VMs.

test/helpers/cmd-client.ts
```diff
@@ -0,0 +1,54 @@
...
+  args.forEach((arg, ind) => {
+    if (arg.charAt(0) === '"') {
+      args[ind] = arg.substring(1, arg.length - 1);
+    } else {
+      // console.info('argis', args[ind]);
+      args[ind] = JSON.parse(args[ind]);
+    }
+  });
+
+  // console.info(args);
```
andycreeth: Looks like a bunch of commented out debug logging in this file.

package.json
holiber: Actually after the script execution the directory will return to project's root. Will try to pass the relative path

andycreeth: K.  It's not a huge deal.

test/api/scenes.ts
holiber: You're right. Was too rushing while implementing this.

PR #104, opened 2017-12-06
v1 of widget files

andycreeth:
Allow exporting and importing of a single widget in a way that preserves the settings, size, and type.

### DIFF
--- a/app/services/widgets.ts
+++ b/app/services/widgets.ts
@@ -420,9 +420,10 @@ export class WidgetsService extends Service {
   /**
    * Save a widget file to the given path
    * @param path the path to the save the widget file
-   * @param widgetItem the widget to save
+   * @param widgetItemId the id of the widget to save
    */
-  async saveWidgetFile(path: string, widgetItem: SceneItem) {
+  async saveWidgetFile(path: string, widgetItemId: string) {
+    const widgetItem = this.scenesService.getSceneItem(widgetItemId);
     const data = this.exportWidgetJSON(widgetItem);
     const json = JSON.stringify(data, null, 2);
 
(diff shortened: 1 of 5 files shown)


app/services/widgets.ts
```diff
@@ -3,30 +3,45 @@ import { Service } from './service';
...
+  FollowerGoal = 2,
+  SubscriberGoal = 3,
+  BitGoal = 4,
+  DonationTicker = 5,
+  ChatBox = 6,
+  EventList = 7,
+  TheJar = 8,
+  ViewerCount = 9,
+  StreamBoss = 10,
+  Credits = 11
```
andycreeth: This `enum` is persisted in a number of locations now.  I have switched to hard coding the values to make it obvious that the values of existing should not be changed, otherwise it will cause lots of issues.

In the future, I would like to use string enums for stuff like this, but we will need to ugprade Typescript.

holiber: We upgraded TypeScript some months ago to 2.4. So we can use string enums now. Last time I told only about upgrade to 2.6

app/services/widgets.ts
```diff
@@ -388,12 +406,115 @@ export class WidgetsService extends Service {
...
+      );
+    });
+  }
+
+  /**
+   * Save a widget file to the given path
+   * @param path the path to the save the widget file
+   * @param widgetItem the widget to save
+   */
+  async saveWidgetFile(path: string, widgetItem: SceneItem) {
```
holiber: I would prefer to use serializable parameters for public methods. It will allow to use such methods from child window and expose this method to our public API if we need. So instead `SceneItem` it can be `ISceneItem` or `sceneItemId`. We always can fetch the sceneItem by id.

app/services/widgets.ts
andycreeth: oh ok cool

PR #107, opened 2017-12-07
Overlay marketplace

andycreeth:
iFrame in the overlay marketplace, and allow installation of overlays.


andycreeth:
The overlay marketplace on streamlabs.com is still being worked on, but I think this support code in slobs is ready to hit staging.  Going to open up for review.

app/services/app/app.ts
```diff
@@ -160,6 +162,20 @@ export class AppService extends StatefulService<IAppState>
...
   }
 
+  /**
+   * Downloads and installs an overlay
+   * @param url the URL of the overlay
+   */
+  async installOverlay(url: string, progressCallback?: (info: IDownloadProgress) => void) {
+    this.START_LOADING();
+
+    const pathName = await this.overlaysPersistenceService.downloadOverlay(url, progressCallback);
```
holiber: looks like we catch infinity load if `downloadOverlay` fails. Or I miss something?

andycreeth: Yeah, should probably `try/catch` and finish loading then re-raise

PR #110, opened 2017-12-12
Update to node-obs integrated with obs-studio-node

Xaymar:
Changes streamlabs-obs to use the node-obs version integrated into obs-studio-node, with all the new stuff it brings:

- OSN can now be used as an [actual node module](https://www.npmjs.com/package/@streamlabs/obs-studio-node), though work on that is lower priority than integrating IPC right now.
- Paths in node-obs are now automatically set to the __dirname of the index.js script of the module.
- Lower memory and runtime impact due to being a single dll instead of two.
- Debugging is now far easier as it will always copy either Debug or RelWithDebugInfo binaries including their pdbs.
- It's no longer necessary to load node-obs before obs-studio-node!

And all the developer benefits:

- Faster iteration: There is no need to modify both node-obs and obs-studio-node anymore!
- Better and faster debugging: Debug all of osn and node-obs at the same time.


eddyStreamlabs:
I would say that we should wait after the 0.8.0 release. A lot is already gonna be in this release that could cause some side effect problems.

andycreeth:
👍 I'm fine with waiting as long as it isn't going to hold up development of the IPC stuff.

Xaymar:
It should be fine to wait, the initial task is only to recreate the OSN api in an IPC compatible way, which is what I'm doing now.

PR #113, opened 2017-12-13
Troubleshooter and notifications

holiber:


app/components/NotificationsArea.vue.ts
```diff
@@ -0,0 +1,86 @@
...
+} from 'services/notifications';
+
+const LIFE_TIME = 15000;
+
+interface IUiNotification extends INotification {
+  outdated?: boolean;
+}
+
+@Component({})
+export default class NotificationsArea extends Vue {
```
andycreeth: Did you ever set up Prettier on your editor?  I think we should be running it on new files.  It will collapse multiple line endings in a row, which I know is a style you like.  However, I think in general as we have more and more people touching the JS it's just easier if everybody is running the same formatter.

app/components/windows/Troubleshooter.vue
```diff
@@ -71,19 +71,21 @@
...
+        When this happens, Streamlabs OBS does not have any resources left over to render frames.
       </p>
       <h4>
         What can I do?
       </h4>
 
       <ul>
-        <li>Enable VSync</li>
+        <li>Cap your in-game framerate</li>
+        <li>Enable VSync in your game</li>
```
holiber: Thanks for fixing

app/components/NotificationsArea.vue.ts
holiber: Oh, I thought Prettier does some changes before commit.. Actually the separating methods with 2 lines makes code more readable and helps to resolve merge conflicts since we use one-line spacing to separate logic blocks in one method. Maybe we can setup Prettier to do that? I know that some code combs have this option

app/components/NotificationsArea.vue.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    Vue.nextTick(() => {
+      if (this.notifications[0]) {
+        this.notifications[0].outdated = true;
+      }
+      Vue.nextTick(() => {
+        this.notifications.push({ ...notify, outdated: false });
+        window.setTimeout(() => this.hideOutdated(), LIFE_TIME);
+      });
+    });
```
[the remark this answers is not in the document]
holiber: Actually this code helps to properly play css animation of notifications. I'll add a comment

app/components/NotificationsArea.vue.ts
andycreeth: I have not enabled any pre-commit hooks for prettier yet, since that seemed a little aggressive.  I will probably end up enabling that soon.

Unfortunately it looks like prettier does not allow that as an option.  Personally I'm not convinced adding an additional newline necessarily makes it more readable.  Between indentation and TypeDoc comments, it's usually pretty easy at a glance to tell where top level functions begin and end.  Personally I'm not convinced that this particular issue is a big enough deal to stop using prettier.

andycreeth:
Hey I'm going to merge this so I can get a test build out to some people in the office.

PR #125, opened 2017-12-15
Fix the notifications and livedock layout

holiber:


andycreeth:
First off, it's called the "Live Dock", not the "Life Dock".  Can you please rename any code that refers to "Life Dock".

andycreeth:
I'm not entirely sure that we've nailed it here.  The live dock is resizable now, so it's still possible to get into some pretty janky states if your live dock is near either extreme of the slider.  I'm not sure basing our media queries based on whether the live dock is expanded is going to work well for us.

There may not be a good CSS-only solution here.

andycreeth:
This may be something we need to have Morgan take a look at when she gets back.  It seems like the new chat resizing just doesn't work well with the new notifications system.  I've been playing around with it for a while now and it just doesn't behave as expected in a lot of situations.

holiber:
Didn't know that we can resize it

andycreeth:
Yeah it's a brand new feature.  Unfortunately it was kind of bad timing to merge in the resizing and the notifications at the same time.  I think this is going to be a pretty challenging problem to solve.

andycreeth:
I've got another solution that I think should be more reliable than this.  Opening a PR soon.

PR #129, opened 2017-12-18
Set shutdown when not visible to true by default when importing from obs

eddyStreamlabs:


app/services/obs-importer.ts
```diff
@@ -142,6 +142,17 @@ export class ObsImporterService extends Service {
...
+            // Check "Shutdown source when not visible" by default for browser sources
+            if (source.type === 'browser_source') {
+              const sourcesProperties = source.getPropertiesFormData();
+              const shutdownPropertyIndex = sourcesProperties.indexOf(sourcesProperties.find(property => {
+                return property.name === 'shutdown';
+              }));
+
+              sourcesProperties[shutdownPropertyIndex].value = true;
+              source.setPropertiesFormData(sourcesProperties);
+            }
```
andycreeth: I think this is actually a lot more complicated and IPC heavy than it needs to be.  I would just modify `sourceJSON.settings` before you pass it to `sourcesService.createSource`.

It would look something like this:
```
if (sourceJSON.id === 'browser_source') {
  sourceJSON.settings.shutdown = true;
}

const source = this.sourcesService.createSource(...
```

eddyStreamlabs: Yeah true. Made the change.

PR #132, opened 2017-12-18
fix mixer memory leaks when switching scenes

andycreeth:
It turns out that `volmeter.addCallback` is not really safe memory-wise, and we should call it as infrequently as possible.

### DIFF
--- a/app/services/audio/audio.ts
+++ b/app/services/audio/audio.ts
@@ -30,6 +30,13 @@ export enum E_AUDIO_CHANNELS {
 
 const VOLMETER_UPDATE_INTERVAL = 100;
 
+interface IAudioSourceData {
+  fader?: obs.IFader;
+  volmeter?: obs.IVolmeter;
+  callbackInfo?: obs.ICallbackData;
+  stream?: Subject<IVolmeter>;
+}
+
 @InitAfter('SourcesService')
 export class AudioService extends StatefulService<IAudioSourcesState> implements IAudioServiceApi {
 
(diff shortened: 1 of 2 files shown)


app/services/audio/audio.ts
```diff
@@ -30,15 +30,22 @@ export enum E_AUDIO_CHANNELS {
 
 const VOLMETER_UPDATE_INTERVAL = 100;
 
+interface IAudioSourceData {
+  fader?: obs.IFader;
+  volmeter?: obs.IVolmeter;
+  callbackInfo?: obs.ICallbackData;
+  stream?: Subject<IVolmeter>;
```
holiber: If you only listen and don't push events to the `stream` from any object with type `IAudioSourceData` it's better to use `Observable` instead `Subject`. Subject inherits Observable.

andycreeth: How would I use an `Observable` in this case?  I see there is no `next` method on `Observable`.

holiber: Subject actually is Observable. So we can allow to emit events only in the service but not in the vue components by using this pattern:
```ts
function getTimer(): Observable<Date> {
  const timer = new Subject<Date>(); // timer is Subject here
  setInterval(() => {
     timer.next(new Date()); // can emit events here
  });
  return timer;
}

const timer = getTimer(); // timer is Observable here

timer.subscribe(time => console.log(time)); // can subscribe here

timer.next(new Date()); // error here, event emitting allowed only in the getTimer function
```
I don't see any code outside `initVolmeterStream()` that needs to emit events

andycreeth: Ah got it.  I will keep a reference to the `Subject` but store it as an `Observable`.

PR #133, opened 2017-12-19
dynamic notifications

andycreeth:
It's a little weird, but I think this is going to be the most reliable solution for notifications.  Basically the solution involves allowing the notifications area to grow/shrink to take up all the remaining space in the footer.  Then, the notifications area is constantly checking the size of itself, and hides the full text of the notification if the area it occupies is too small.

### DIFF
--- a/app/components/NotificationsArea.vue
+++ b/app/components/NotificationsArea.vue
@@ -17,17 +17,18 @@
     <span class="fa fa-info-circle"></span>
   </div>
 
-  <div class="notifications__container">
+  <div class="notifications__container flex--grow" ref="notificationsContainer">
     <div
-        v-for="notify in notifications"
-        class="notification"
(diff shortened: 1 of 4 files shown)


holiber:
It's better to use the event-based ResizeSensor from http://marcj.github.io/css-element-queries/ for observing elements size without creating additional `setInterval` cycles

andycreeth:
Hmm.  I would love to know what that library is doing under the hood.  There's no such thing as a resize event on an element.

andycreeth:
This library looks interesting, but it seems high risk to add right now.  I'm going to investigate it later.  There are a few different places this would be really useful to us.

holiber:
I used on some projects. It attaches an additional element to you element and check it's size with some magic on requestAnimationFrame. I don't remember exactly how, but fortunately the code of ResizeSensor is not big to investigate.

PR #136, opened 2017-12-19
upgrade vue class component

andycreeth:


andycreeth:
This has some issues with CI and only gets us partway to fixing the memory leak, so I'm going to close it for now.

PR #144, opened 2018-01-04
Visit every scene after loading to apply the shutdown property

eddyStreamlabs:


app/services/app/app.ts
```diff
@@ -97,6 +97,19 @@ export class AppService extends StatefulService<IAppState>
 
       this.performanceMonitorService.start();
 
+      // Workaround for the browser source shutdown settings being not applied correctly
+      this.scenesService.getScenes().forEach(scene => {
+        scene.items.forEach(item => {
+          let isShowing: Boolean = true;
```
andycreeth: you can remove the type declaration here, since you are assigning it immediately.  Anyway, `boolean` (lowercase) is the type you actually wanted.

app/services/app/app.ts
```diff
@@ -97,6 +97,25 @@ export class AppService extends StatefulService<IAppState>
 
       this.performanceMonitorService.start();
 
+      // Workaround for the browser source shutdown settings being not applied correctly
+      const activeScene = this.scenesService.activeScene;
+
+      this.scenesService.makeSceneActive(this.scenesService.getScenes()[0].id);
+      this.scenesService.getScenes().forEach((scene,index) => {
```
andycreeth: space after `scene,`

app/services/app/app.ts
```diff
@@ -97,6 +97,25 @@ export class AppService extends StatefulService<IAppState>
...
+        const previousScene = this.scenesService.activeScene;
+        this.scenesService.makeSceneActive(scene.id);
+        previousScene.items.forEach(item => {
+          let isShowing: Boolean = true;
```
andycreeth: You don't need to add a type annotation here since you are immediately assigning it.  Also, the type you would have wanted to use here is `boolean` (lowercase b)

app/services/app/app.ts
eddyStreamlabs: Done.

app/services/app/app.ts
eddyStreamlabs: Done.

PR #146, opened 2018-01-04
Add title attribute to BrowserWindow

AuroraAri:
This change makes it so that "slobs-client" isn't visible in Task Switcher.
  
image


andycreeth:
Thank you!

PR #151, opened 2018-01-05
Add scene collection help tip

andycreeth:


andycreeth:
Hmm this help tip might be getting in the way of the automated tests.

andycreeth:
Yeah something weird is going on with the tests.  I'm out next week unfortunately.  Maybe when he has some time @holiber can take a look into what might be going on.

holiber:
I can take a look next week

app/styles/classes.less
```diff
@@ -30,6 +30,15 @@
...
 }
 
+// Margins
+.margin-bot--20 {
+  margin-bottom: 20px;
+}
+
+.margin-right--20 {
+  margin-right: 20px;
+}
```
andycreeth: We should use a less loop to generate these helper classes for for all sides (top, bottom, left, right, vertical, horizontal) at some base unit of our choosing.

webpack.config.js
```diff
@@ -99,6 +99,15 @@ module.exports = {
...
           publicPath: 'bundles/'
         }
+      },
+      // Handles custom fonts. Currently used for icons.
+      {
+        test: /\.woff$/,
+        loader: 'file-loader',
+        options: {
+          name: '[name].[ext]',
+          outputPath: 'app/fonts/',
```
andycreeth: This is actually outputting to `bundles/app/fonts` which is a little strange.  I think you want this:
```
outputPath: 'fonts/',
publicPath: 'bundles/'
```
Similar to what we do for media files above.

app/components/windows/SourcesShowcase.vue
```diff
@@ -11,7 +11,8 @@
       v-if="inspectedSource === 'image_source'"
       @clickAdd="selectSource('image_source')"
       name="Image"
-      description="Add images to your scene.">
+      description="Add images to your scene."
+      key="1">
```
andycreeth: Why are we adding a key to these elements?  Normally you would only use a key when you have a `v-for` loop of dynamically generated elements, and you want to distinguish them.

app/components/windows/SourcesShowcase.vue
```diff
@@ -204,8 +223,14 @@
...
       @clickAdd="selectWidget(widgetTypes.DonationTicker)"
       name="Donation Ticker"
-      description="Show off your most recent donations to your viewers.">
-      <img slot="media" src="../../../media/source-demos/donationticker.gif"/>
+      description="Show off your most recent donations to your viewers."
+      key="15">
+      <video class="source__demo source__demo--day" autoplay loop slot="media">
+        <source src="../../../media/source-demos/day/source-donation-ticker.mp4">
+      </video>
+      <video class="source__demo source__demo--night" autoplay loop slot="media">
```
andycreeth: Does this work?  Rendering two elements into the same slot?

app/components/windows/SourcesShowcase.vue
```diff
@@ -415,7 +503,7 @@
           <div
             class="source source--widget"
             :class="{'source--active': inspectedSource === widgetTypes.TheJar}"
-            @click="inspectSource(widgetTypes.TheJar)"
+            @click="inspectSource(widgetTypes.TheJar, 'theJar')"
```
andycreeth: Does this string here do anything?  As far as I can tell, this function only takes 1 argument.

app/components/windows/SourcesShowcase.vue
```diff
@@ -465,7 +553,7 @@
           <div
             class="source source--widget"
             :class="{'source--active': inspectedSource === widgetTypes.ChatBox}"
-            @click="inspectSource(widgetTypes.ChatBox)"
+            @click="inspectSource(widgetTypes.ChatBox, 'chatbox')"
```
andycreeth: Same comment as above.

app/components/windows/AddSourceInfo.vue
```diff
@@ -14,7 +14,6 @@
     <div class="source-info__media">
       <slot name="media"></slot>
     </div>
-    <!--<button @click="$emit('clickAdd')" class="button button--action">Add Source</button>-->
```
andycreeth: If we're removing this can we please delete it rather than commenting it out?

app/styles/classes.less
andycreeth: This doesn't have to be part of this PR, was just a general comment.

app/components/windows/SourcesShowcase.vue
mbiemiller: I changed the media from gifs to mp4s which caused a bug where the path of the source would change but the actual video element wouldn't reload so the wrong video would play. This can happen when you have multiple on the same component on a page so I had to add a key attribute to fix this.

app/components/windows/SourcesShowcase.vue
mbiemiller: Yep. One gets display: none based on the day/night mode

app/components/windows/AddSourceInfo.vue
mbiemiller: This is a deletion.

app/styles/classes.less
mbiemiller: Thanks- will change for next time

app/components/windows/SourcesShowcase.vue
mbiemiller: Ideally I will refactor this whole page in the future to be more dynamic and use a JavaScript object to fill the data

andycreeth: Ah I see.  That's fine.  I agree eventually we should refactor this to render off of a JSON file or something.

app/components/windows/SourcesShowcase.vue
andycreeth: Huh ok cool.

app/components/windows/AddSourceInfo.vue
andycreeth: I apparently hadn't had enough of my coffee this morning when I posted this.  Oops.  😴

PR #152, opened 2018-01-08
Save input and filter sources before caching

computerquip-streamlabs:


computerquip-streamlabs:
I've also added the obs-text binaries to the obs-studio-node update with this PR. I'll add the frontend changes in another PR.

PR #153, opened 2018-01-09
All output channels must be set to null before shutdown

computerquip-streamlabs:


app/services/sources/sources.ts
```diff
@@ -170,6 +170,17 @@ export class SourcesService extends StatefulService<ISourcesState> implements IS
...
+     * sure we reset the channel it's set to,
+     * otherwise OBS thinks it's still attached
+     * and won't release it. */
+    const channel = this.state.sources[id].channel;
+
+    /* Be consistent with initial check */
+    if (channel !== void 0) {
+      obs.Global.setOutputSource(channel, null);
+    }
+
```
holiber: Just to simplify put the code above `const source = this.getSource(id);`
```
const source = this.getSource(id);
 if (source.channel !== void 0) {
    obs.Global.setOutputSource(channel, null);
 }
```

computerquip-streamlabs: Fixed in forced push.

app/services/scenes-transitions.ts
```diff
@@ -64,6 +64,11 @@ export class ScenesTransitionsService extends StatefulService<ISceneTransitionsS
     this.getCurrentTransition().release();
   }
 
+  reset() {
+    this.getCurrentTransition().release();
```
andycreeth: NBD, but for DRY-ness this could just be `this.release()`

app/services/sources/sources.ts
```diff
@@ -171,6 +171,15 @@ export class SourcesService extends StatefulService<ISourcesState> implements IS
 
   removeSource(id: string) {
     const source = this.getSource(id);
+
+   /* When we release sources, we need to make
```
andycreeth: This is misaliagned.

PR #155, opened 2018-01-11
custom resolution support

holiber:


eddyStreamlabs:
Seems to work great : 
 - Upscaling / downscaling works
 - Importing custom resolutions from obs works 

-> Require an update of obs-studio-node.

holiber:
Lets merge it after obs-studio-node will be merged

eddyStreamlabs:
Sounds good

holiber:
Yeah, we are trying to use the `multiselect` like `autocomplete` component. Actually, if we have plans to use that component only for the resoultion selector, I prefer to use the standard `multiselect` with a  `custom` option which can switch the multiselect view to the 2 simple text fields - `width` and `height`

holiber:
@eddyStreamlabs are node-obs changes merged?

eddyStreamlabs:
Should be now.

PR #156, opened 2018-01-11
Import obs current scene collection

eddyStreamlabs:


app/services/obs-importer.ts
```diff
@@ -51,6 +51,18 @@ export class ObsImporterService extends Service {
...
     this.importProfile(selectedprofile);
 
+    // Select current scene collection
+    const globalConfigFile = path.join (this.OBSconfigFileDirectory, 'global.ini');
+
+    const data = fs.readFileSync(globalConfigFile).toString();
+
+    if (data) {
+      const currentSceneCollection =
+        data.match(/^SceneCollection\=(.*)$/gm).toString().replace('SceneCollection=', '');
```
andycreeth: Rather than doing a global match and a string substitution, you can just use the regex capture group you already have in there.  Also, you should probably check that you got a match otherwise this will just raise an exception and kill the import.  I don't see any reason why this should be a fatal exception, so let's gracefully recover.  This would look like:

```typescript
const match = data.match(/^SceneCollection\=(.*)$/m);
if (match && match[1]) this.appService.loadConfig(match[1]);
```

eddyStreamlabs: Done.

PR #157, opened 2018-01-11
Add fix for custom fonts

computerquip-streamlabs:
There's a new migration pattern for sources that will add face properties to the settings.font object. This is similar to a normal font but wasn't required before. 

This should also fix custom fonts and outlines.

### DIFF
--- a/app/services/font-library.ts
+++ b/app/services/font-library.ts
@@ -45,6 +45,27 @@ export class FontLibraryService extends Service {
     return Promise.resolve(this.manifest);
   }
 
+  /* This scans the entire manifest (which is slow)
+   * to find a filename, then finds the family name
+   * in the parent. This also doesn't take into account
+   * font files yet. This function is used mostly for
+   * upgrading config schema */
+  findFontFile(baseFileName: string): Promise<IFontFamily> {
+    return this.getManifest().then(manifest => {
+      for (const family_name in manifest.families) {
+        const family = manifest.families[family_name];
+
+        for (const style_name in family.styles) {
+          const style = family.styles[style_name];
+
+          if (baseFileName === style.file) {
+            return family;
+          }
+        }
+      }
+    });
+  }
+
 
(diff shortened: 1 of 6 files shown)


computerquip-streamlabs:
This also requires this PR first since it it updates to the required version of obs-studio-node: https://github.com/stream-labs/streamlabs-obs/pull/152

app/services/font-library.ts
```diff
@@ -45,6 +45,27 @@ export class FontLibraryService extends Service {
...
   }
 
+  /* This scans the entire manifest (which is slow)
+   * to find a filename, then finds the family name
+   * in the parent. This also doesn't take into account
+   * font files yet. This function is used mostly for
+   * upgrading config schema */
+  findFontFile(baseFileName: string): Promise<IFontFamily> {
+    return this.getManifest().then(manifest => {
+      for (let i = 0; i < manifest.families.length; ++i) {
```
andycreeth: Dat pre-increment optimization.

app/services/scenes-collections/nodes/sources.ts
```diff
@@ -1,12 +1,14 @@
 import { Node } from './node';
 import { SourcesService, Source, TSourceType } from '../../sources';
 import { FiltersNode } from './filters';
+import { FontLibraryService } from '../../font-library';
```
andycreeth: We've started importing things relative to `app` rather than relative to the file it's in.  This makes things very easy when moving files around.  So for example you can do this import as `import { FontLibraryService } from 'services/font-library';`.  We have webpack set up so it will recognize imports relative to the `app` directory.  The only exception to this is when two files are very tightly coupled and will always be in the same place (part of the same service).  In that case it's fine to do a relative import.

computerquip-streamlabs: I changed this and for consistency, changed the others. See if that looks alright.

PR #158, opened 2018-01-11
Add a restart streamlabels session button in the settings

eddyStreamlabs:


app/components/ExtraSettings.vue.ts
```diff
@@ -67,4 +69,10 @@ export default class ExtraSettings extends Vue {
     });
   }
 
+  restartStreamlabelsSession() {
+    if (this.streamlabelsService.restartSession()) {
```
andycreeth: This returns a promise, so you will need to wait on the promise to actually make sure it actually succeeded.

eddyStreamlabs: Done.

PR #168, opened 2018-01-13
Return property values directly, even if it's undefined

computerquip-streamlabs:


andycreeth:
Gonna merge this since I think Eddy wanted to do a preview release today.

PR #169, opened 2018-01-15
force per machine installation

andycreeth:
This forces a per machine installation and will require admin rights for install.  I am going to do some testing.  We may not end up releasing this.


andycreeth:
Super bizarre that appveyor is failing.  I think our VM must be choking or something.  We don't package for tests so this has zero impact.

PR #170, opened 2018-01-15
update electron builder

andycreeth:
Contains a fix to our `.obj` issue.


andycreeth:
Currently blocked by this: https://github.com/electron-userland/electron-builder/issues/2477

I may end up upgrading to `19.53.1` since that seems to work currently.

andycreeth:
We're going to move to 19.52.1

PR #171, opened 2018-01-16
Fetch scene collections shema

holiber:
We need this PR to allow Streamdeck to fetch some parts of the config tree:
image

### DIFF
--- a/app/services/scenes-collections/config.ts
+++ b/app/services/scenes-collections/config.ts
@@ -2,9 +2,9 @@ import { PersistentStatefulService } from '../persistent-stateful-service';
 import { mutation } from '../stateful-service';
 import { Inject } from '../../util/injector';
 import { RootNode } from './nodes/root';
-import { SourcesNode } from './nodes/sources';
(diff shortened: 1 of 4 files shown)


app/services/scenes-collections/config.ts
```diff
@@ -301,6 +301,42 @@ export class ScenesCollectionsService extends PersistentStatefulService<IScenesC
   }
 
 
+  fetchSceneCollectionsSchema(): Dictionary<ISceneCollectionSchema> {
+    const schemes: Dictionary<ISceneCollectionSchema> = {};
+    this.state.scenesCollections.forEach(configName => {
+      const data = fs.readFileSync(this.getConfigFilePath(configName)).toString();
```
andycreeth: I'm not sure how often streamdeck is going to be calling this function, but this could potentially get a little slow.  We may need to make this async eventually.

holiber: Ok

andycreeth:
Gonna merge so we can put this in the preview release.

PR #172, opened 2018-01-16
Remove the process priority setting from the blacklist

eddyStreamlabs:
This PR needs to have the latest version of obs-studio-node in order to work.


andycreeth:
We should release a new OSN and update it in slobs before merging this.

PR #185, opened 2018-01-18
save the livedock and chat size

holiber:
also, the customization service has been refactored

### DIFF
--- a/app/services/customization.ts
+++ b/app/services/customization.ts
@@ -8,7 +8,11 @@ interface ICustomizationServiceState {
   leftDock: boolean;
   hideViewerCount: boolean;
   livedockCollapsed: boolean;
+  previewSize: number;
+  livedockSize: number;
 }
+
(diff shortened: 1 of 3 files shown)


app/services/customization.ts
```diff
@@ -22,46 +26,26 @@ export class CustomizationService extends PersistentStatefulService<ICustomizati
...
-  @mutation()
-  SET_LIVE_PREVIEW_ENABLED(enabled: boolean) {
-    this.state.livePreviewEnabled = enabled;
-  }
-
-  @mutation()
-  SET_LEFT_DOCK(enabled: boolean) {
-    this.state.leftDock = enabled;
+  setSettings(settingsPatch: Partial<ICustomizationSettings>) {
+    this.SET_SETTINGS(settingsPatch);
```
andycreeth: Lol I've been meaning to make this change forever but kept being lazy.  👍

PR #195, opened 2018-01-23
preformance mode

holiber:


app/components/MixerVolmeter.vue
```diff
@@ -0,0 +1,48 @@
...
+</div>
+
+</template>
+
+<script lang="ts" src="./MixerVolmeter.vue.ts"></script>
+
+<style lang="less" scoped>
+@import "../styles/index";
+
+.volmeter {
```
andycreeth: It looks like this CSS was copied from `MixerItem.vue`.  Can we remove from there if this is its new home?

app/services/customization.ts
```diff
@@ -10,6 +11,7 @@ interface ICustomizationServiceState {
   livedockCollapsed: boolean;
   previewSize: number;
   livedockSize: number;
+  previewEnabled: boolean;
```
andycreeth: Can we call this performance mode instead?

app/util/menus/EditMenu.ts
```diff
@@ -145,6 +147,17 @@ export class EditMenu extends Menu {
         label: 'Unlock all sources',
         click: () => this.scenesService.setLockOnAllScenes(false)
       });
+
+      this.append({ type: 'separator' });
+
+      this.append({
+        label: 'Enable Preview',
```
andycreeth: I think we should call this performance mode.

app/services/customization.ts
```diff
@@ -28,16 +30,20 @@ export class CustomizationService extends PersistentStatefulService<ICustomizati
...
+  settingsChanged = new Subject<Partial<ICustomizationSettings>>();
+
   init() {
     super.init();
     this.setLiveDockCollapsed(true);// livedock is always collapsed on app start
   }
 
   setSettings(settingsPatch: Partial<ICustomizationSettings>) {
     this.SET_SETTINGS(settingsPatch);
+    this.settingsChanged.next(settingsPatch);
```
andycreeth: Does anything subscribe to this?  What is this for?

app/components/MixerVolmeter.vue
holiber: 👍 +1

app/services/customization.ts
holiber: Ups, my first attempt was event-based. Forgot to delete this.

andycreeth:
I'm going to merge this so we can get it into the preview build I am about to release.

PR #196, opened 2018-01-23
remove CPU stat from the UI

andycreeth:
We've discovered that this number doesn't fully account for all application processes, and is usually lower than the number reported by task manager.  We don't want to be giving users inaccurate information, so we are just going to disable this until we can get more accurate data.  In the meantime we can point people towards task manager or other 3rd party tools for accurately tracking CPU usage.

### DIFF
--- a/app/components/PerformanceMetrics.vue.ts
+++ b/app/components/PerformanceMetrics.vue.ts
@@ -7,16 +7,8 @@ import { compact } from 'lodash';
 
 @Component({})
 export default class PerformanceMetrics extends Vue {
-
-  @Inject()
-  streamingService: StreamingService;
-
-  @Inject()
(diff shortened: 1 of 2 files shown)


app/components/PerformanceMetrics.vue.ts
```diff
@@ -27,7 +19,9 @@ export default class PerformanceMetrics extends Vue {
   }
 
   get percentDropped() {
-    return (this.performanceService.state.percentageDroppedFrames || 0).toFixed(1);
+    return (this.performanceService.state.percentageDroppedFrames || 0).toFixed(
+      1
+    );
```
andycreeth: Prettier does weird things sometimes...

PR #202, opened 2018-01-24
Scene Collections V2: Backup and Sync

andycreeth:
A rewrite of the scene collections service to support multiple user accounts and backup/sync.


app/services/scene-collections/index.ts
```diff
@@ -0,0 +1,887 @@
...
+    channel: number;
+  }[];
+}
+
+/**
+ * V2 of the scene collections service:
+ * - Completely asynchronous
+ * - Server side backup
+ */
+export class SceneCollectionsService extends StatefulService<ISceneCollectionsManifest> {
```
holiber: We have to expose some API for Streamdeck. To do that create a file with `#serviceName#-api.ts` name and define an interface SceneCollectionsServiceApi. List the methods to expose and add some jsdoc for it. To have a typecheck add `implements SceneCollectionsServiceApi` here.

app/services/scene-collections/scene-collections-api.ts
```diff
@@ -0,0 +1,34 @@
+export interface ISceneCollectionsServiceApi {
+  load(id: string, shouldAttemptRecovery?: boolean): Promise<void>;
+  create(name?: string, setupFunction?: () => void): Promise<void>;
```
holiber: We can't expose functions as arguments to the API. You can just omit the second param here.

app/services/scene-collections/scene-collections-api.ts
```diff
@@ -0,0 +1,34 @@
+export interface ISceneCollectionsServiceApi {
+  load(id: string, shouldAttemptRecovery?: boolean): Promise<void>;
```
holiber: We need some jsdoc comments here for docs

app/services/scene-collections/scene-collections.ts
```diff
@@ -0,0 +1,865 @@
...
+  activeId: string;
+  collections: ISceneCollectionsManifestEntry[];
+}
+
+/**
+ * V2 of the scene collections service:
+ * - Completely asynchronous
+ * - Server side backup
+ */
+export class SceneCollectionsService extends StatefulService<ISceneCollectionsManifest> {
```
holiber: `implements ISceneCollectionsServiceApi`

app/services/scene-collections/scene-collections.ts
```diff
@@ -0,0 +1,865 @@
...
+                      (sceneItemData: ISceneItemInfo) => {
+                        return {
+                          sceneItemId: sceneItemData.id,
+                          sourceId: sceneItemData.sourceId
+                        };
+                      }
+                    )
+                  };
+                }
+              ),
```
holiber: omg, is it autoformatting? Looks like the Lisp hell

app/services/scene-collections/scene-collections.ts
```diff
@@ -0,0 +1,865 @@
...
+    this.ADD_COLLECTION(id, name, (new Date()).toISOString());
+  }
+
+  /**
+   * Deletes on the server and removes from the store
+   */
+  private async removeCollection(id: string) {
+    this.DELETE_COLLECTION(id);
+
+    // TODO: Delete the file (for safety and recoverability we soft delete for now)
```
holiber: Should  this TODO be a part of this PR?

app/services/scene-collections/scene-collections.ts
```diff
@@ -0,0 +1,865 @@
...
+    } catch (e) {
+      console.error('Scene collection sync failed: ', e);
+    }
+  }
+
+  /**
+   * Duplicates the current scene collection
+   * @param name the name of the new scene collection
+   */
+  async duplicate(name: string) {
```
holiber: it is not obvious that this method also switch the scene collection. Maybe rename it to something like `duplicateAndSetActive` ?

app/services/scene-collections/scene-collections.ts
```diff
@@ -0,0 +1,865 @@
...
+  /**
+   * Install a new overlay from a URL
+   * @param url the URL of the overlay file
+   * @param name the name of the overlay
+   * @param progressCallback a callback that receives progress of the download
+   */
+  async installOverlay(url: string, name: string, progressCallback?: (info: IDownloadProgress) => void) {
+    this.startLoadingOperation();
+
+    // TODO: Handle Errors
```
holiber: should this TODO be a part of this PR?

app/services/scene-collections/scene-collections-api.ts
andycreeth: 👍 It should be used over the API anyway.  It's just a hack for the OBS importer to use.

app/services/scene-collections/scene-collections.ts
andycreeth: Yeah, meant to do that.

app/services/scene-collections/scene-collections.ts
andycreeth: Haha yeah I guess so.  This function does have a lot of indentation, which the auto formatter seems to strange things with.

app/services/scene-collections/scene-collections.ts
andycreeth: Yeah I'm still undecided as to whether we should hard delete files.  I'll make a final decision and remove the comment.

andycreeth:
Thanks for reviewing, I know it's a big PR.  It looks like this is going to be on hold for at least another day due to current memory leak issues on staging.  So I am going to take advantage of that time to write a bunch of tests.

holiber:
We should update `docs/README.md` `How to switch a scene collection?` section

PR #205, opened 2018-01-24
Appearance settings and chat resize

holiber:
image


andycreeth:
There are some issues here:

- Changing the chat size requires refreshing the chat.  Right now it just looks like the slider has no effect when you change it.  I'm not sure why this was done like this, since it is possible to change the zoomfactor of a webview without refreshing it.

- I was talking to Eddy and we think that "Restore Defaults" is a little confusing because it only restores the defaults on this tab.  I think people are going to be worried about resetting all of their settings.  I think for now we should just remove the button.

- Despite the fact that these settings all share the fact that they are not OBS settings, and are SLOBS specific settings, they really don't belong grouped together from a behavioral standpoint.  Night mode, left side dock, and text size are all definitely appearance settings and belong here.  I think the stream info and performance mode should be moved back to the general tab.

andycreeth:
This is looking good.  Let's do 1 last thing before this is ready:
- I think we should move the magnifying glass slider under the chat into the appearance settings.  We can call it "Chat Width".  I don't think it makes sense for us to be taking up screen real estate with something that will be so infrequently changed.

PR #211, opened 2018-01-25
BTTV emotes

holiber:


app/services/customization/customization.ts
```diff
@@ -31,7 +31,8 @@ export class CustomizationService
     previewSize: 300,
     livedockSize: 0.28,
     performanceMode: false,
-    chatZoomFactor: 1
+    chatZoomFactor: 1,
+    enableBBTVEmotes: false
```
holiber: The code of including BBTTV emotes based on my reverse-engineering of BetterTTV plugin behavior. So I'm not sure that this code will work forever and prefer to disable this feature by default.

app/components/Chat.vue.ts
```diff
@@ -25,25 +26,41 @@ export default class Chat extends Vue {
...
+        
+        localStorage.setItem('bttv_clickTwitchEmotes', true);
+        
+        var bbtvscript1 = document.createElement('script');
+        bbtvscript1.setAttribute('src','https://cdn.betterttv.net/betterttv.js');
+        document.head.appendChild(bbtvscript1);
+        
+        var bbtvscript2 = document.createElement('script');
+        bbtvscript2.setAttribute('src','https://legacy.betterttv.net/betterttv.js');
+        document.head.appendChild(bbtvscript2);
```
andycreeth: This is a potential security risk.  I am sticking to a blanket rule of never executing any code loaded over the internet, since this code has full access to all of the node APIs.  Even though `bettertv.net` is probably a semi-trusted source, it's just not worth the risk to give downloaded scripts access to node APIs.

Instead, you should download the scripts using `fetch` (or just commit them to the repo), and inject them into the webview using `webview.executeJavaScript`, which will execute the javascript inside the sandboxed iframe.  Do not use `webview.preload` since that is meant for injecting privileged scripts to set up restricted APIs.

andycreeth: Can you indent this one level further.  I didn't see the `executeJavascript` call above.

PR #214, opened 2018-01-25
Youtube token refresh

kuhasuki:
-Add a requiresToken decorator to youtube service which handles token refresh on 401 status and retries failed request
-Change request errorHandler to return rejected promise with request data for failing requests
-Only show chat while actively streaming for Youtube
-Add a mutator for updating platform token to user service

### DIFF
--- a/app/components/Chat.vue.ts
+++ b/app/components/Chat.vue.ts
@@ -1,13 +1,13 @@
 import Vue from 'vue';
 import { Subscription } from 'rxjs/Subscription';
 import { Component } from 'vue-property-decorator';
-import { UserService } from '../services/user';
-import { Inject } from '../util/injector';
-import { getPlatformService } from '../services/platforms';
-import { CustomizationService } from '../services/customization';
+import { UserService } from 'services/user';
+import { Inject } from 'util/injector';
+import { getPlatformService } from 'services/platforms';
+import { CustomizationService } from 'services/customization';
 import url from 'url';
 import electron from 'electron';
-import { ICustomizationSettings } from '../services/customization/customization-api';
+import { ICustomizationSettings } from 'services/customization/customization-api';
 
 @Component({})
 export default class Chat extends Vue {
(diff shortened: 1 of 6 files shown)


app/components/Chat.vue.ts
```diff
@@ -7,7 +7,7 @@ import { getPlatformService } from '../services/platforms';
 import { CustomizationService } from '../services/customization';
 import url from 'url';
 import electron from 'electron';
-import { ICustomizationSettings } from "../services/customization/customization-api";
+import { ICustomizationSettings } from '../services/customization/customization-api';
```
andycreeth: We have webpack configured so you can reference things relative to the `app` directory.  It makes things easier when moving files around.  So this can just be: `import { ICustomizationSettings } from 'services/customization/customization-api';`

There are lots of places we don't do this, but we are trying to opportunistically change them as we touch files.

app/services/platforms/youtube.ts
```diff
@@ -155,8 +188,32 @@ export class YoutubeService extends Service implements IPlatformService {
...
+    return {
+      ...descriptor,
+      value(...args: any[]) {
+        return original.apply(target.constructor.instance, args)
+        .catch((error:Response) => {
+          if (error.status === 401) {
+            return target.fetchNewToken().then(() => {
+              return original.apply(target.constructor.instance, args);
+            });
+          }
```
andycreeth: Let's add an `else` case here that returns a rejected promise with the response object.  Otherwise this decorator is just going to eat any non-401 errors and not surface them.

app/services/platforms/youtube.ts
```diff
@@ -155,8 +188,32 @@ export class YoutubeService extends Service implements IPlatformService {
...
+ * fresh youtube api token is available
+ */
+function requiresToken() {
+  return (target: any, methodName: string, descriptor: PropertyDescriptor) => {
+    const original = descriptor.value;
+    return {
+      ...descriptor,
+      value(...args: any[]) {
+        return original.apply(target.constructor.instance, args)
+        .catch((error:Response) => {
```
andycreeth: Space after `:`

andycreeth:
A couple small changes, but this is awesome 👍

app/services/platforms/youtube.ts
kuhasuki: fixed in dc4ea8348996ab54bd0d1c2307f2c6abd69100dc

PR #218, opened 2018-01-29
modules: Update stream-labs/node-libuiohook

Xaymar:
This changes our hotkey approach to "passthrough", the same approach OBS is using. These are the changes done in the module itself:

> # Switch to "passthrough" hotkey model
> 
> This type of key checking is easier to maintain, better supported on other platforms and was already partially in use by the previous code. In addition to that it's also completely free of any delayed errors that may happen, since all the code that does the checking is based on checking manually instead of registering something with the kernel.
> 
> Additionally I have removed libuiohook. It has not been used since the implementation of the RegisterHotkey based one and will not be used in the future. This code can practically be merged into obs-studio-node as a utility. If we do merge it with obs-studio-node, we can match OBS's behavior 1:1.

While creating this i noticed that Streamlabs OBS does not register _any_ hotkeys until the user opens the Settings > Hotkey page. This is a rather serious bug and explains several bug reports I have seen.


Xaymar:
Related issue: https://tracker.streamlabs.com/view.php?id=151

PR #221, opened 2018-01-30
Automated osn builds

computerquip-streamlabs:


package.json
```diff
@@ -77,7 +77,7 @@
     "font-manager": "git+https://github.com/computerquip-streamlabs/font-manager.git",
     "node-fontinfo": "^0.0.2",
     "node-libuiohook": "file:./node-libuiohook",
-    "obs-studio-node": "https://github.com/stream-labs/obs-studio-node/releases/download/v0.0.8/obs-studio-node.tar.gz",
+    "obs-studio-node": "https://ci.appveyor.com/api/buildjobs/imlwvrtbc59x17ly/artifacts/streamlabs-x86_64.tar.gz",
```
andycreeth: So does `imlwvrtbc59x17ly` uniquely identify the build we are currently using?  And when we want to update we just change that string?

computerquip-streamlabs: The URL itself is taken from here: https://ci.appveyor.com/project/computerquip-streamlabs/obs-studio-node/build/artifacts

That string represents the build. When we want to change builds, we can just take the URL from the build we want to use. There's also permalinks we can use here: https://www.appveyor.com/docs/packaging-artifacts/#permalink-to-the-last-successful-build-artifact

Not sure if they might become useful eventually.

andycreeth: Should we move that to the streamlabs appveyor account?

computerquip-streamlabs: We don't really benefit from it (public repo and all). We can eventually for organization purposes I suppose. I just did it on my account since I don't think I had permission to add projects to the streamlabs account.

andycreeth: I think organizationally we should keep things in an account that is shared/owned by multiple people on the team.  It's much better for everyone when people go on vacation and whatnot.  I can make you an admin on the streamlabs appveyor account.

computerquip-streamlabs: Alright. Shouldn't take but a few minutes to move everything over.

computerquip-streamlabs:
Actually before pulling this, I'm going to add the signed binaries from obs to binaries so it will all just work out of the box. I can also get the other binaries signed although that isn't quite as pressing at the moment.

andycreeth:
Oh yeah I forgot about that.  Good call.

computerquip-streamlabs:
Should be good now. 
Things to test:
1. Each *.dll, *.exe, and *.node file should be signed with General Workings with the exception of all files within `obs-studio-node/libobs/data/obs-plugins/win-capture` which should be signed by Hugh Bailey. 

2. Browser source and facemask filter (things that are not distributed with obs-studio-node in otherwords) should still load and work correctly. 

3. There should not be any negative impacts on startup in general.

PR #224, opened 2018-01-31
Multiselection

holiber:
- multiple selection (multiple drag, transform, copy/past, `Ctr+A` shortcut)
- copy/past now preserves scene items settings
- sceneItem API has been reduced - some simple methods like `setPosition`, `setPositionAndCrop` now replaced with `setSettings`


app/components/StudioEditor.vue.ts
```diff
@@ -229,7 +245,7 @@ export default class StudioEditor extends Vue {
       });
     });
 
-    this.scene.getItem(source.sceneItemId).setPositionAndCrop(rect.x, rect.y, rect.crop);
+    this.scene.getItem(source.sceneItemId).setSettings({ x: rect.x, y: rect.y, crop: rect.crop });
```
andycreeth: When I started working on Undo/Redo, I had a branch that rolled up a lot of these methods into a single function called `setTransform`, that took an `ITransform` (or maybe a `Partial<ITransform>`).

An `ITransform` looked like this:
```typescript
interface ITransform {
  position: IVec2;
  scale: IVec2;
  crop: ICrop;
  rotation: number;
}
```

I honestly thought I merged this branch, but I guess I didn't.  I think this was a good system because it grouped together some very logical things into a single object describing the position/shape of the item.  The undo/redo system was going to store any change to this object as a "Transform" event and store the before/after state of this object.

What do you think about this approach?  Maybe since we're in here we should make that change now?  Let me see if I can find that branch anywhere...

app/services/app/app.ts
```diff
@@ -49,6 +50,7 @@ export class AppService extends StatefulService<IAppState> {
   @Inject() private ipcServerService: IpcServerService;
   @Inject() private tcpServerService: TcpServerService;
   @Inject() private performanceMonitorService: PerformanceMonitorService;
+  @Inject() private selectionService: SelectionService;
```
andycreeth: As far as I can see, this service isn't used here.

app/services/clipboard.ts
```diff
@@ -4,63 +4,69 @@ import { SourcesService } from './sources';
 import { shortcut } from './shortcuts';
 import { Inject } from '../util/injector';
 import { SourceFiltersService } from './source-filters';
+import { SelectionService } from 'services/selection';
 
 
 interface IClipboardState {
-  sourcesIds: string[];
+  sceneItemsIds: string[];
```
andycreeth: `sceneItemIds`

app/services/clipboard.ts
```diff
@@ -4,63 +4,69 @@ import { SourcesService } from './sources';
...
   filtersIds: string[];
```
andycreeth: I know you didn't change this, but correct grammar would be `filterIds` to be consistent.

app/services/clipboard.ts
```diff
@@ -4,63 +4,69 @@ import { SourcesService } from './sources';
...
+  @Inject() private sourcesService: SourcesService;
+  @Inject() private sourceFiltersService: SourceFiltersService;
+  @Inject() private selectionService: SelectionService;
 
   @shortcut('Ctrl+C')
   copy() {
-    const source = this.scenesService.activeScene.activeItems[0];
-    if (!source) return;
-    this.SET_SOURCES_IDS([source.sourceId]);
+    const itemsIds = this.selectionService.getIds();
```
andycreeth: `itemIds`

app/services/scenes/scene-item.ts
```diff
@@ -1,10 +1,14 @@
-import { ScenesService, Scene, ISceneApi, ISceneItem, ISceneItemApi, ISceneItemInfo } from './index';
+import { ScenesService, Scene, ISceneItem, ISceneItemApi, ISceneItemInfo } from './index';
 import { mutation, ServiceHelper } from '../stateful-service';
 import Utils from '../utils';
-import { Source, SourcesService, TSourceType, ISource } from '../sources';
+import { SourcesService, TSourceType, ISource } from 'services/sources';
+import { VideoService } from 'services/video';
+import { ScalableRectangle } from '../../util/ScalableRectangle';
```
andycreeth: Import can be from `util/ScalableRectangle`

app/services/scenes/scene.ts
```diff
@@ -61,18 +59,8 @@ export class Scene implements ISceneApi {
...
-      return !this.activeItemIds.includes(sourceModel.sceneItemId);
-    }).map(source => {
-      return this.getItem(source.sceneItemId);
-    });
-  }
-
-
-  get activeItems(): SceneItem[] {
-    return this.activeItemIds.map(itemId => this.getItem(itemId));
+  getItemsIds(): string[] {
```
andycreeth: `getItemIds`

app/components/StudioEditor.vue.ts
holiber: My main goal in this refactoring was to rid of multiples `sceneItemChanged` events that makes a lot of spam in our API events when our code do something like:
```
item.setCrop(crop);
item.setPosition(position);
item.setVisible(visible);
// ^ 3 sceneItemChanged will be generated here!
```

So I created an interface 
```
ISceneItemSettings {
  x: number;
  y: number;
  scaleX: number;
  scaleY: number;
  visible: boolean;
  crop: ICrop;
  locked: boolean;
  rotation: number;
}
```
It's almost the same as ITransform but it also contains `locked` and `visible` values that I think the undo system can also store. I think I can change ISceneItemSettings to something like:
```
ISceneItemSettings {
  transform: ITransform,
  visible: boolean;
  locked: boolean;
}
```

I think it's a good idea to have a Transform class instead of `Rectalnge`  and `ScalableRectangle` that also can provide such things as `width` and `height`. For example, in Unity3d each object has `transform` property with `position` and `rotation` fields that manage the all Math with transforms.

andycreeth: I think it's important that `width` and `height` are kept out of the Transform.  Width and height are properties of the underlying source, whereas a Transform can be applied to each scene item separately.

holiber:
I rewrite SceneItem to use `ITransform`. It's a lot of changes in many files. We should hardly test this PR before release

holiber:
I created "Experimental" section that available in dev-mode and put multiselection there
image

PR #231, opened 2018-02-02
upgrade osn

andycreeth:


andycreeth:
Well, this doesn't compile at all

PR #256, opened 2018-02-09
Rename default sources

holiber:
Since we have id based source system instead name based, the `source.displayName` seems pretty unless and I removed it. Also, the new OBS release allows to rename default sources, so I added this feature to this PR. As a bonus, if users are going to choose some non-default device as the default source SLOBS will use the device.
The one downside of this PR users that already have set up default audio sources will get other names for them. For example `DesktopAudioDevice1` instead `Desktop Audio`. And I'm not sure is it worth to make a migration for that,

I decided to make this PR now until guys from Elgato didn't fully implement the streamdeck support. They are working on the default audio sources now.
image

### DIFF
--- a/app/services/scene-collections/nodes/sources.ts
+++ b/app/services/scene-collections/nodes/sources.ts
@@ -249,17 +249,6 @@ export class SourcesNode extends Node<ISchema, {}> {
 
       this.data.items.forEach(source => {
 
-        if (source.name === 'AuxAudioDevice1') {
-          source.name =  'Mic/Aux';
-          return;
-        }
-
-        if (source.name === 'DesktopAudioDevice1') {
-          source.name = 'Desktop Audio';
-          return;
-
-        }
-
         const desktopDeviceMatch = /^DesktopAudioDevice(\d)$/.exec(source.name);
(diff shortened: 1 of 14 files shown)


andycreeth:
Unfortunately it appears that when existing users receive this update, their audio devices will be called:
"DesktopAudioDevice1"
"AuxAudioDevice"
until they remove and re-add them manually.

I think we will need to come up with a better way of migrating existing scene collections.

app/services/scene-collections/nodes/sources.ts
```diff
@@ -240,4 +240,42 @@ export class SourcesNode extends Node<ISchema, {}> {
...
+        if (source.name === 'AuxAudioDevice1') {
+          source.name =  'Mic/Aux';
+          return;
+        }
+
+        if (source.name === 'DesktopAudioDevice1') {
+          source.name = 'Desktop Audio';
+          return;
+
+        }
```
andycreeth: Are these needed with the more general Regex matchers below?

holiber: Good catch. I just copied this code from the removed `dispayName` getter.

PR #258, opened 2018-02-10
fix selection

holiber:


app/services/selection/selection.ts
```diff
@@ -135,7 +135,6 @@ export class SelectionService
 
   selectAll() {
     this.select(this.getScene().getItems().map(item => item.sceneItemId));
-    return this;
```
holiber: API doesn't work correctly with unserializable values like that

PR #262, opened 2018-02-12
Simplify backtrace use case

computerquip-streamlabs:
The reason for this is to prevent the surface area for backtrace-node to cause issues. The dialog is nice but it doesn't seem to treat it too well and distorts the report being sent.

### DIFF
--- a/main.js
+++ b/main.js
@@ -17,25 +17,8 @@ const bt = require('backtrace-node');
 
 app.disableHardwareAcceleration();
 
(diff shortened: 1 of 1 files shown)


andycreeth:
Yeah I'm mixed about this.  Without the dialog, we just show the user a big JS stack trace.  Which honestly looks kind of unprofessional.

andycreeth:
Going to close this, since I don't think we want to do this at the moment.

PR #268, opened 2018-02-13
hack to prevent slider width issue

andycreeth:


app/components/shared/Slider.vue.ts
```diff
@@ -1,42 +1,31 @@
...
+  @Prop() sliderStyle: object;
+  @Prop() usePercentages: boolean;
+
+  $refs: { slider: any };
+
+  mounted() {
+    // Hack to prevent transitions from messing up slider width
+    setTimeout(() => {
+      if (this.$refs.slider) this.$refs.slider.refresh();
+    }, 500);
```
holiber: don't see any clearTimeout, it's a possible memory leak here

andycreeth: It's just a timeout so it only fires once.  Instead of adding a `clearTimeout` I just added the `if` statement to check if the slider is still there.  I don't think this should be a leak.

holiber: Oh, my bad. Somehow I have seen a `setInterval` here

PR #271, opened 2018-02-14
Add YouTube Features

kuhasuki:
- Add a link to your stream in live dock
- Add a link to live dashboard in live dock
- Allow user to change title and description for a YouTube stream in progress
- Fetch donation_url for user and attach to description when enabled by user
- Log users into YouTube chat automatically

### DIFF
--- a/app/components/Chat.vue.ts
+++ b/app/components/Chat.vue.ts
@@ -8,6 +8,7 @@ import { CustomizationService } from 'services/customization';
 import url from 'url';
 import electron from 'electron';
 import { ICustomizationSettings } from 'services/customization/customization-api';
+import { YoutubeService } from 'services/platforms/youtube';
 
 @Component({})
 export default class Chat extends Vue {
@@ -29,7 +30,18 @@ export default class Chat extends Vue {
     const webview = this.$refs.chat;
     const settings = this.customizationService.getSettings();
 
-    service.getChatUrl(nightMode).then(chatUrl => this.chatUrl = chatUrl);
+    if (service instanceof YoutubeService) {
+      service.getChatUrl(nightMode).then(chatUrl => {
+        this.chatUrl = `https://youtube.com/signin`;
+
+        webview.addEventListener('did-navigate', () => {
+          this.chatUrl = chatUrl;
+        });
+      });
+    } else {
+      service.getChatUrl(nightMode).then(chatUrl => this.chatUrl = chatUrl);
+    }
+
 
     webview.addEventListener('new-window', e => {
       const protocol = url.parse(e.url).protocol;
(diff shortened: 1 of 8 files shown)


app/components/LiveDock.vue
```diff
@@ -48,8 +48,12 @@
       </div>
 
       <div class="live-dock-info">
-        <a @click="showEditStreamInfo" v-if="isTwitch">Edit Stream Info</a>
-        <a @click="refreshChat">Refresh Chat</a>
+        <div class="platform-tools">
```
andycreeth: Let's make this class `live-dock-platform-tools` just to match everything else in this component and ensure our class names are sufficiently unique.

app/services/platforms/youtube.ts
```diff
@@ -57,6 +57,13 @@ export class YoutubeService extends Service implements IPlatformService {
     });
   }
 
+  fetchDescriptionString(): Promise<string> {
```
andycreeth: I would call this `fetchDescription`.  The nice thing about typescript is you already know it returns a string.

app/services/platforms/youtube.ts
```diff
@@ -150,25 +157,28 @@ export class YoutubeService extends Service implements IPlatformService {
...
-        headers,
-        body: JSON.stringify(data)
-      }
-    );
+  putChannelInfo(streamTitle: string, streamDescription: string): Promise<boolean> {
+    return this.fetchDescriptionString().then(autopublishString => {
+      const headers = new Headers();
+      headers.append('Content-Type', 'application/json');
+
+      const fullDescription = autopublishString.concat(streamDescription);
```
andycreeth: Is a simple concatenation enough?  Should we be adding a newline or something?  Not really sure.

kuhasuki: The autopublish string has a newline in it if enabled.

app/services/platforms/youtube.ts
kuhasuki: Good Point

app/components/LiveDock.vue
kuhasuki: fixed 3d9cce1c30c4e8096068868317ba472e91a5cf4a

app/components/Chat.vue.ts
```diff
@@ -29,7 +30,18 @@ export default class Chat extends Vue {
     const webview = this.$refs.chat;
     const settings = this.customizationService.getSettings();
 
-    service.getChatUrl(nightMode).then(chatUrl => this.chatUrl = chatUrl);
+    if (service instanceof YoutubeService) {
+      service.getChatUrl(nightMode).then(chatUrl => {
+        this.chatUrl = 'https://youtube.com/signin';
```
andycreeth: If the user is already logged in, this should immediately redirect, right?

kuhasuki: Yup

PR #274, opened 2018-02-14
Source groupping 1

holiber:
I have plans to split the source groping feature in several PRs. This is the first one.
* Added `Selection` class to expose API to manage multiple sceneItems (still not sure about naming)
* Group Scene Items into a Scene (have to be enabled experimental settings)

### DIFF
--- a/app/components/windows/NameScene.vue.ts
+++ b/app/components/windows/NameScene.vue.ts
@@ -36,12 +36,19 @@ export default class NameScene extends Vue {
   } = this.windowsService.getChildWindowQueryParams();
 
   mounted() {
-    this.name = this.options.rename ?
-      this.options.rename :
-      this.sourcesService.suggestName(
-        this.options.sceneToDuplicate ||
-        this.options.itemsToGroup ? 'Group' : 'NewScene'
-      );
+    let name = '';
+
+    if (this.options.rename) {
+      name = this.options.rename;
+    } else if (this.options.sceneToDuplicate) {
+      name = this.options.sceneToDuplicate;
+    } else if (this.options.itemsToGroup) {
+      name = `${this.scenesService.activeScene.name} Group`;
+    } else {
+      name = 'New Scene';
+    }
+
+    this.name = this.sourcesService.suggestName(name);
   }
 
   submit() {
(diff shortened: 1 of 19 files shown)


app/components/windows/NameScene.vue.ts
```diff
@@ -25,28 +26,47 @@ export default class NameScene extends Vue {
...
+    itemsToGroup?: string[]
+  } = this.windowsService.getChildWindowQueryParams();
 
   mounted() {
     this.name = this.options.rename ?
       this.options.rename :
-      this.sourcesService.suggestName(this.options.sceneToDuplicate || 'NewScene');
+      this.sourcesService.suggestName(
+        this.options.sceneToDuplicate ||
+        this.options.itemsToGroup ? 'Group' : 'NewScene'
```
andycreeth: Just a minor suggestion, but it might be cool to name this dynamically based on the name of one of the scene items they are grouping.  For example, the pseudocode might look like:
```
const firstSceneItemName = sceneItemsToGroup[0].name;
const name = this.suggestName(`${firstSceneItemName} Group`);
```

andycreeth: Also for the new scene case it should be "New Scene" (with a space in between)

app/services/utils.ts
```diff
@@ -92,4 +92,25 @@ export default class Utils {
     });
     return result as Partial<T>;
   }
+
+  /**
+   * @see https://www.typescriptlang.org/docs/handbook/mixins.html
+   */
+  static applyMixins(derivedCtor: any, baseCtors: any[]) {
```
andycreeth: Nice.  I've been wanting to start utilizing mixins more.

app/services/selection/selection.ts
```diff
@@ -43,26 +41,74 @@ export class SelectionService
...
+  getIds: () => string[];
+  getInvertedIds: () => string[];
+  getInverted: () => SceneItem[];
+  getBoundingRect: () => IRectangle;
+  getLastSelected: () => SceneItem;
+  getSize: () => number;
+  isSelected: (item: string | ISceneItem) => boolean;
+  copyReferenceTo: (sceneId: string) => SceneItem[];
+  copyTo: (sceneId: string) => SceneItem[];
+  moveTo: (sceneId: string) => SceneItem[];
```
andycreeth: What's the purpose of declaring all these methods here?  They're defined below, and are all declared additionally in `ISelection`.

holiber: SelectionService use Selection class as a mixin. This is the way how typescript documentation recommends to use mixins. Unfortunately, it requires this additional boilerplate code. https://www.typescriptlang.org/docs/handbook/mixins.html

andycreeth: Ah, got it.

PR #296, opened 2018-02-21
Screentest fixes

holiber:
- failed sceen-tests now makes the build failed
- fixed some stupid bug in the source-props tests


holiber:
The build failed because the screen-tests in staging branch are broken

PR #304, opened 2018-02-28
Update facemask to v0.6.2

eddyStreamlabs:


crowhound:
The title of the PR says Update to facemask version 0.6.2, but in the commit it shows the code being changed to 0.6.3. Is the PR title wrong or the code wrong. Just curious if the code was meant to be 0.6.2 instead of 0.6.3 and if that might cause some issues.

PR #308, opened 2018-03-01
webpack 4

andycreeth:
- Upgrade to webpack 4
- Upgrade loaders to work with webpack 4
- Use `NODE_ENV` in renderer processes provided by webpack rather than fetching it from the main process

Remaining todo items:
- Use webpack for `main.js` so we can get rid of electron-builder injected environment for the main process
- Fix minification


webpack.config.js
```diff
@@ -108,11 +108,15 @@ module.exports = {
...
           outputPath: 'fonts/',
-          publicPath: 'bundles/'
+          publicPath: 'bundles/fonts/'
         }
       }
     ]
   },
 
+  optimization: {
+    minimize: false
```
andycreeth: By default, production webpack builds now run UglifyJS`.  Unfortunately, Uglify is failing on our `updater.js` entry with an extremely helpful error message of `undefined`.  I'll look into it later, but minification is of relatively limited value to us since we're not trying obfuscate code, and we're not loading it over the web on page load.

PR #311, opened 2018-03-02
Add Mixer Support

kuhasuki:
**Mixer**
Enable log in with Mixer
Enable streaming to Mixer (defaults to FTL)
Allow user to set stream title and game
Embed Mixer chat 
Show mixer viewer count
Add test widgets for Mixer

**Misc**
Fix bug with list input (internal search was previously always on)
Refactor requiresToken decorator


andycreeth:
You have a change to the libuiohook submodule in here.  I'm assuming that was a mistake.

PR #313, opened 2018-03-04
Added Deinterlacing to Streamlabs OBS for retro consoles

Alex-Hosein:
In OBS Studio we had the option to deinterlace video without needing to tweak our drivers. Just re-implementing that into Streamlabs OBS

### DIFF
--- a/app/services/scenes/scene-item.ts
+++ b/app/services/scenes/scene-item.ts
@@ -289,6 +289,20 @@ export class SceneItem implements ISceneItemApi {
     });
   }
 
+
+
+  deinterlaceOn(){
+  const source = this.sourcesService.getSource(this.sourceId).getObsInput();
+  source.deinterlaceMode = 1;
+  source.deinterlaceFieldOrder = 1;
+}
+
+deinterlaceOff(){
+  const source = this.sourcesService.getSource(this.sourceId).getObsInput();
+  source.deinterlaceMode = 0;
+  source.deinterlaceFieldOrder = 0;
+}
+
   /**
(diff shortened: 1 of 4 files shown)


computerquip-streamlabs:
Hi @Alex-Hosein, I'm reluctant to pull this in for a couple of reasons: 

- You should use the enumeration values rather than literals to express the mode you want.

- All of the available modes should be exposed here. The modes available are exposed in the bindings as an enumeration here: https://github.com/stream-labs/obs-studio-node/blob/staging/js/module.ts#L48

The current mode you force things to would be the "Discard" mode which I don't think is what you want. 

- There should be a capability to change the field order separately. Also 0 is a valid field order which means Top (or upper field order), exposed as an enumeration here: https://github.com/stream-labs/obs-studio-node/blob/staging/js/module.ts#L43

Thanks, Zach

computerquip-streamlabs:
I'm going to also add that deinterlacing doesn't work with all sources. It only works with Async sources. 

The enumeration value for that is here: https://github.com/stream-labs/obs-studio-node/blob/staging/js/module.ts#L146

The flag to fetch the capabilities of the output of a source can be fetched here by using `outputFlags` and checking each flag with a bitwise AND: https://github.com/stream-labs/obs-studio-node/blob/staging/js/module.ts#L1192

andycreeth:
Hey @Alex-Hosein thanks for your contribution.  I'm going to close this for now per @computerquip-streamlabs comments.  Feel free to open another PR in the future with the above requested changes and we can try to get it integrated.

PR #315, opened 2018-03-05
TCP server is always ON now

holiber:


andycreeth:
You committed node-libuiohook changes

andycreeth:
Also, given that we're forcing this onto a specific port, let's also force `allowRemote: false`

PR #316, opened 2018-03-06
save only non-default settings to the local storage

holiber:
These changes allow us to change the default settings if the user has not changed it before

### DIFF
--- a/app/services/persistent-stateful-service.ts
+++ b/app/services/persistent-stateful-service.ts
@@ -1,4 +1,5 @@
 import { StatefulService } from './stateful-service';
+import Utils from './utils';
 
 // Extends StatefulService with code that will persist the
 // state across executions of the application.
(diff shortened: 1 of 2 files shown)


app/services/persistent-stateful-service.ts
```diff
@@ -25,10 +26,13 @@ export abstract class PersistentStatefulService<TState extends object> extends S
   init() {
     this.store.watch(
       () => {
-        return JSON.stringify(this.state);
+        return this.state;
```
andycreeth: This change actually completely breaks the persistent stateful service, and it will no longer correctly persist your changes.  Vuex watchers are actually really poorly documented, but essentially this first function acts as a change detector.  It will run whenever the store changes, and will compare the return value to the last time it ran.  It does this with a simple `===` comparison.  Since `this.state` will be a reference to the same object, your bottom function will never get run.  This will result in changes to persistent stateful service state never actually getting persisted.

holiber: Good to know, thanks

andycreeth:
I'm closing this until we can come up with a solution that doesn't break existing installations.

PR #331, opened 2018-03-12
FrankerFaceZ support for chat

jamacanbacn:
Adds support for the FrankerFaceZ chat extension as an alternative to BTTV. They should also both work together.


app/components/Chat.vue.ts
```diff
@@ -71,6 +71,30 @@ export default class Chat extends Vue {
         `, true);
       }
     });
+    webview.addEventListener('new-window', e => {
+      const protocol = url.parse(e.url).protocol;
+
+      if (protocol === 'http:' || protocol === 'https:') {
+        electron.remote.shell.openExternal(e.url);
+      }
+    });
```
holiber: duplicated code here. You have the same piece of code above

app/components/Chat.vue.ts
```diff
@@ -71,6 +71,30 @@ export default class Chat extends Vue {
...
+          
+          localStorage.setItem('ffz_clickTwitchEmotes', true);
+          localStorage.setItem('ffz_darkenedMode', ${ settings.nightMode ? 'true' : 'false' });
+          
+          var ffzscript1 = document.createElement('script');
+          ffzscript1.setAttribute('src','https://cdn.frankerfacez.com/script/script.min.js');
+          document.head.appendChild(ffzscript1);
+        `, true);
+      }
+    });
```
holiber: We already have `dom-ready` listener above. Don't create listeners when you can reuse a created one. Your code should looks like:
```
if (this.isTwitch) {

  if (settings.enableBTTVEmotes) {
    // do stuff with BBTV....
  }

  if (settings.enableFFZEmotes) {
        // do stuff with FFZ....
  }


}


```
Did you check that both plugins work well? If they dont, you need to replace checkboxes in the settings to listbox or radiobuttons

andycreeth:
Nice.  Maybe we should disable screen tests until we have time to fix them.  They seem to be breaking the build more often than not at the moment.

PR #332, opened 2018-03-12
Output state refactor

andycreeth:


andycreeth:
TODO:
- [ ] Fix tests
- [ ] Use an actual release of OSN
- [ ] Wait until recording is fixed on the backend

test/api/streaming.ts
```diff
@@ -1,26 +1,51 @@
...
+    t.pass();
+    return;
+  }
+
   const client = await getClient();
   const streamingService = client.getResource<IStreamingServiceApi>('StreamingService');
-  let streamingState: IStreamingServiceState;
+  const settingsService = client.getResource<ISettingsServiceApi>('SettingsService');
+
+  const streamSettings = await settingsService.getSettingsFormData('Stream');
```
holiber: No need `await` here. All API methods are synchronous like child window IPC requests

test/api/streaming.ts
```diff
@@ -1,26 +1,51 @@
...
+  streamSettings.forEach(subcategory => {
+    subcategory.parameters.forEach(setting => {
+      if (setting.name === 'service') setting.value = 'Twitch';
+      if (setting.name === 'key') setting.value = process.env.SLOBS_TEST_STREAM_KEY;
+    });
+  });
+  await settingsService.setSettings('Stream', streamSettings);
+
+  let streamingStatus = await streamingService.getModel().streamingStatus;
```
holiber: no `await`

app/services/streaming/streaming.ts
```diff
@@ -1,261 +1,221 @@
...
 
-  @mutation()
-  START_STREAMING(startTime: string, usingDelay: boolean) {
-    this.state.isStreaming = true;
-    this.state.streamStartTime = startTime;
-    this.state.streamEndTime = null;
-    this.state.usingDelay = usingDelay;
-  }
+  // Dummy subscription for stream deck
+  streamingStateChange = new Subject<void>();
```
holiber: If we don't emit this event now, is it just for preventing errors with streamdeck?

app/services/streaming/streaming.ts
```diff
@@ -267,34 +227,79 @@ export class StreamingService extends StatefulService<IStreamingServiceState> im
...
+      const time = new Date().toISOString();
+
+      if (info.signal === EOBSOutputSignal.Start) {
+        this.SET_RECORDING_STATUS(ERecordingState.Recording, time);
+      } else if (info.signal === EOBSOutputSignal.Starting) {
+        this.SET_RECORDING_STATUS(ERecordingState.Starting, time);
+      } else if (info.signal === EOBSOutputSignal.Stop) {
+        this.SET_RECORDING_STATUS(ERecordingState.Offline, time);
+      } else if (info.signal === EOBSOutputSignal.Stopping) {
+        this.SET_RECORDING_STATUS(ERecordingState.Stopping, time);
```
holiber: Don't we expose recording events for Streamdeck?

app/services/streaming/streaming.ts
andycreeth: Yeah basically.  I think stream deck probably ignores the error, but I wanted to make sure there weren't any issues with subscribing.  It turns out they're not actually using the event for anything anyway.

app/services/streaming/streaming.ts
andycreeth: We did, but they don't appear to be using it.  It might be worth adding back in though just to match the streaming output.

PR #333, opened 2018-03-14
Source grouping part 4

holiber:
The tree-view component page https://github.com/stream-labs/sl-vue-tree


app/components/SourceSelector.vue
```diff
@@ -18,27 +18,61 @@
...
-        @click.stop="toggleVisibility(props.item.value)"
-        @dblclick.stop="() => {}" />
-    </template>
-  </selector>
+
+    <sl-vue-tree
+        :value="nodes"
+        @select="makeActive"
+        @drop="handleSort"
+        @toggle="toggleFolder"
```
andycreeth: Looks like these are being done with 4 space indents.  I'm not sure if your IDE is doing that, but I think we've generally been doing normal 2 space indents for this situation.

andycreeth:
Some feedback on playing around with this branch:
- Sources don't seem to have `cursor: pointer` anymore
- Dragging to recorder sources when scrolling is required seems to be a little difficult.  I think the area you have to hover over to get it to scroll is maybe I little smaller or something.  I was struggling a bit with it.
- As much as I hate to say it, we should consider adding a bulk IPC method for moving multiple scene items.  The IPC overhead of dragging a lot of items at once is a little too high I think.
- Right clicking seems to only work when right clicking on the text.  I am able to left click anyway on the "source bar" to select it, so I think right clicking should have the same target area.
- When ungrouping a nested folder, it should add them back to their parent folder, instead of the top level.
- When converting a folder into a scene group, the naming window does not close (likely due to an exception).
- I think we should make folders more discoverable.  I wonder if we should add a button to the top of the source selector that creates a new empty folder.  Then you can drag sources into the folder.  As far as I can tell, your code seems to support empty folders, so hopefully this won't be too hard to do.

I will likely have more feedback as I continue to play around with this but I wanted to get my initial feedback on here.  By the way, this is really awesome and I'm super excited to release this.  Nice work!

holiber:
Thanks, for review

andycreeth:
image
I'm getting this when running your branch.

andycreeth:
Also you have node-libuiohook changes in this PR that you need to revert.

andycreeth:
Were we going to add that button above sources to create a new folder?  Was @mbiemiller working on that, or were you doing that?

app/components/SourceSelector.vue.ts
```diff
@@ -56,27 +84,56 @@ export default class SourceSelector extends Vue {
...
+    // const treeContaier = this.$refs.treeContainer;
+    // this.scrollIntervalId = window.setInterval(() => {
+    //   if (!slVueTree.draggingNode) {
+    //     this.stopScroll();
+    //     return;
+    //   }
+    //   const deltaScroll = direction === 'top' ? -2 : +2;
+    //   treeContaier.scrollTop += deltaScroll;
+    // }, 500);
+  }
```
andycreeth: I take it this is the old scrolls system, or an experiment?

holiber: Looking at your screenshot I could say I forgot to push the last commit. Now it should be ok. Morgan said that she is busy with some anther task now, so I added a day mode and i'm gonna to add a "add folder" button. Unfortunately, Fontawesome doesn't have standart "add folder" icon, so I have to figure out something with it.

holiber:
Changes added:
- "Add folder" button now groups items into the folder if more than one item is selected
- Actions like "lock" and "hide" is not shown for folders without items inside.
- Empty folders don't have toggle arrow

PR #335, opened 2018-03-14
Remove browser-source from the install-plugins script

eddyStreamlabs:
obs-browser is now part of obs-studio-node


andycreeth:
We should also remove it from the slobs repo.  No sense making people download a 50MB zip file they are never going to need.

PR #340, opened 2018-03-17
Shrink PNG image space by 22%

chevycam94:
Just optimized the size of PNG files to take up less disk space.  Dimensions, alpha, etc, are all still in tact, just as they should be.


andycreeth:
Hey, thanks for your contribution!  Sorry about the spam, our CI server is having some issues right now and we are looking into it.

chevycam94:
Not a problem.  As I dissect more about the app, I'll try to make some more changes.  But this is a good start.  :D

PR #343, opened 2018-03-19
Tooltips

mbiemiller:
added tooltips throughout the app


app/components/LiveDock.vue
```diff
@@ -49,9 +49,25 @@
 
       <div class="live-dock-info">
         <div class="live-dock-platform-tools">
-          <a title="Edit Stream Info" @click="showEditStreamInfo" v-if="isTwitch || isMixer || (isYoutube && isStreaming)"><i class="fa fa-pencil" /></a>
-          <a title="View Stream" @click="openYoutubeStreamUrl" v-if="isYoutube && isStreaming"><i class="fa fa-video-camera" /></a>
-          <a title="Live Dashboard" @click="openYoutubeControlRoom" v-if="isYoutube && isStreaming"><i class="fa fa-cogs" /></a>
+          <!-- STAGING -->
```
andycreeth: I'm not sure I understand what this comment means.

app/components/LiveDock.vue.ts
```diff
@@ -10,7 +10,10 @@ import Slider from './shared/Slider.vue';
 import electron from 'electron';
 import { getPlatformService } from 'services/platforms';
 import { YoutubeService } from 'services/platforms/youtube';
+import VTooltip from 'v-tooltip';
 
+Vue.use(VTooltip);
+VTooltip.options.defaultContainer = '#mainWrapper';
```
andycreeth: I think maybe we should only do these 2 lines once, instead of in every component that uses them.  Maybe we should add it in `app/app.ts`.

PR #344, opened 2018-03-19
Await on obsImporterService that was causing an exception in some cases

eddyStreamlabs:


andycreeth:
Going to merge this since it only failed due to not having my CI fixes.

PR #370, opened 2018-03-28
folders bugfixing

holiber:
fixes for some bugs and sentry errors


yarn.lock
```diff
@@ -8173,7 +8173,7 @@ single-line-log@^1.1.2:
 
 "sl-vue-tree@https://github.com/stream-labs/sl-vue-tree.git":
   version "1.0.0"
-  resolved "https://github.com/stream-labs/sl-vue-tree.git#cc9dc0c143d57d3c8391635e89681867bd9b4c31"
+  resolved "https://github.com/stream-labs/sl-vue-tree.git#b1e8edf41ac1c333d3dfbf031c1aa241905ee03d"
```
andycreeth: I'm curious why the checksum for this package changed when the version number stayed the same.

holiber: I updated the plugin without changing the version. Not sure is it worth to change version while only one projects uses this plugin.

andycreeth: Ah ok got it

PR #371, opened 2018-03-29
Patch notes in SLOBS

andycreeth:
We now have the ability to show patch notes inside SLOBS.

### DIFF
--- a/app/services/patch-notes/index.ts
+++ b/app/services/patch-notes/index.ts
@@ -1,0 +1,23 @@
+import { PersistentStatefulService } from 'services/persistent-stateful-service';
+
+interface IPatchNotesState {
+  lastVersionSeen: string;
+}
+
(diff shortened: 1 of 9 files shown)


app/services/patch-notes/index.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (Util.isPreview()) return;
+
+    const currentVersion = electron.remote.process.env.SLOBS_VERSION;
+
+    // If the notes don't match the current version, we shouldn't
+    // show them.
+    if (notes.version !== currentVersion) return;
+
+    // The user has already seen the current patch notes
+    if (currentVersion === this.state.lastVersionSeen) return;
```
holiber: we already have a DismissablesService to hide information that user has already seen. Do you think here is some another case?

PR #375, opened 2018-04-02
fix copy/paste order and add tests

holiber:


app/services/clipboard/clipboard.ts
```diff
@@ -1,11 +1,12 @@
...
-import { ScenesService } from './scenes';
-import { SourcesService } from './sources';
-import { shortcut } from './shortcuts';
-import { Inject } from '../util/injector';
-import { SourceFiltersService } from './source-filters';
+import { mutation, StatefulService } from 'services/stateful-service';
+import { ScenesService } from 'services/scenes';
+import { SourcesService } from 'services/sources';
+import { shortcut } from 'services/shortcuts';
+import { Inject } from '../../util/injector';
```
andycreeth: Webpack will recognize any folder name in the `app` directory, so this can alternatively be written as `util/injector`

holiber: good to know, thanks!

PR #377, opened 2018-04-03
Hide Optimized Encoding for Mixer

jamacanbacn:
Mixer's FTL does not support Bframes from our optimized encoder so this hides the checkbox for mixer users.


app/components/windows/EditStreamInfo.vue
```diff
@@ -27,7 +27,7 @@
         @search-change="debouncedGameSearch"
         @input="onGameInput"/>
       <div v-if="areAvailableProfiles">
-        <div class="input-container">
+        <div class="input-container" v-if="isTwitch || isYoutube">
```
andycreeth: FYI I don't think it's even possible to use optimized encoder settings on Youtube right now since this window doesn't show up if you are streaming to Youtube.  I believe the reason is that Youtube doesn't support editing stream info before going live unless you set it up as a scheduled stream.  @kuhasuki would know more than I.

PR #379, opened 2018-04-04
Start moving to new OAuth token headers for Streamlabs APIs

andycreeth:
TODO:
- [ ] Move over usage events when supported
- [ ] Move over scene collection APIs when supported


andycreeth:
Youtube token refresh is now tested and fixed on the server side.  Testing mixer next.

andycreeth:
Mixer token refresh is working

PR #382, opened 2018-04-06
add a valid remotecontrol URL

holiber:


holiber:
Also the qrcode size has been increaced

PR #383, opened 2018-04-09
refresh twitch token when expired

andycreeth:
We really shouldn't rely on this having a really long expiration.


andycreeth:
Darn looks like this broken the streaming test

PR #387, opened 2018-04-10
documentation for source folders

holiber:
Also, some API methods for consistency added.


holiber:
Thanks for corrections 👍

PR #391, opened 2018-04-11
prevent text selection

andycreeth:


holiber:
If it works fine now, I suggest to return the shortcut "Ctrl+A" in `SelectionService`

PR #398, opened 2018-04-16
Add filters to a scene

holiber:
Since we keep all scenes in the `SourcesService` that is pretty easy

### DIFF
--- a/app/components/SceneSelector.vue.ts
+++ b/app/components/SceneSelector.vue.ts
@@ -2,16 +2,16 @@ import Vue from 'vue';
 import { Component } from 'vue-property-decorator';
 import { Inject } from '../util/injector';
 import Selector from './Selector.vue';
-import { ScenesService } from '../services/scenes';
+import { ScenesService } from 'services/scenes';
 import { Menu } from '../util/menus/Menu';
-import { ScenesTransitionsService } from '../services/scenes-transitions';
-import { SceneCollectionsService } from '../services/scene-collections';
-import { AppService } from '../services/app';
+import { ScenesTransitionsService } from 'services/scenes-transitions';
+import { SceneCollectionsService } from 'services/scene-collections';
+import { AppService } from 'services/app';
 import DropdownMenu from './shared/DropdownMenu.vue';
 import HelpTip from './shared/HelpTip.vue';
 import { EDismissable } from 'services/dismissables';
 import Fuse from 'fuse.js';
-import { SourceFiltersService } from "../services/source-filters";
+import { SourceFiltersService } from 'services/source-filters';
 
 @Component({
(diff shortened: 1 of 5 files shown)


app/components/SceneSelector.vue.ts
```diff
@@ -11,6 +11,7 @@ import DropdownMenu from './shared/DropdownMenu.vue';
 import HelpTip from './shared/HelpTip.vue';
 import { EDismissable } from 'services/dismissables';
 import Fuse from 'fuse.js';
+import { SourceFiltersService } from "../services/source-filters";
```
andycreeth: Use `"services/source-filters"`

holiber:
Oh, didn't know that. Just took this task from Asana.

app/services/scene-collections/nodes/scene-filters.ts
```diff
@@ -0,0 +1,32 @@
+import { ArrayNode } from './array-node';
+import { Inject } from '../../../util/injector';
+import { ISourceFilter, SourceFiltersService } from 'services/source-filters';
+
+
+interface IContext {
+  sceneId: string;
+}
+
+export class SceneFiltersNode extends ArrayNode<ISourceFilter, IContext, ISourceFilter> {
```
andycreeth: I would like us to never use an outside interface as the `TSchema` for any nodes in the config system.  The reason being is that the `TSchema` interface is responsible for enforcing the schema for that particular node of the persistence system.  If it is an outside interface, then it is far too easy to accidentally modify the configuration schema without realizing it.

This is probably best illustrated with an example.  Let's say that an engineer comes along and starts working on source filters, and adds a new mandatory field to `ISourceFilter`.  Because this class uses `ISourceFilter` without accessing its contents directory, you won't receive any type errors in this file, and you would stand a good chance of not seeing anything out of the ordinary in your dev environment.  However, if you were to load up an old config file, it would fail because that mandatory field would be missing.

Now instead imagine that you were using an interface in this file that represented the serialized version of that interface, and this class was responsible for translating that into / out of an `ISourceFilter`.  When you made the change to `ISourceFilter`, you would immediately get type errors in this file, and it would indicate that you need handle a config migration to support old config file.

So, as a rule, I think that all nodes in the config should define their own schema in the same file.  This results in more boilerplate code, but I think it is a clear situation where trying to be too DRY ends up breaking things.

andycreeth: I would add that in this case, I think it would be OK to share the same interface `IFilterInfo` that is used by the sources node, since that is part of the same config file.

holiber: That makes sense. 👍

PR #401, opened 2018-04-17
recommend adding by reference for widgets of the same type

andycreeth:
Changes:
- When there is already an existing widget of the same type, show the "Add Source" screen instead of the "Name Source" screen
- Consolidate logic that is used to determine if an existing source of that type exists
- Add a recommended label for adding by reference when adding a new widget

### DIFF
--- a/app/services/sources/source.ts
+++ b/app/services/sources/source.ts
@@ -5,14 +5,17 @@ import {
   ISource,
   SourcesService,
   TPropertiesManager,
+  ISourceComparison,
   PROPERTIES_MANAGER_TYPES
 } from './index';
-import { mutation, ServiceHelper } from '../stateful-service';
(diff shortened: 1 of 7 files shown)


app/services/sources/source.ts
```diff
@@ -51,6 +54,30 @@ export class Source implements ISourceApi {
     return this.getObsInput().settings;
   }
 
+  /**
+   * Compares the details of this source to another, to determine
+   * whether adding as a reference makes sense.
+   * @param comparison the comparison details of the other source
+   */
+  isSameType(comparison: ISourceComparison): boolean {
+    if (this.channel) return;
```
holiber: Since we expose this function to public API the returning type should be consistent with the method definition. Because we don't know how other clients parse the API responses. So should be `return false` here instead `return`

PR #407, opened 2018-04-18
Copy/paste between collections

holiber:


app/services/scene-collections/scene-collections-api.ts
```diff
@@ -56,7 +56,6 @@ export interface ISceneCollectionsServiceApi {
 }
 
 export interface ISceneCollectionCreateOptions {
-  setupFunction?: () => boolean;
```
holiber: we can't expose unserializable params to public API

PR #413, opened 2018-04-20
cleanup invalid imports

holiber:
I rid of `declare module '*'` because of it allowed to have some invalid imports in the project

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -157,6 +157,10 @@
   version "9.1.8"
   resolved "https://registry.yarnpkg.com/@types/highlight.js/-/highlight.js-9.1.8.tgz#d227f18bcb8f3f187e16965f2444859a04689758"
 
+"@types/jquery@*":
+  version "3.3.1"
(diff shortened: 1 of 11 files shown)


yarn.lock
```diff
@@ -199,6 +203,12 @@
   version "1.4.31"
   resolved "https://registry.yarnpkg.com/@types/socket.io-client/-/socket.io-client-1.4.31.tgz#86b8986409384a9ee1a74993a70822c36114dc79"
 
+"@types/urijs@^1.15.37":
+  version "1.15.37"
+  resolved "https://registry.yarnpkg.com/@types/urijs/-/urijs-1.15.37.tgz#42cd9d5af585053e9af5d009d77bfeb6b0b4cfe6"
+  dependencies:
+    "@types/jquery" "*"
```
andycreeth: What on earth is this doing here.  I checked... urijs definitely doesn't depend on jQuery.  And if it did, I would be removing it tomorrow morning haha.

holiber: Looks like `urijs` has a jQuery-plugin version.

PR #414, opened 2018-04-23
Refactor Streamlabels Service Create Websocket Service

kuhasuki:
Decouples websocket listener from streamlabels service.
Emits socket events to RXJs Subject allowing multiple handlers for a single socket event.


app/services/streamlabels/index.ts
```diff
@@ -3,7 +3,8 @@ import { UserService } from 'services/user';
 import { Inject } from 'util/injector';
 import { HostsService } from 'services/hosts';
 import { handleErrors, authorizedHeaders } from 'util/requests';
-import io from 'socket.io-client';
+import { WebsocketService, TSocketEvent } from 'services/websocket';
+// import io from 'socket.io-client';
```
andycreeth: This can be removed now, right?

app/services/websocket.ts
```diff
@@ -0,0 +1,96 @@
+import { Service } from './service';
+import { Inject } from '../util/injector';
```
andycreeth: Can be `util/injector`

app/services/websocket.ts
```diff
@@ -0,0 +1,96 @@
...
+import { UserService } from './user';
```
andycreeth: Should be `services/user`

app/services/streamlabels/index.ts
kuhasuki: Yeah

PR #418, opened 2018-04-24
fix item z index issues

andycreeth:
The crux of this issue is that when calling `placeAfter` where the `destNode` is a folder, we weren't always taking into account all items in the folder when calculating the new index.

This PR fixes that issue, and also simplifies the logic.  Rather that handling the complex logic of moving nodes twice (once to move the nodes, and again to move the items in OBS), it simply moves the nodes, and then tells OBS the new indexes.


holiber:
We should add some tests for issues that this PR fixed. I already have some in 'test/api/selection.ts'

PR #426, opened 2018-04-27
confirm before removing scenes

andycreeth:
This is a potentially very destructive action, so we should confirm it.


andycreeth:
Just realized this definitely breaks the tests.  Will take a look.

PR #432, opened 2018-05-02
Ux improvements

mbiemiller:


app/components/Selector.vue
```diff
@@ -16,7 +16,7 @@
       </div>
       <div class="selector-actions">
         <slot name="actions" :item="item"/>
-        <i v-if="draggable" class="icon-btn fa fa-bars fa-rotate-90 selector-drag-handle"/>
+        <!--<i v-if="draggable" class="icon-btn icon-move selector-drag-handle"/>-->
```
andycreeth: Assuming this is going away forever, can we delete this?

app/components/SourceSelector.vue.ts
```diff
@@ -5,17 +5,26 @@ import { SourcesService } from 'services/sources';
...
 @Component({
   components: { SlVueTree }
 })
 export default class SourceSelector extends Vue {
-
   @Inject() private scenesService: ScenesService;
   @Inject() private sourcesService: SourcesService;
   @Inject() private selectionService: SelectionService;
 
+  layersTooltip = 'The building blocks of your scene. Also contains widgets.';
```
andycreeth: Can we change all these references to "layers" in code to "sources"?  I think an average dev might be very confused about what a layer is if we never make that terminology switch.

app/components/StudioFooter.vue.ts
```diff
@@ -67,4 +67,6 @@ export default class StudioFooterComponent extends Vue {
       }
     }
   }
+
+  recordTooltip = 'Saves to computer at Videos > Slobs_recoding.';
```
andycreeth: This is actually configurable in the Settings.

app/components/SourceSelector.vue
```diff
@@ -1,99 +1,69 @@
...
-          <i v-else-if="node.data.type === 'dshow_input'" class="fa fa-camera"></i>
-          <i v-else-if="node.data.type === 'wasapi_input_capture'" class="fa fa-microphone"></i>
-          <i v-else-if="node.data.type === 'wasapi_output_capture'" class="fa fa-volume-up"></i>
-          <i v-else-if="node.data.type === 'monitor_capture'" class="fa fa-desktop"></i>
-          <i v-else-if="node.data.type === 'game_capture'" class="fa fa-gamepad"></i>
-          <i v-else-if="node.data.type === 'browser_source'" class="fa fa-globe"></i>
-          <i v-else-if="node.data.type === 'scene'" class="fa fa-sitemap"></i>
-          <i v-else-if="node.data.type === 'color_source'" class="fa fa-paint-brush"></i>
-          <i v-else class="fa fa-file"></i>
+    <sl-vue-tree :value="nodes" ref="slVueTree" @select="makeActive" @drop="handleSort" @toggle="toggleFolder" @contextmenu.native="showContextMenu()" @nodecontextmenu="(node, event) => showContextMenu(node.data.id, event)" @nodedblclick="node => sourceProperties(node.data.id)" :scrollAreaHeight="50" :maxScrollSpeed="15">
```
andycreeth: This is a really long line, can we keep the old formatting for this one?

mbiemiller: yeah agreed - I wonder if there's a way we can set Prettier not to change these

andycreeth:
Merging since I've run all tests locally.

PR #438, opened 2018-05-07
Input: Test for number of properties first

Xaymar:
Ran into this with the obs-studio-node `ipc-integration` branch, SLOBS is not properly testing if the source actually has properties. Instead we get an exception and a white window.

This happens due to different behavior in osn staging and ipc-integration. ipc-integration usually does not return an object when there's nothing to return (i.e. the returned object would be a lie), while staging actually returns an object even if that object would be invalid. So for ipc-integration to work correctly, testing for null/undefined must be done.

### DIFF
--- a/app/components/shared/forms/Input.ts
+++ b/app/components/shared/forms/Input.ts
@@ -298,6 +298,7 @@ export function getPropertiesFormData(obsSource: obs.ISource): TFormData {
   const obsSettings = obsSource.settings;
 
   if (!obsProps) return null;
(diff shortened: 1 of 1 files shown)


andycreeth:
These changes seem to result in the app completely locking up.

andycreeth:
Apparently `count()` is an infinite loop in the current bindings.  @computerquip-streamlabs is looking into it.

computerquip-streamlabs:
Fixed in v0.0.31, will add to branch.

PR #440, opened 2018-05-08
Performance Monitor: Stop monitoring if we're stopping monitoring

Xaymar:
Another one encountered rarely in obs-studio-node `ipc-integration`, at random we crash the server (and thus the client) as the performance monitor calls out into obs after shutting down obs.

### DIFF
--- a/app/services/performance-monitor/performance-monitor.ts
+++ b/app/services/performance-monitor/performance-monitor.ts
@@ -60,6 +60,7 @@ export class PerformanceMonitorService extends StatefulService<IMonitorState> {
 
 
   stop() {
+    this.performanceService.stop();
     clearInterval(this.intervalId);
     this.intervalId = null;
     if (this.droppedFramesSubscr) this.droppedFramesSubscr.unsubscribe();


Xaymar:
^ Above AppVeyor failure is from the Twitch test.

app/services/performance-monitor/performance-monitor.ts
```diff
@@ -60,6 +60,7 @@ export class PerformanceMonitorService extends StatefulService<IMonitorState> {
 
 
   stop() {
+    this.performanceService.stop();
```
andycreeth: It's somewhat semantics, but... just because we stopped the performance monitor it doesn't necessarily mean we want to also stop the performance service.  The performance service also provides the CPU and bitrate indicators in the footer.  Whereas the performance monitor is only for keeping dropped/skipped/lagged frame averages for displaying notifications.

What you have here works because right now the only time we stop the performance monitor is when we shut down the app.  However, I think it would be more correct to also shut down the performance service in the shutdown hook of the app service.  Since in the future, if we do decide to stop the performance monitor for some other reason, it would probably be unexpected if the stats in the footer shut off also.

PR #443, opened 2018-05-08
Translations

holiber:


app/services/i18n/i18n.service.ts
```diff
@@ -0,0 +1,161 @@
+import electron from 'electron';
```
andycreeth: Are you trying to introduce a new pattern here of naming services with `.service.ts`?  To me it seems a little redundant because it's already in `app/services` but if I'm missing something let me know.

app/i18n/ar-SA/common.json
```diff
@@ -0,0 +1,6 @@
+{
+  "Scene": "Scene",
+  "Scenes": "Scenes",
```
andycreeth: Does our localization system have a proper pluralization system?  I think some languages have multiple plural forms depending on whether it is "many" or "few".

app/services/i18n/i18n.service.ts
```diff
@@ -0,0 +1,161 @@
...
+import { FileManagerService } from 'services/file-manager';
+import { IListInput, TFormData } from 'components/shared/forms/Input';
+import { I18nServiceApi } from './i18n.service-api';
+
+
+interface II18nState {
+  locale: string;
+}
+
+const I18N_PATH = path.resolve('app/i18n');
```
andycreeth: The `app` directory is not included when we ship our app.  I think we should move i18n files out of `app` and into an `i18n` folder in the root of the repo.  Then I can manually include it in our package.

app/services/i18n/i18n.service.ts
holiber: oh, I just have a homemade project with this style, so this is a mistake

app/i18n/ar-SA/common.json
holiber: This plugin has some kind of pluralization http://kazupon.github.io/vue-i18n/en/pluralization.html. However, I prefer to avoid this stuff to not to make translations overcomplicated. In addition for example in the Russian language pluralization can be done only with help of very big dictionaries because the ending of a plural word depends not only how much things you have, it also depends on what kind of things you have (each word have gender `he`, `she` and 'it' that play a big role in pluralization). And some other languages have the same situation. So this is a big pain in ass and  instead `You have 5 cars` for multilingual projects I suggest to use `Your cars amount: 5`

app/services/i18n/i18n.service.ts
holiber: hmm, I added `i18n` in package.json but didn't test it yet

app/i18n/ar-SA/common.json
andycreeth: Cool sounds good.  You're the expert.  I never really know what looks/sounds weird in other languages and what doesn't.

app/services/i18n/i18n.service.ts
andycreeth: Yeah I saw you added `i18n` to extra files, but that is relative to the root of the repo, not the `app` directory.  So I think if you just move `<root>/app/i18n` to `<root>/i18n` everything will work.

PR #446, opened 2018-05-13
Source Transform Axis Centering

ocgineer:
Added ability to center a source on either the horizontal or vertical axis only. Touched up Transform context menu.

### DIFF
--- a/app/util/ScalableRectangle.ts
+++ b/app/util/ScalableRectangle.ts
@@ -16,6 +16,11 @@ export enum AnchorPoint {
   Center
 }
 
+export enum CenteringAxis {
+  X,
+  Y,
+  Both
+}
 
 // Positions on a positive unit grid
 const AnchorPositions = {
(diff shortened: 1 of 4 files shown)


app/util/ScalableRectangle.ts
```diff
@@ -295,14 +300,23 @@ export class ScalableRectangle implements IScalableRectangle {
...
+            case CenteringAxis.X:
+              this.x = rect.x;
+              break;
+            case CenteringAxis.Y:
+              this.y = rect.y;
+              break;
+            default:
+              this.x = rect.x;
+              this.y = rect.y;
+          }
```
andycreeth: If you wanted to get fancy here, you could use Typescripts support for bit flags with enums.  You would define your enum like this:

```TypeScript
enum ECenteringAxis {
  X = 1,
  Y = 1 << 1,
  Both = X | Y
}
```
Then you could use it like this:

```TypeScript
if (axis & ECenteringAxis.X) this.x = rect.x;
if (axis & ECenteringAxis.Y) this.y = rect.y;
```

The advantage here is that you don't have to separately handle the "both" case.  Both of those if statements will evaluate to true when the "both" enum is used.

andycreeth: Ah got it, makes sense.

PR #450, opened 2018-05-17
Update Electron to 2.0.1

computerquip-streamlabs:


andycreeth:
There's a few other places you have to change this.  Just do a global search for `1.8.4` in the repo and you'll find a handful of places we hardcode the electron version.

andycreeth:
I think one last spot you missed is `rebuild-node-obs.js`.  I have no idea if that script is still used though.  I certainly don't use it.

computerquip-streamlabs:
Nah, we don't use that since the artifacts were put in place. We should probably remove that script actually.

computerquip-streamlabs:
Woh... let me rebuild this locally and see what happened.

computerquip-streamlabs:
A couple issues arose here from some upgrades and updates. In particular, libobs is back to using a hardcoded path which I didn't get quite right the first time it seems. Another is a simple one from node-obs where a locale is now required to be specified. Should have a fix soon.

andycreeth:
Cool I think we're all set.  Gonna merge this once tests pass.

PR #455, opened 2018-05-18
Add the ability to hide audio sources in the mixer

andycreeth:
This is implemented in the same way that OBS does it.  Also, they should work for overlays in the library if we start adding hiding them when exporting.


app/services/scene-collections/nodes/overlays/slots.ts
```diff
@@ -29,7 +30,7 @@ interface IFilterInfo {
 interface IItemSchema {
   id: string;
   name: string;
-  sceneNodeType: TSceneNodeType;
+  sceneNodeType: 'item';
```
andycreeth: This change makes `TSceneNodeType` a discriminated union, which lets us remove the type casting we were permanent doing.  Instead, Typescript makes sure we have the appropriate type guards in place, and only lets us use it appropriately after a type guard.

PR #458, opened 2018-05-21
OpenVR Plugin Compatibility

ocgineer:
Whitelisting of OpenVR plugin with id 'openvr_capture' with added source example images.


andycreeth:
Will this only show up if you have some other OpenVR software installed on your computer?  The reason I'm curious is I think that we are getting to point where we have too many options when clicking "Add Source".  I would like to find a way to make niche sources like this available without adding to the overall number of sources people need to sift through.

ocgineer:
Not sure about tha tbh, it will also only show if the plugin is added which isnt in this PR as I want to fix a few things of the plugin itself; https://github.com/ocgineer/OBS-OpenVR-Input-Plugin

If there is maybe a menu source called custom or external, that could list the available installed plugins as-it like in OBS Studio, and that opens up all plugins.

ocgineer:
Might as well push this in with VR, LIV plugin is almost out of beta. Would be the same with OpenVR will only show up when the plugin is installed and they distribute the plugin. So having the UI of new sources a bit wonky would not affect regular users at this point.

PR #470, opened 2018-05-29
copy filters across scene collections

holiber:


holiber:
Yeah, just checked that. Looks like we can switch the scene collection only once in our tests. Going to investigate why

PR #472, opened 2018-05-30
package i18n into asar

holiber:


app/services/utils.ts
```diff
@@ -34,7 +34,7 @@ export default class Utils {
   }
 
   static isDevMode() {
-    return process.env.NODE_ENV !== 'production';
+    return electron.remote.process.env.NODE_ENV !== 'production';
```
holiber: for some reason, the old construction returned `development` mode when called on application start here

andycreeth: Yeah it's because we use we webpack to differentiate between development and production in the renderer these days.  I did that to cut down on IPC.  You need to use `yarn compile:production` to get webpack to build in production mode.

holiber:
Oh, good catch. By some reason, my IDE highlight those files as committed  🙁  but `git status` shows it is not

PR #473, opened 2018-05-30
Translations

holiber:


holiber:
Waiting for Morgan's HTML fixes here, after that is good to merge

app/components/StartStreamingButton.vue
```diff
@@ -10,12 +10,11 @@
...
+  min-width: 150px;
   text-transform: none;
 
-  @media(max-width: 1200px) {
-    font-size: 11px;
-    width: 112px;
-  }
+  // @media(max-width: 1200px) {
+  //   font-size: 11px;
+  // }
```
holiber: @mbiemiller do we need `@media` here now?

PR #477, opened 2018-05-31
fix startup with empty cache

holiber:


app/services/i18n/i18n.ts
```diff
@@ -32,7 +32,7 @@ export function $t(...args: any[]) {
 const LANG_CODE_MAP = {
   cs: { lang: 'Czech', locale: 'cs-CZ' },
   de: { lang: 'German', locale: 'de-DE' },
-  en:	{ lang: 'English', locale: 'en-US' },
+  'en-US':	{ lang: 'English', locale: 'en-US' },
```
andycreeth: Should be have both `en` and `en-US`?  It notice for portuguese we have both `pt` and `pt-BR`.

holiber: I changed it only for consistency, because we have translations only for `en-US` (as a source language), at the same time we have both `pt` and `pt-BR` translations.

andycreeth: Ah, got it.

app/services/i18n/i18n.ts
```diff
@@ -98,7 +98,7 @@ export class I18nService extends PersistentStatefulService<II18nState> implement
 
     // load dictionary if not loaded
     if (!this.loadedDictionaries[locale]) {
-      await this.loadDictionary(this.state.locale);
+      await this.loadDictionary(locale);
```
holiber: this is the main issue - I have to use `locale` instead `this.state.locale` when app cache is clean

PR #478, opened 2018-05-31
Popout recent events

andycreeth:


app/components/pages/Live.vue.ts
```diff
@@ -26,16 +27,33 @@ export default class Live extends Vue {
...
 
   mounted() {
     this.i18nService.setWebviewLocale(this.$refs.webview);
+
+    this.$refs.webview.addEventListener('new-window', e => {
+      const match = e.url.match(/dashboard\/([^\/^\?]*)/);
+
+      if (match && match[1] === 'recent-events') {
+        this.popout();
+      } else if (match) {
```
holiber: I'm a bit curious in which case we should redirect here?

app/components/windows/RecentEvents.vue
```diff
@@ -0,0 +1,20 @@
+<template>
+<modal-layout
+  title="Recent Events"
```
holiber: $t('Recent Events') here

app/components/pages/Live.vue.ts
andycreeth: There are a couple buttons in the recent events feed that link to the spin wheel and credits widget settings.  Before this change they have just never worked.  I saw an opportunity to make them start doing the correct thing pretty easily in the context of this larger change.

app/components/windows/RecentEvents.vue
andycreeth: Thanks.  I need to start thinking about translations when I add UI now.

andycreeth: Do I need to add this to one of the JSON files?

holiber: Yeah if it is not present there

PR #488, opened 2018-06-06
updated patch notes graphic

mbiemiller:


andycreeth:
Cool I'm gonna extract just this commit and make it against the `preview` branch because we have some stuff in `staging` that hasn't gone through QA.

PR #498, opened 2018-06-09
Korean translation

alexkoala:


andycreeth:
Hey, thanks for putting all this effort into translations!  Right now these files are actually generated automatically and come from our CrowdIn translation management website.  We would love it if you were able to contribute these translations through that service so we can include them in the app.  Can you please email michelle.henderson@streamlabs.com and she can set you up with an account on CrowdIn to get these translations included in the system.  Thank you!

PR #501, opened 2018-06-13
Add outage notifications system

andycreeth:
We will use this system to push notifications to the SLOBS UI to warn users about critical infrastructure outages.

### DIFF
--- a/app/services/outage-notifications.ts
+++ b/app/services/outage-notifications.ts
@@ -100,9 +100,12 @@ export class OutageNotificationsService extends Service {
 
   async fetchMessageJson(): Promise<IOutageNotification> {
     const req = new Request(this.messageUrl);
+    const headers = new Headers();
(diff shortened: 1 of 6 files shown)


app/services/outage-notifications.ts
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    // The current message is still in effect
+    if (this.currentMessageId === msg.id) return;
+
+    // There is a new message to display to the user
+    const notification = this.pushNotification(msg.message, msg.url);
+    this.currentMessageId = msg.id;
+    this.currentNotificationId = notification.id;
+  }
```
holiber: looks like that should be a private method

andycreeth: Yeah actually pretty much all of these methods should be.  This service should be treated as a black box that the rest of the app doesn't need to call into.

PR #505, opened 2018-06-14
Face mask alerts

kuhasuki:
Add facemask service
Add profanity filter service
Add new websocket event for face mask donation
Add test face mask button when facemask service is active
Add guest api interaction in dashboard


app/services/facemasks.ts
```diff
@@ -0,0 +1,617 @@
...
+    let fileContent = '';
+    return new Promise((resolve, reject) => {
+      this.ensureFacemasksDirectory();
+
+      https.get(this.libraryUrl(uuid), response => {
+        const writeStream = fs.createWriteStream(maskPath);
+        response.on('data', chunk => fileContent += chunk);
+        writeStream.on('finish', () => {
+          try {
+            const data = JSON.parse(fileContent) as IFacemask;
```
andycreeth: Cool so here we are downloading the file to disk and also keeping a copy in memory at the same time to pull out the modtime, right?  👍

kuhasuki: Yup.

app/components/pages/Dashboard.vue.ts
```diff
@@ -29,4 +40,28 @@ export default class Dashboard extends Vue {
   get dashboardUrl() {
     return this.userService.dashboardUrl(this.params.subPage || '');
   }
+
+  async getStatus() {
```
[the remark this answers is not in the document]
kuhasuki: They need to return a promise.

PR #506, opened 2018-06-14
Test widgets improvement

mbiemiller:
- fix for test widgets so it doesn't get cut off from preview on live page

### DIFF
--- a/app/app.less
+++ b/app/app.less
@@ -194,15 +194,11 @@ h4 {
 
 .slide-open__menu {
   position: absolute;
-  background: @day-secondary;
-  left: -100px;
-  right: -100px;
-  width: 170px;
-  margin: 0 auto;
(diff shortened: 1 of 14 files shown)


andycreeth:
Thanks for fixing this.  I'm notification as I use this that it feels like it should dismiss itself once I've chosen a widget to test.  Otherwise it has a tendency to just stick around and block important UI (like you notifications area).  What do you think?

mbiemiller:
I don't think we should close on click. People usually have a different variation for every event type so they test them all. I do however think we should automatically close it if you change pages or go live.

andycreeth:
Cool that sounds like a good solution to me.

andycreeth:
We can circle back and add that, since it's a little complicated.  Going to merge this for now.

PR #519, opened 2018-06-20
Hide follower goal for YouTube

jamacanbacn:
- Hide follower goal if you are using YouTube (they only have subscribers)
- Fix the lingo from youtube follows to subscribers


app/i18n/en-US/sources.json
```diff
@@ -43,7 +43,7 @@
   "Set a goal for your viewers to help you reach.": "Set a goal for your viewers to help you reach.",
   "Follower Goal": "Follower Goal",
   "Twitch Follows": "Twitch Follows",
-  "Youtube Follows": "Youtube Follows",
+  "Youtube Follows": "Youtube Subscribers",
```
andycreeth: This won't get translated into other languages unless you change the key (and all references to it).  I think this should be changed to be safe.

PR #522, opened 2018-06-23
Transitions V2: Named transitions and scene connections

andycreeth:
The goals of the new system are:
- Support multiple, named transitions
- Support creating connections for specific scene->scene transitions
- Maintain full backwards compatibility with existing scene collections

This will be a breaking change on our RPC API.  I'm not sure there's much we can do it about that.  The transitions API is undocumented so we shouldn't have any 3rd parties using it.

Still TODO (Future PR)
- Studio Mode support for choosing a named transition

Screenshots:

image

image

image


app/components/windows/SceneTransitions.vue
```diff
@@ -1,18 +1,90 @@
...
-      v-model="form.type"
-      @input="setTransitionType"/>
-
-    <IntInput
-      v-model="form.duration"
-      @input="setTransitionDuration"/>
-
-    <GenericForm v-model="properties" @input="saveProperties"></GenericForm>
+    <div v-if="!transitionsEnabled">
+      You need at least 2 scenes to edit transitions.
```
holiber: $t is missed

PR #523, opened 2018-06-25
Add tooltip to scene selector icons

gettinToasty:
This brings them up to parity with the source tooltips and explains the iconography more clearly. Because I'm a dirty overachiever I went ahead and added Korean translations as well :)


app/components/SceneSelector.vue.ts
```diff
@@ -29,6 +29,10 @@ export default class SceneSelector extends Vue {
 
   searchQuery = '';
 
+  addSceneTooltip = $t('Add a new Scene.');
+  removeSceneTooltip = $t('Remove Scene.');
+  showTransitionsTooltip = $t('Open the Scene Transitions.');
```
andycreeth: I appreciate you copying the tone/style of the source properties tooltip here.  That said, I think our current one is poorly worded so I'm wondering if we can phrase this a little better.  I would probably call this something more like "Edit Scene Transitions."

app/i18n/ko-KR/scenes.json
```diff
@@ -28,5 +28,8 @@
   "Your media files are currently being synced with the cloud. ": "현재 미디어 파일들이 클라우드에 동기화되고 있습니다. ",
   "It is recommended that you wait until this finishes before going live.": "방송하기 전에 작업이 끝날 떄까지 기다리는 것을 권장합니다.",
   "Go Live Anyway": "무시하고 방송하기",
-  "Wait": "기다리기"
+  "Wait": "기다리기",
+  "Add a new Scene.": "장면을 추가합니다.",
+  "Remove Scene": "장면을 제거합니다.",
+  "Open the Scene Transitions.": "장면의 변환을 엽니다."
```
andycreeth: Unfortunately submitting these translations here won't work.  You'll have to add them to crowdin, since all files other than english are automatically generated.

PR #524, opened 2018-06-25
Add confirmation dialogue to deletion of sources

gettinToasty:
Expands the selection service to add a confirmation dialogue when removing sources whether it be via the Delete key, the context menu, or the minus button.

### DIFF
--- a/app/services/selection/selection.ts
+++ b/app/services/selection/selection.ts
@@ -1,4 +1,5 @@
 import { uniq } from 'lodash';
+import electron from 'electron';
 import { mutation, StatefulService, ServiceHelper } from 'services/stateful-service';
 import {
   Scene,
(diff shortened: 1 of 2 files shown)


app/services/selection/selection.ts
```diff
@@ -95,7 +97,19 @@ export class SelectionService
 
   @shortcut('Delete')
   remove() {
-    return this.getSelection().remove.call(this);
+    const name = this.getLastSelected().name;
+    electron.remote.dialog.showMessageBox(
+      electron.remote.getCurrentWindow(),
+      {
+        type: 'warning',
+        message: $t('Are you sure you want to remove %{sceneName}?', { sceneName: name }),
```
gettinToasty: I kept this as `sceneName` so we wouldn't need a new translation key for what amounts to non-visible text, even though it isn't 100% descriptive anymore. Changing this and the one in `SceneSelector` to be `name` would be an alternative.

andycreeth: Ah, good thinking.

PR #528, opened 2018-07-02
chat-box widget settings

holiber:
Settings for the chat-box widget. Custom HTML/CSS/JS editing still ongoing


holiber:
We use `I` prefix for interfaces, `T` for types and generic types, `E` for enums. `W` prefix is for widget input components since we have input components for OBS settings.

PR #529, opened 2018-07-03
Add Media Gallery

gettinToasty:


holiber:
Good job! I like the selection style here much more than on the website. I'm going to add small fixes requests with my fresh view. Only one architecture issue that I see - you don't have to keep the selected item in the service state. This is definitely a vue component's responsibility. The knowledge of selected item makes sense only when the component window is open and unless if it is not. The common pattern for dialogs like that is to have a Promise-based function to simplyfy usage of this gallery in every place of code where you want. Here is an example how it can be done:

Usage:
````ts

// open media gallery and return selected item
const selectedFileUrl: string = await this.mediaGalleryService.showChoseFile()
if (selectedFileUrl) alert(`User has selected a file ${selectedFileUrl}`)
````

Implementation:
````ts

export class MediaGalleryService {

  private promises: Dictionary<{resolve: (fileUrl: string) => string, reject: () => void }> = {};

  // this method is called from other components and services to select a file from Media Gallery
  showChoseFile(): Promise<string> {
    const promiseId = ipcRenderer.sendSync('getUniqueId');
    this.showMediaGallery(promiseId);
    return new Promise(resolve, reject) {
       this.promises[promiseId] = { resolve, reject };
    }
  }

  // this method is called from the Media Gallery component when file is selected
  resolveFileSelect(promiseId: string, fileUrl: string) {
    this.promises[promiseId].resolve(fileUrl) ;
    delete this.promises[promiseId];
  }
}
````

**Comment by gettinToasty** [comment]
@holiber while i agree about only the component needing to know about the selected file, are there any use cases where a file would be selected while not in the component? if there aren't, that solution seems a bit like over-engineering when we can just pass the file into the parameters of anything that needs to use the selected file.

**Comment by holiber** [comment]
image
Here is some notes from my fresh view. This window is too big for my 13'' screen, so we have to rid of some unused space. The text `copy URL` will be translated on different languages and it will have  different length, so it is important to have the right position.

**Comment by holiber** [comment]
I see only one case when it's worth to keep selected file in the service for now. It's the case when you are going to open Media Gallery and restore the previous selected file. But for this way you should care about which of `My Uploads` or `Stock Files` tabs must be selected.

However this is not a big deal. The using the Promise-based pattern for all async widndows like that is more convinient, and doesn't require a lot of coding.

File: app/services/media-gallery/media-gallery.ts [hunk 1]
```diff
@@ -0,0 +1,274 @@
...
+          };
+          return reader.read().then(readStream);
+        }
+      );
+    }
+  }
+
+  deleteSelectedFile() {
+    const a = document.createElement('a');
+    a.href = this.state.selectedFile.href;
```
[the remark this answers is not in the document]
gettinToasty: That is just copied code from the dashboard app, but I left it in assuming it's a reliable way to get the `pathname` of the upload if the host is undeterministic

app/components/windows/MediaGallery.vue.ts
```diff
@@ -0,0 +1,167 @@
...
+    return;
+  }
+
+  close() {
+    return;
+  }
+
+  handleClose() {
+    return;
+  }
```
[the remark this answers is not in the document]
gettinToasty: i actually got rid of those with the redesign, they were originally to handle the cancel button

gettinToasty:
@holiber i'm working on that promise-based system you described, but i noticed this line: `this.showMediaGallery(promiseId);`, how would the `promiseId` be used in the showMediaGallery function? (which currently just opens the dialogue window)

holiber:
Ok, don't care about promise-based system right now. I'll just add a commit to your PR to be clear

gettinToasty:
@holiber in that case i'll push up the other CR changes so you can work on an updated branch

gettinToasty:
@holiber i think i managed to implement the promise system you described correctly, the important methods for that are `handleSelect` in `MediaGallery.vue.ts` as well as `selectFileFromGallery` and `resolveFileSelect` from `meda-gallery.ts`. let me know if i messed up anywhere!

i also kept the type and category in the service for now in case we need that information separate from the component, but if you think they should be moved to the component then i can do that as well

app/services/media-gallery/media-gallery.ts
```diff
@@ -46,62 +49,63 @@ const concatUint8Arrays = (a: Uint8Array, b: Uint8Array) => {
...
-        .pop();
-      if (item.size) {
-        totalUsage += parseInt(item.size, 10);
-      }
-
-      item.filename = filename;
-      item.type = filetypeMap[ext];
-      return item;
+  async pickFile(): Promise<IMediaGalleryFile> {
+    const promiseId = uuid();
```
gettinToasty: is there a reason to use `uuid()` here instead of `ipcRenderer.sendSync('getUniqueId');`?

app/services/media-gallery/stock-library.ts
```diff
@@ -0,0 +1,321 @@
...
+  },
+  {
+    href: 'http://uploads.twitchalerts.com/image-defaults/dBXACr6.gif',
+    fileName: 'dBXACr6.gif'
+  },
+  {
+    href: 'http://uploads.twitchalerts.com/image-defaults/sJJBQOT.gif',
+    fileName: 'sJJBQOT.gif'
+  }
+];
```
holiber: I removed the `map` loops from this file. The potential issue here - the code in such files will be executed on application startup even if you are not going to open the gallery. This is not a big problem for one file but, it can affect the application starting time if we will have many files like this. I moved `map` loops into the service here. Our services have lazy initialization, and services initialize the data only when is your first time of using it.

app/services/media-gallery/media-gallery.ts
holiber: ipcRenderer.sendSync('getUniqueId'); is a blocking IPC request, that is not too fast

app/services/media-gallery/stock-library.ts
gettinToasty: that's totally fine by me, that was just copypasted from the website

gettinToasty:
looks pretty good to me! there's not really a busy indicator on the website, so i created a pretty minimal darkening overlay as a placeholder for the busy state basically until we can get an actual design on it. i'm not gonna approve the PR cuz i've touched too much of the code but the refactor looks solid.

holiber:
When the first time I saw this darkening I thought that interface hungs because of I can't scroll it. As an option we can just add `Uploading...` message maybe somewhere here to be clear: 
image

holiber:
Looks good

PR #530, opened 2018-07-04
obs-studio-node IPC integration

eddyStreamlabs:


main.js
```diff
@@ -274,6 +276,19 @@ function startApp() {
...
+    var d = new Date().getTime();
+    if (typeof performance !== 'undefined' && typeof performance.now === 'function'){
+        d += performance.now(); //use high-precision timer if available
+    }
+    return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function (c) {
+        var r = (d + Math.random() * 16) % 16 | 0;
+        d = Math.floor(d / 16);
+        return (c === 'x' ? r : (r & 0x3 | 0x8)).toString(16);
+    });
+  }
```
andycreeth: We already have a v4 UUID generator required in this process.  Is there a reason we can't use that?  Do these IDs need to conform to another standard?

eddyStreamlabs: Yeah I know why the tests are failing, fixed a few already.

Oh yeah we can probably use it.

app/services/source-filters.ts
```diff
@@ -118,10 +118,10 @@ export class SourceFiltersService extends Service {
     const source = this.sourcesService.getSource(sourceId);
     const obsFilter = obs.FilterFactory.create(filterType, filterName);
     source.getObsInput().addFilter(obsFilter);
+    if (settings) obsFilter.update(settings);
```
holiber: This was an issue with filters copy/paste?

andycreeth: He just moved it up above the release.  The issue was that IPC didn't like updating the settings on a technically released reference, which worked fine in OSN.

eddyStreamlabs: Yes it was causing the copy/paste not to work properly. The filter was getting copies just fine but the settings were not ported over.

PR #538, opened 2018-07-06
Add styling to sliders with intervals

gettinToasty:


app/components/shared/widget-inputs/WSliderInput.vue
```diff
@@ -17,6 +17,14 @@
...
       ref="slider"
     />
+    <div v-if="metadata.interval && metadata.interval > 3">
+      <div
+        v-for="n in (metadata.max/metadata.interval + 1)"
+        :key="n"
+        class="slider-pip"
+        :class="[value >= n * interval ? 'active': '']"
+        :style="{ left: `${((n - 1)/(metadata.max/metadata.interval)) * 100}%` }" />
+    </div>
```
holiber: As I mentioned in Asana the slider component that we use already has "pips" like here  (`https://nightcatsama.github.io/vue-slider-component/example/`) and looks like it only needs some styles. We can go ahead with it, just check the documentation for the components we use next time.

gettinToasty: nah i'll redo it to make sure it's correct, but it would be super useful to clarify your language in the ticket. i thought you were referring to the snapping feature which is what i was asking about, not the display of the pips themselves

PR #544, opened 2018-07-07
fix(CI): Added yarn.lock check

timothystewart6:
Hi.  This fix will solve your dirty lock file woes.  It will fail the build if `yarn.lock` does not match your dependencies.

https://yarnpkg.com/lang/en/docs/cli/install/

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,7 +3,7 @@ build: off
 install:
   - ps: Install-Product node 8
   - git submodule update --init --recursive
-  - yarn install
+  - yarn install --frozen-lockfile --check-files
   - yarn install-plugins
   - yarn compile:ci
 


andycreeth:
Hey thanks for submitting this.  This makes a lot of sense and I would definitely like to merge this.

Before I click merge I need to get you to sign a simple Contributor License Agreement.  Unfortunately legal is still working on the language so I need to get that finalized before I can get this merged in.

Anyway, thanks a lot for the submission, I think this will help!

timothystewart6:
Sounds good.  I get it!  Thanks for considering, thanks for an awesome product, and thanks for merging this in (when we can).  I'd love to help more.  I use this product every day!

andycreeth:
@timothystewart6 we're finally all set.  If you have time to resolve the merge conflict and agree to the CLA above we can get this merged in.  Thanks again for your great suggestion and we'd love to accept your help again in the future.

timothystewart6:
@andycreeth all set!  It looks like the `install-plugins` `npm` script was also missing this check for plugins.  I added it there too.  Hopefully this helps with CI/development and pinning dependencies!  Thanks!

timothystewart6:
PS I think the pending CLA tag can be removed if you have that automated (once the CLA has been signed) 😄

PR #545, opened 2018-07-09
Take async flag into account when populating filters

computerquip-streamlabs:


app/services/source-filters.ts
```diff
@@ -109,19 +109,26 @@ export class SourceFiltersService extends Service {
...
 
   add(sourceId: string, filterType: TSourceFilterType, filterName: string, settings?: Dictionary<TObsValue>) {
     const source = this.sourcesService.getSource(sourceId);
-    const obsFilter = obs.FilterFactory.create(filterType, filterName);
+    let obsFilter = null;
+
+    if (settings)
+      obsFilter = obs.FilterFactory.create(filterType, filterName, settings);
+    else
+      obsFilter = obs.FilterFactory.create(filterType, filterName);
```
andycreeth:
```TypeScript
const obsFilter = obs.FilterFactory.create(filterType, filterName, settings);
```
Should work fine if `settings` is `undefined`.  If it's `null`, I'm not entirely how that translates once it hits C++ land, but in that case, this might be safer:
```TypeScript
const obsFilter = obs.FilterFactory.create(filterType, filterName, settings || {});
```

computerquip-streamlabs: Well, this is actually a work around of my own undoing. In the C++ API, I do checks on the length of the arguments array. If you pass undefined explicitly, it increases the length of the array. I kinda want to fix that but we're in-between for OSN right now.

andycreeth: Ah I see.  Does the second option work?  Empty object should just apply all defaults right?

computerquip-streamlabs: Yeah. I've changed it to reflect that.

That said and unrelated, I actually found a bug in this logic. Synchronous filters work with asynchronous sources but not the other way around. Right now, for instance, the video capture source is gimped in filter selection. Will fix now...

PR #547, opened 2018-07-10
Truncate long scene names

gettinToasty:


andycreeth:
This seems to be a little over aggressive on the truncation by my testing:
image

gettinToasty:
@andycreeth i tested using strings of the longest character size (w) since we don't use mono-spaced fonts there's not really a good way to calculate how long the strings will actually be pixel-wise. I erred on the side of caution but if you think there's a better metric please let me know!

Xaymar:
If electron has CSS 3 support, CSS 3 has an automatic truncation option that inserts three dots:

```
overflow: hidden;
white-space: nowrap;
text-overflow: ellipsis;
```

gettinToasty:
@Xaymar it does but there was a lot of weirdness when trying to implement a solution using pure CSS, this ended up being the best way to ensure zero UI obstruction/clipping

Xaymar:
The CSS rules I've posted above do work, and are actually currently being used. What is breaking them is the `display: flex;` set on .selector-item-text. Disabling it didn't seem to have any negative effect so far and properly added ellipsis at the end of the displayable text.

Something similar could probably be added for sources, though I'm unsure about that one.

gettinToasty:
@Xaymar it works until you start trying to compress the window, resulting in portions of the UI being clipped off-screen at resolutions less than about 1420px. There are ways to make it work with flexbox as well but those resulted in similar issues. The way I did it through JS means that even at small resolutions (widths of about 540px or halfscreen at 1080p) all of our UI elements still appear, notably the Go Live button and the top-right toolbar.

Xaymar:
I see and can confirm the long names causing issues with the screen size. This is the case for both Scene and Source names. I still think this isn't a very ideal solution, but it's a working workaround to an existing issue.

I do not know why our current flex design doesn't correctly size itself to the content and instead increases the min-width of whatever is above it, even if the content is cropped off. I've tried to debug it, but it didn't make sense.

gettinToasty:
I agree that this is not ideal and a pure CSS solution would be much cleaner, but I can't seem to find a way to find a way to debug the issue without spending way too much time with it.

andycreeth:
I'm closing to keep down our number of open PRs.  Feel free to reopen when we want to address this again.

PR #550, opened 2018-07-10
close source properties when underlying source is closed

andycreeth:


app/components/windows/SourceProperties.vue.ts
```diff
@@ -35,8 +36,19 @@ export default class SourceProperties extends Vue {
   source = this.sourcesService.getSource(this.sourceId);
   properties: TFormData = [];
 
+  sourcesSubscription: Subscription;
+
   mounted() {
     this.properties = this.source ? this.source.getPropertiesFormData() : [];
+    this.sourcesSubscription = this.sourcesService.sourceRemoved.subscribe(source => {
```
holiber: Does it work when you switch scene collection?

andycreeth: Yup.  All of these events are fired even when sources are removed as part of swapping a scene collection.

PR #551, opened 2018-07-10
Widgets code editor

holiber:
image


holiber:
The tabs always are shown for now. However, I'm thinking about to use side nav instead of tabs, because of some widgets have too many tabs. About injections - you can't create a robust solution against XSS on javascript side, this is mostly a backend task.

gettinToasty:
i figured if anything it would be on the backend, was just wondering if we have something in place :)

holiber:
I'm preparing a PR with input validations for now. That will be enough for the frontend side.

PR #552, opened 2018-07-11
Source property optimization

computerquip-streamlabs:
This should fix the Elgato `Configure Video` button along with a myriad of other optimizations.

### DIFF
--- a/app/components/shared/forms/Input.ts
+++ b/app/components/shared/forms/Input.ts
@@ -293,11 +293,11 @@ export function inputValuesToObsValues(
 
 export function getPropertiesFormData(obsSource: obs.ISource): TFormData {
 
-  setupSourceDefaults(obsSource);
-
   const formData: TFormData = [];
   const obsProps = obsSource.properties;
   const obsSettings = obsSource.settings;
+
+  setupConfigurableDefaults(obsSource, obsProps, obsSettings);
 
   if (!obsProps) return null;
   if (!obsProps.count()) return null;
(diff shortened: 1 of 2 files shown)


app/components/shared/forms/Input.ts
```diff
@@ -407,6 +407,15 @@ export function setPropertiesFormData(obsSource: obs.ISource, form: TFormData) {
     }
   });
 
+  /* Don't fetch properties unless we use it. */
+  if (buttons.length !== 0)
+    properties = obsSource.properties;
```
andycreeth: I noticed you omit the curly braces a lot for single line if-statements.  We loosely follow Airbnb's style guide which says that curly braces should only be omitted if the entire statement is a single line.

Here are some examples of what is acceptable and what isn't: https://github.com/airbnb/javascript#blocks

And here's a good explanation of why this style can potentially lead to bugs down the road: https://medium.com/@jonathanabrams/single-line-if-statements-2565c62ff492

I believe Prettier should be doing this automatically, but if it isn't, let me know.

app/components/shared/forms/Input.ts
```diff
@@ -417,38 +426,59 @@ export function setPropertiesFormData(obsSource: obs.ISource, form: TFormData) {
     }
   });
 
-  obsSource.update(settings);
+  /* Don't update unless we need to. */
+  if (formInputs.length === 0)
+    return;
```
andycreeth: Same as above.

computerquip-streamlabs:
I've made some changes and reorganized the commits a little bit to closer reflect each title.

holiber:
I can't change the camera resolution. Is that ok?

image

PR #559, opened 2018-07-11
Fix filter compatibility

computerquip-streamlabs:
Second iteration of the filter compatibility check for asynchronous sources. I'm still not 100% if this will cover all cases, the logic surrounding asynchronous sources is a lot more annoying and confusing than I originally thought. I think the best way to test this is to check each filter and compare to the Qt OBS UI.

### DIFF
--- a/app/services/source-filters.ts
+++ b/app/services/source-filters.ts
@@ -11,7 +11,6 @@ import namingHelpers from '../util/NamingHelpers';
 import { $t } from 'services/i18n';
 import { EOrderMovement } from 'obs-studio-node';
 
-
 export type TSourceFilterType =
   'mask_filter' |
   'crop_filter' |
@@ -109,19 +108,35 @@ export class SourceFiltersService extends Service {
   getTypesForSource(sourceId: string): ISourceFilterType[] {
     const source = this.sourcesService.getSource(sourceId);
     return this.getTypes().filter(filterType => {
-      return (filterType.audio && source.audio) || (filterType.video && source.video);
+      /* Audio filters can be applied to audio sources. */
+      if (source.audio && filterType.audio) {
+        return true;
+      }
+
+      /* We have either a video filter or source */
+      /* Can't apply asynchronous video filters to non-asynchronous video souces. */
+      if (!source.async && filterType.async) {
+        return false;
+      }
+
+      /* Video filters can be applied to video sources. */
+      if (source.video && filterType.video) {
+        return true;
+      }
+
+      return false;
     });
   }
 
 
   add(sourceId: string, filterType: TSourceFilterType, filterName: string, settings?: Dictionary<TObsValue>) {
     const source = this.sourcesService.getSource(sourceId);
-    const obsFilter = obs.FilterFactory.create(filterType, filterName);
+    const obsFilter = obs.FilterFactory.create(filterType, filterName, settings || {});
+
     source.getObsInput().addFilter(obsFilter);
     // There is now 2 references to the filter at that point
     // We need to release one
     obsFilter.release();
-    if (settings) obsFilter.update(settings);
     return obsFilter;
   }
 
--- a/app/services/sources/source.ts
+++ b/app/services/sources/source.ts
@@ -25,6 +25,7 @@ export class Source implements ISourceApi {
   type: TSourceType;
   audio: boolean;
   video: boolean;
+  async: boolean;
   muted: boolean;
   width: number;
   height: number;
--- a/app/services/sources/sources-api.ts
+++ b/app/services/sources/sources-api.ts
@@ -9,6 +9,7 @@ export interface ISource extends IResource {
   type: TSourceType;
   audio: boolean;
   video: boolean;
+  async: boolean;
   muted: boolean;
   width: number;
   height: number;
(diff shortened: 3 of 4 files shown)


computerquip-streamlabs:
OBS's Jim was kind enough to reformat the code I adapted this from. It reads like this:
```
static bool filter_compatible(bool creatingAsyncPopupMenu, uint32_t sourceFlags,
        uint32_t filterFlags)
{
    bool filterIsVideo       = (filterFlags & OBS_SOURCE_VIDEO) != 0;
    bool filterIsAsync       = (filterFlags & OBS_SOURCE_ASYNC) != 0;
    bool filterIsAudio       = (filterFlags & OBS_SOURCE_AUDIO) != 0;
    bool sourceHasAudio      = (sourceFlags & OBS_SOURCE_AUDIO) != 0;
    bool sourceHasAudioOnly  = (sourceFlags & OBS_SOURCE_VIDEO) == 0;
    bool sourceIsAsyncSource = (sourceFlags & OBS_SOURCE_ASYNC) != 0;

    /* async popup menu means that the filter has to be either audio or
     * async video */
    if (creatingAsyncPopupMenu) {
        /* if creating an async popup menu, and the filter is video,
         * but the source only has audio, fail */
        if (sourceHasAudioOnly && filterIsVideo) {
            return false;
        }
        /* if creating an async popup menu, and the source has no
         * audio, and has no async video, fail. */
        if (!sourceHasAudio && !sourceIsAsyncSource) {
            return false;
        }
        /* otherwise, if the filter has audio, or if the filter is
         * async video, succeed. */
        if (filterIsAudio || filterIsAsync) {
            return true;
        }
    } else {
        /* if creating a non-async popup menu, filter can only be a
         * synchronous video effect filter */
        if (!filterIsAudio && !filterIsAsync)
            return true;
    }

    /* everything else, return false */
    return false;
}
```

Our case is a little different since we don't have a popup menu and we don't separate async filters but still an alright reference.

Xaymar:
From what I understood, the following logic tables should apply :

|               | Video | Audio | Video & Async | Video & Audio | All |
| ------------- | ----- | ----- | ------------- | ------------- | --- |
| Video         | Yes   | No    | No            | Yes           | No  |
| Audio         | No    | Yes   | No            | Yes           | Yes |
| Video & Async | No    | No    | Yes           | No            | Yes |

(Down is Filter, Right is Source)

Which should result in the following:

```
// Source has neither Audio or Video, so abort early.
if (!source.audio && !source.video) {
	return false;
}

// Audio Filters can be applied to Audio Sources.
if (source.audio && filter.audio) {
	return true;
}

// Can't apply synchronous video filters to asynchronous video.
if ((source.async) && (source.async != filter.async)) {
	return false;
}

// Video Filters can be applied to Video Sources.
if (source.video && filter.video) {
	return true;
}		

return false;
```

That is, if I read everything correctly. I don't know if the async flag co-exists with the video flag.

computerquip-streamlabs:
Is it possible for an input source to have neither video or audio?

computerquip-streamlabs:
Put that code into practice, seems to allow non-async sources to use async filters for instance, the color source is listed to be compatible with video delay which is incorrect.

computerquip-streamlabs:
I've pushed the cleaner more readable version of the code. I've also removed the need to check for audio/video sources since I don't think there's ever a case there.

Xaymar:
> That logic table is slight off as well, where non-async filters can apply to async sources.

The logic table is to be read as boolean flags:

- Video Source and Video Filter are "Yes"
- Video Source and Audio Filter are "No"
- Video Source and Async Video Filter are "No"
- Audio Source and Video Filter are "No"
- Audio Source and Audio Filter are "Yes"
- Audio Source and Async Video Filter are "No"
- Async Video Source and Video Filter are "No"
- Async Video Source and Audio Filter are "No"
- Async Video Source and Async Video Filter are "Yes"
- Video & Audio Source and Video Filter are "Yes"
- Video & Audio Source and Audio Filter are "Yes"
- Video & Audio Source and Async Video Filter are "No"
- "All" Source and Video Filter are "No" (because Async)
- "All" Source and Audio Filter are "Yes"
- "All" Source and Async Video Filter are "Yes" (because Async)

Or in short:

- Video Filters can be applied to the Video Flag.
- Audio Filters can be applied to the Audio Flag.
- Async Filters can be applied to the Async Flag.

> Put that code into practice, seems to allow non-async sources to use async filters for instance, the color source is listed to be compatible with video delay which is incorrect.

The async test needs to be changed a little bit, as the way it was it would not apply the ruleset. I noticed the error just now:

```
// Block any filter that is not using the same synchronization as the source.
if (source.async != filter.async) {
	return false;
}
```

Xaymar:
> Is it possible for an input source to have neither video or audio?

Yes, at least in the case of OBS itself. It is possible to create/load sources of an invalid type, which in OBS Studio render with nothing. I'm not sure if this could ever happen for us, unless a user messed with things.

computerquip-streamlabs:
```
// Block any filter that is not using the same synchronization as the source.
if (source.async != filter.async) {
	return false;
}
```
This doesn't appear to be correct since you can apply filters that don't claim to be async to sources that are async. For instance, you can apply a color key filter to a media source or a video capture .

Xaymar:
Ah so non-async video filters can be applied to async video. In that case the code you've written is correct.

computerquip-streamlabs:
Thought I could reset review but I guess not. This PR should be good to go now 👍

andycreeth:
Cool, was wondering what the status was on this.  Will merge.

PR #561, opened 2018-07-13
Style disabled button

gettinToasty:
disabled_button_day
disabled_button_night


holiber:
does it work in the day mode?

gettinToasty:
@holiber yup! they grey color is not specific to a color scheme, but i'll take some screenshots and post them on the orig PR message for easy visual comparison

PR #567, opened 2018-07-16
Debounce the slider to prevent constant updates

computerquip-streamlabs:
This makes the slider a bit more optimal and smooth, with simple changes. 

This could be used with number input boxes as well (and should be) which should also apply to the formatting applied to the input box. This way, the user can enter whatever they want and only the intended value gets formatted rather than every value in between as well. 

Let me know if anything unorthodox is done here.

### DIFF
--- a/app/components/shared/Slider.vue.ts
+++ b/app/components/shared/Slider.vue.ts
@@ -1,6 +1,5 @@
 import Vue from 'vue';
 import VueSlider from 'vue-slider-component';
-import { throttle } from 'lodash-decorators';
 import { Component, Prop } from 'vue-property-decorator';
 
 @Component({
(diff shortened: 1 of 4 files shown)


app/components/shared/Slider.vue.ts
```diff
@@ -27,7 +26,11 @@ export default class SliderInput extends Vue {
     }, 500);
   }
 
-  @throttle(500)
+  sliderUpdate(value: number) {
+    this.$refs.inputbox.value = value;
```
andycreeth: We really shouldn't need to use refs for this.  Can you explain to me what this is supposed to be doing?

computerquip-streamlabs: Input box value doesn't seem to get updated until the input event is handled by the parent component. Since the input event handler is delayed, it doesn't seem get updated along with the slider.

andycreeth: Ah ok got it.  I think we should use an intermediate local state variable for that.  I think that should be handled by `SliderInput.vue.ts`  since it is the OBS-specific component doing the debouncing.  Mind if I push a commit to this branch?  Will be easier than typing it all out here.

PR #572, opened 2018-07-17
Preserve previous window component

holiber:
I used Vue's `<keep-alive>` component to store the previously shown component in the cache. For now, if you select a source from `SourcesShowcase` and close `SourceName` window you will go back to the  `SourcesShowcase` window with the same state.


andycreeth:
What's the status on this PR?

holiber:
I'm going to take it to work right after inputs refactoring.

andycreeth:
It seems to me that what you're effectively implementing here is a component stack.

When opening a new window with the "preserve previous" option set to true, you want to push a new component onto the stack.

When closing a window you want to pop a component off the stack.  If you popped the last component off the stack, then just close the window as normal.

When opening a new window without the "preserve previous" option set to true, you want to just throw out the entire stack and create a new one with 1 component.

I guess what I'm saying is, you've essentially implemented this system already, but capped it to a stack height of 2.  I'm not convinced that a general purpose solution is actually much more complex here, and doesn't have the downside of being capped to only 2 components.  Just something to think about.

holiber:
Yes, it is a stack in the `WidowsService` that caped with 2 components, it was an original idea from the beginning of this PR. My goal was to rid of the disadvantages of the solution I have before with using uncontrollable `keep-alive`. Do you wish to have some general purpose solution?

gettinToasty:
i think what andy is saying is that making it a non-limited general use system has a pretty low cost and potentially big benefit later down the road

holiber:
Just sent a packaged version of this PR for more testing

PR #580, opened 2018-07-18
Optimize heavy areas with IPC use

computerquip-streamlabs:
This was a small optimization that's slightly noticeable when dragging sources around. It prevents fetching the scene as well as the scene item for every single scene item being dragged. This modifies the scene-item api in those hot areas to allow passing an obs object that's already created instead of creating it on the fly. In addition, we use already created objects or cache objects where possible to prevent fetching new ones.

### DIFF
--- a/app/services/scenes/scene-item.ts
+++ b/app/services/scenes/scene-item.ts
@@ -99,8 +99,9 @@ export class SceneItem extends SceneItemNode implements ISceneItemApi {
     return this.source.getObsInput();
   }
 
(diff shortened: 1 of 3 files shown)


computerquip-streamlabs:
BY the way, this can be reviewed but I probably wouldn't pull it in, it's more or less a staging area for hot areas I find and work arounds.

andycreeth:
Got distracted, going to put this one on the back burner for a little bit.

PR #581, opened 2018-07-18
adding basic vue files and updating some relative imports

Nitsorn:


app/components/page-components/Chatbot/ChatbotCommands.vue.ts
```diff
@@ -0,0 +1,13 @@
+import Vue from 'vue';
+import { Component, Prop } from 'vue-property-decorator';
+
+@Component({
+})
+export default class ChatbotModules extends Vue {
+  data() {
```
andycreeth: FYI we generally prefer to use class attributes for reactive data on components.

app/components/pages/Chatbot.vue.ts
```diff
@@ -0,0 +1,29 @@
...
+    Commands: 'icon-suggestions',
+    Timers: 'icon-time',
+    'Mod Tools': 'icon-settings-3-1'
+  };
+  data() {
+    return {
+      tabNames: ['Modules', 'Commands', 'Timers', 'Mod Tools'],
+      selectedTab: 'Modules'
+    };
+  }
```
andycreeth: These would be better defined on the class, for example:

```TypeScript
export default class Chatbot extends Vue {
  icons: Dictionary<string> = {
    Modules: 'icon-widgets',
    Commands: 'icon-suggestions',
    Timers: 'icon-time',
    'Mod Tools': 'icon-settings-3-1'
  };

  tabNames = ['Modules', 'Commands', 'Timers', 'Mod Tools'];
  selectedTab = 'Modules';
}
```

Since you've initialized them to a value, they will automatically be reactive, just as if you had put them in the data object of an object-style Vue component.

app/services/chatbot-api.ts
```diff
@@ -0,0 +1,40 @@
...
+        .then(response => response.json())
+        .then(response => {
+          debugger;
+          resolve(response);
+        })
+        .catch(err => {
+          reject(err);
+        });
+    })
+  }
```
andycreeth: This function could be a lot simpler if you didn't wrap your own promise but instead used the promise returned by fetch.  This code should be equivalent:

```TypeScript
  logIn() {
    const url = this.apiEndpoint('login');
    const headers = authorizedHeaders(this.userService.apiToken);
    headers.append('Content-Type', 'application/json');
    const request = new Request(url, { 
      headers,
      method: 'POST',
      body: JSON.stringify({})
    });

    return fetch(request)
      .then(handleErrors)
      .then(response => response.json());
  }
```

andycreeth:
Going to close this until chatbot is ready to merge in.

PR #636, opened 2018-07-23
update tslint config, fix some code issues

andycreeth:
In the PR:
- Switch tslint base to a "prettier" base
- Remove old rules that don't exist anymore or don't do anything for our codebase
- Change some rules we had to better reflect current style
- Update and places in code that don't adhere to new style

Still to be resolved:
I think we should find a replacement for prettier.  I've become increasingly unimpressed with some of the changes it makes.  I think I would prefer getting a rock solid TSLint config and relying on the auto-fix that TSLint can do more most rules.  I'm not going to make that change right now though because I don't have time to explore all the merits of both directions.

### DIFF
--- a/app/services/scene-collections/nodes/sources.ts
+++ b/app/services/scene-collections/nodes/sources.ts
@@ -48,7 +48,6 @@ export interface ISourceInfo {
 }
 
 export class SourcesNode extends Node<ISchema, {}> {
-
   schemaVersion = 3;
 
   @Inject() private fontLibraryService: FontLibraryService;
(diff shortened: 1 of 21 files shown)


app/services/scene-collections/nodes/sources.ts
```diff
@@ -218,7 +216,7 @@ export class SourcesNode extends Node<ISchema, {}> {
             name: filter.name,
             type: filter.type,
             settings: filter.settings,
-            enabled: (filter.enabled === void 0) ? true : filter.enabled
```
holiber: I always thought it is nothing bad to add additional brackets, especially in a ternary operator O_o

andycreeth: Yeah I actually ran Prettier on this file and this was one of the changes it made, which had nothing to do with TSLint.  This is actually a great example of Prettier changing something that really didn't need to be changed, and arguably the end result is worse.

PR #637, opened 2018-07-23
Add new credits settings

gettinToasty:


app/components/widget-settings/Credits.vue.ts
```diff
@@ -0,0 +1,36 @@
...
+      value: theme
+    }));
+  }
+
+  textColorTooltip = $t('A hex code for the base text color.');
+  delayTimeTooltip = $t('Wait time before rerunning the credit reel.');
+  rollSpeedTooltip = $t('Speed of the rolling credits.');
+  rollTimeTooltip = $t('Duration of the rolling credits.');
+  creditsSubtitleTooltip = $t(
+    'When the credits roll, this will be the format of the subtitle. Available tokens: {total_donated_amount}, {total_cheer_amount}, {top_donor}, {top_donated_amount}, {top_cheer_donor}, {username}, {top_cheer_amount}, {new_subscriber_count}, {new_follower_count}.'
```
andycreeth: I wonder if there's some way we can indicate to the translators that these tokens shouldn't be translated.  Maybe it's obvious, I dunno.

app/services/widget-settings/credits.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  patchAfterFetch(data: any): ICreditsData {
+    // transform platform types to simple booleans
+    return data;
+  }
+
+  patchBeforeSend(settings: ICreditsSettings): any {
+    // the API accepts an object instead of simple booleans for platforms
+    return settings;
+  }
```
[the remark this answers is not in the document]
holiber: these methods already implemented in `WidgetSettingsService` you need touch it only if you wish to make some changes

app/components/widget-settings/Credits.vue.ts
gettinToasty: ohhhhh you're specifically referring to the `{total_donated_amount}` etc tokens? ok yea that makes a lot more sense to me now hahaha i thought you were just talking about tooltips in general :X

holiber: @gettinToasty I think Andy means to add something like
````js
creditsSubtitleTooltip = $t('When the credits roll, this will be the format of the subtitle. Available tokens:') +
 '{total_donated_amount}, {total_cheer_amount}, {top_donor}, {top_donated_amount}, {top_cheer_donor}, {username}, {top_cheer_amount}, {new_subscriber_count}, {new_follower_count}.'
````

gettinToasty:
i just realized i forgot to add YT/mixer/etc support for this. what's the best way to determine what platform a user is on?

andycreeth:
The user service can tell you what platform a user is logged in to.  Is there anything to do for that?  Unless there are different settings for different platforms, it shouldn't require any additional work.

gettinToasty:
@andycreeth the api gives back data for non-twitch versions of these settings:
credits

andycreeth:
Ah got it.  Yeah just query the user service to determine what platform they are logged into.

gettinToasty:
LOL wait nevermind it totally wasn't this component-- must have been tip jar or something

PR #639, opened 2018-07-23
Fix w textarea input

Nitsorn:
fixing w-textarea-input metadata to be of type Text and not basic Input, and update textarea font size to match twitchalertscss


Nitsorn:
will do another PR after removing it from submodules

app/components/shared/widget-inputs/WTextAreaInput.vue.ts
```diff
@@ -15,5 +16,9 @@ export default class WTextAreaInput extends WInput<string, IWTextMetadata> {
   @Prop({ default: () => ({}) })
   metadata: IWTextMetadata;
 
-
+  getValidations() {
+    return {
+      ...super.getValidations(),
+    };
+  }
```
holiber: Hmm, do we really need that here?

Nitsorn: true :) just copy pasted. removing now

holiber: WTextAreaInput extends WInput that already has `getValidations` method

andycreeth:
`yarn-error.log` was committed.  I've been meaning to add that to the `.gitignore` file.

Nitsorn:
@andycreeth will add it, thanks

app/components/shared/widget-inputs/WTextAreaInput.vue
```diff
@@ -1,11 +1,18 @@
 <template>
-  <textarea
+  <span>
+    <textarea
       type="text"
       :placeholder="metadata.placeholder"
```
holiber: Little fix that I forgot about - replace `metadata.placeholder` to `options.placeholder`. The `metadata` property can be undefined while `options` always exist.  After that, it's good for me.

Nitsorn: @holiber done! thanks for info

andycreeth:
Hmm weird, `yarn-error.log` was added to the gitignore, but it's also still listed in the diff. I think you might need to delete it.

PR #642, opened 2018-07-23
Use prebuilt node-libuiohook for consistent and faster builds

computerquip-streamlabs:


computerquip-streamlabs:
Seems that cmake-js was pulled in as a dependency which complicated our lock and environment a bit. Good to have that stuff out.

computerquip-streamlabs:
This will require https://github.com/stream-labs/streamlabs-obs/pull/640 to be pulled in before a build will succeed.

computerquip-streamlabs:
The branch will refuse to build here since node-fontinfo was pulling in a required dependency. This branch is fine.

PR #647, opened 2018-07-24
Refactor SourcesShowcase

gettinToasty:
Would love to hit the `add-source-info portion` of this as well but that would take a lot more work/architectural changes I think. This is at least a partial solution to getting rid of some of the tight coupling between our widget types being explicitly referenced, the end goal of which is to make it much easier to add or remove widgets by simply changing the enum/definition instead of having to touch many different files.

### DIFF
--- a/app/services/widgets.ts
+++ b/app/services/widgets.ts
@@ -442,7 +442,6 @@ export class WidgetsService extends Service {
   }
 
   getWidgetPlatform(type: WidgetType): string {
-    console.log(type);
     return WidgetDefinitions[this.getWidgetComponent(type)].platform;
   }
 
--- a/app/services/widgets.ts
+++ b/app/services/widgets.ts
@@ -15,16 +15,17 @@ import { WidgetSettingsService } from './widget-settings/widget-settings';
 import { ServicesManager } from '../services-manager';
 import { authorizedHeaders } from 'util/requests';
 
+// Do not alter the order of this enum, it is coupled to the user's local config
 export enum WidgetType {
   AlertBox = 0,
-  EventList = 1,
-  TheJar = 2,
-  DonationGoal = 3,
+  DonationGoal = 1,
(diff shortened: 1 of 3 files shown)


app/services/widgets.ts
```diff
@@ -17,14 +17,14 @@ import { authorizedHeaders } from 'util/requests';
...
+  EventList = 1,
+  TheJar = 2,
+  DonationGoal = 3,
   BitGoal = 4,
   DonationTicker = 5,
   ChatBox = 6,
-  EventList = 7,
-  TheJar = 8,
+  FollowerGoal = 7,
+  SubscriberGoal = 8,
```
andycreeth: This is not a safe change make.  These values are stored in the config file.  This means the next time their config is loaded, the widget will be created as something completely different.

Ideally we would be using string enums here, but at the time I wrote this code, TypeScript didn't support string enums.

gettinToasty: i was wondering if that was breaking, good to know it is. the only reason i did it was to sort the "essential" elements to the top, but that could also just be done via a JS sort on `interableWidgetTypes` i suppose

andycreeth: Yeah.  I think we probably just want to keep a separate display order.

Also, it's kind of bad how non-obvious it is that this value is stored in the config file.  I would like to figure out a way to make that more obvious.  Maybe just a comment is good for now.

app/services/widgets.ts
```diff
@@ -430,6 +437,15 @@ export class WidgetsService extends Service {
     return WidgetType[type];
   }
 
+  getWidgetName(type: WidgetType): string {
+    return WidgetDefinitions[this.getWidgetComponent(type)].name;
+  }
+
+  getWidgetPlatform(type: WidgetType): string {
+    console.log(type);
```
holiber: console.log

PR #651, opened 2018-07-25
Use prebuilt font-manager node-libuiohook for consistent and faster builds

computerquip-streamlabs:
This includes a commit from https://github.com/stream-labs/streamlabs-obs/pull/6427 in order to build. I've closed the other as a result.


andycreeth:
Unfortunately the node-gyp merge caused a conflict here

PR #653, opened 2018-07-25
Inputs refactor: part 2

holiber:
Example of usage:

```
// basic component
<number-input v-model="numberValue" :metadata="numberMetadata"/>

// generic component
<form-input v-model="myValue" type="number" :metadata="myValueMetadata"/>

// wrap the component into 2 columns layout (just for reducing the template code)
<form-group v-model="myValue" type="number"  title="My input" :metadata="myValueMetadata"/>

// wrap several components into 2 columns layout 
<form-group title="My checkboxes">
  <bool-input title"Check Me" v-model="value1"/>
  <bool-input title"Check Me" v-model="value1"/>
  <bool-input title"Check Me" v-model="value1"/>
</form-group>

```

# Vue props vs metadata object

Why instead of fancy syntax like
````
<slider-input  :mix="0" :min="10" :interval="1"/>
````

I should use the single metadata object?
````
const sliderMetadata = metadata.slider({ min: 0, max: 10, interval: 1});
....
<slider-input metadata="sliderMetadata"/>
````

This is because of Vue can't do typeckecking it templates like  `Angular` and `React tsx` does. Vue does typeckeking only in runtime level. Some developers are trying to implement typechecking for templates. https://github.com/vuejs/vetur/pull/681 However for now we are also thinking to move our templates to `tsx` to resolve it.

### CODE REVIEW

File: app/components/widgets/DonationTicker.vue.ts [hunk 1]
```diff
@@ -12,11 +12,11 @@ import {
...
   IPlatformService
-} from '../../services/platforms';
-import { CustomizationService } from '../../services/customization';
+} from '../../services/platforms/index';
+import { CustomizationService } from '../../services/customization/index';
 import Raven from 'raven-js';
-import { AppService } from 'services/app';
-import { SceneCollectionsService } from 'services/scene-collections';
+import { AppService } from 'services/app/index';
+import { SceneCollectionsService } from 'services/scene-collections/index';
```
gettinToasty: do you need to explicitly refer to `/index` here?

app/components/widgets/EventList.vue.ts
```diff
@@ -12,11 +12,11 @@ import {
...
   IPlatformService
-} from '../../services/platforms';
-import { CustomizationService } from '../../services/customization';
+} from '../../services/platforms/index';
+import { CustomizationService } from '../../services/customization/index';
 import Raven from 'raven-js';
-import { AppService } from 'services/app';
-import { SceneCollectionsService } from 'services/scene-collections';
+import { AppService } from 'services/app/index';
+import { SceneCollectionsService } from 'services/scene-collections/index';
```
gettinToasty: see above

app/components/widgets/SpinWheel.vue.ts
```diff
@@ -12,11 +12,11 @@ import {
...
   IPlatformService
-} from '../../services/platforms';
-import { CustomizationService } from '../../services/customization';
+} from '../../services/platforms/index';
+import { CustomizationService } from '../../services/customization/index';
 import Raven from 'raven-js';
-import { AppService } from 'services/app';
-import { SceneCollectionsService } from 'services/scene-collections';
+import { AppService } from 'services/app/index';
+import { SceneCollectionsService } from 'services/scene-collections/index';
```
gettinToasty: see above

app/components/widgets/EventList.vue.ts
holiber: I personally wish to delete these files. It's not a good idea to store unused components in the repo.

gettinToasty: @holiber i agree with that

PR #655, opened 2018-07-26
fix sliders in filters

andycreeth:
Also fix VST plugin not working

### DIFF
--- a/app/components/windows/SourceFilters.vue
+++ b/app/components/windows/SourceFilters.vue
@@ -43,7 +43,7 @@
 
       <div class="modal-container--side-nav">
         <div v-if="selectedFilterName">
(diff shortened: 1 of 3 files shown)


app/components/windows/SourceFilters.vue
```diff
@@ -43,7 +43,7 @@
 
       <div class="modal-container--side-nav">
         <div v-if="selectedFilterName">
-          <GenericForm v-model="properties" @input="save"></GenericForm>
+          <GenericForm v-model="properties" @input="save" :key="selectedFilterName"></GenericForm>
```
andycreeth: This is an extra level of security.  It forces us to remake the entire form when selecting a different filter, at the expense of some performance when swapping filters.

PR #656, opened 2018-07-26
Refactor SourcesShowcase Part 2

gettinToasty:
Not a huge loss of lines here but makes things that we should iterate over more iterable and separates concerns between rendering and storing data. From what we learned here we should be able to apply a similar pattern to the non-widget sources which should remove >200 lines from `SourcesShowcase.vue` and make that component much more manageable.

This PR also refactors `services/widgets.ts`, splitting up the ~500 line godfile into more reasonable chunks: the main service, the api, and the data shapes


holiber:
We use `-api` postfix only for files that expose SLOBS API via interfaces. These files don't contain executable code and used by `typedoc` parser for auto-generated documentation. I'm thinking about to join `WidgetSettingsService` and `WidgetsService` to a single service after we finish with widgets

holiber:
Oh, looks like I misunderstood some your changes, nevermind

app/components/windows/SourcesShowcase.vue.ts
```diff
@@ -66,19 +66,19 @@ export default class SourcesShowcase extends Vue {
     }
   }
 
+  getSrc(type: string, theme: string) {
+    return require(`../../../media/source-demos/${theme}/${this.widgetData(type).demoFilename}`);
```
andycreeth: Still can't believe the webpack magic here.

PR #661, opened 2018-07-27
Refactor Standard Sources Showcase

gettinToasty:
the winner of this season's Biggest Loser: `SourcesShowcase.vue` from 850 lines down to a trim 333!


app/services/sources/sources-data.ts
```diff
@@ -0,0 +1,120 @@
...
+    description: $t('Captures your desktop audio for the purpose of playing sound, such as music or speech.'),
+    demoFilename: 'audio-output.png',
+    supportList: [$t('Desktop audio')]
+  },
+  // vvv Not sure if this is still supported, there is no showcase data for it vvv
+  text_ft2_source: {
+    name: $t('Text (FreeType 2)'),
+    description: ''
+  }
+});
```
holiber: I feel like this object is what `SourcesService.getAvailableSourcesTypesList` have to return

gettinToasty: possibly! i just didn't want to change too much (at least in the first version of this) in case that breaks something elsewhere

PR #665, opened 2018-07-27
<NumberInput @input /> to emit number instead of string of number (or "")

Nitsorn:


app/components/shared/inputs/NumberInput.vue.ts
```diff
@@ -1,16 +1,27 @@
 import { Component, Prop } from 'vue-property-decorator';
 import { BaseInput } from './BaseInput';
 import { INumberMetadata } from './index';
+import { String } from '../../../../node_modules/aws-sdk/clients/ssm';
```
holiber: ?

Nitsorn: fixed! sorry was a typo that vscode assumed was something else :(

andycreeth: haha automatic import in VSCode sometimes produces weird results lol

PR #675, opened 2018-08-01
add "inputStyle" to metadata for widget inputs

Nitsorn:
GOAL: to allow custom styling for widget input components by passing style object in the metadata.

Adding `metadata.inputStyle` as optional property of metadata object, responsible for adding additional styling to the <input /> element inside widget inputs. 

Reason:

I came across `width: 220px` in `<NumberInput />` scoped made by someone else who wanted to style the width of NumberInput in `EventList`. Since its scoped in the input component itself, it affected all number inputs globally. (ditto for ColorInput and FontSize)

What I did:

- Removed the scoped styling to make all inputs default to 100% width.
- Add `inputStyle: object` as valid key in `IInputMetadata` interface
- Update vue templates to read `metadata.inputStyle` or `options.inputStyle`
- For widget inputs that extends the other (FontSize extends SliderInput) merge the propped metadata with slider metadata.


andycreeth:
Just to confirm, we are abandoning the proposal in favor of Oat using separate wrapper components, right?

Nitsorn:
yes. That is correct. @gettinToasty can confirm.

andycreeth:
Cool, will close.

PR #679, opened 2018-08-01
Refactor tooltips

gettinToasty:


app/components/shared/inputs/BaseInput.ts
```diff
@@ -25,6 +26,10 @@ export abstract class BaseInput<TValueType, TMetadataType extends IInputMetadata
     return { required: this.options.required };
   }
 
+  fontFamilyTooltip = $t(
+    'The Google Font to use for the text. Visit http://google.com/fonts to find one! Popular Fonts include: Open Sans, Roboto, Oswald, Lato, and Droid Sans.'
+  );
```
andycreeth: This looks like a really specific tooltip.  Does this belong in `BaseInput`?

gettinToasty: every `type="fontFamily"` input will have it, wasn't really sure where else to put it to ensure it appears on every input

PR #684, opened 2018-08-03
Add SponsorBanner settings

gettinToasty:


app/components/widgets/inputs/ImageLayout.vue.ts
```diff
@@ -0,0 +1,25 @@
+import { Component, Prop } from 'vue-property-decorator';
+import { IListMetadata } from '../../shared/inputs';
+import { BaseInput } from 'components/shared/inputs/BaseInput';
+import ImagePicker from '../../shared/inputs/ImagePicker.vue';
+
+@Component({
+  components: { ImagePicker }
+})
+export default class AnimationInput extends BaseInput<string, IListMetadata<string>> {
```
holiber: AnimationInput -> ImageLayoutInput

app/components/widgets/inputs/ImageLayout.vue
```diff
@@ -0,0 +1,5 @@
+<template>
+  <image-picker :value="value" :metadata="meta" @input="value => emitInput(value)" />
+</template>
+
+<script lang="ts" src="./ImageLayout.vue.ts"></script>
```
holiber: We use `Input` postfix for the classes that we extend from BaseInput. I think it is a good convention to follow. I have the same comment for the  `image-picker`

PR #685, opened 2018-08-03
Fix accidental tooltip duplication

gettinToasty:


app/components/shared/inputs/index.ts
```diff
@@ -26,6 +26,7 @@ export interface IInputMetadata {
   hint?: string;
   type?: EInputType;
   title?: string;
+  tooltip?: string;
```
holiber: it looks like `hint` is the same like `tooltip` can you check that we use the same parameter name for the tooltips before we merge this?

gettinToasty: `hint` is used nowhere in our repo, for all of the tooltips we've been using `tooltip`

holiber: if so just delete `hint` from `IInputMetadata` interface

PR #689, opened 2018-08-04
Widgets: enable new widgets settings by default

holiber:
This is good for merge only after #685 - #688 will be completed


holiber:
This PR is good to merge when we will be ready to publish new widgets

PR #691, opened 2018-08-06
Add Test Buttons

gettinToasty:


gettinToasty:
@holiber those were just copy-pasted from the StreamBoss widget (TipJar has the same ones)

PR #694, opened 2018-08-06
Add Sound Input

gettinToasty:


app/components/windows/MediaGallery.vue.ts
```diff
@@ -155,14 +159,22 @@ export default class MediaGallery extends Vue {
   }
 
   selectFile(file: IMediaGalleryFile, select: boolean) {
-    this.selectedFile = file;
+    if (this.audioOnly && file.type !== 'audio') {
+      this.$toasted.show($t('Not a supported file type'), {
+        duration: 1000,
+        position: 'top-right',
+        className: 'toast-alert'
+      });
```
holiber: just use `return` here instead of additional `else` bracers

PR #696, opened 2018-08-06
Add widget specific source icons

gettinToasty:
widget_icons

### DIFF
--- a/app/components/SourceSelector.vue.ts
+++ b/app/components/SourceSelector.vue.ts
@@ -12,6 +12,42 @@ import SlVueTree, {
 } from 'sl-vue-tree';
 import { $t } from 'services/i18n';
 
+const widgetIconMap = {
+  'Alert Box': 'fas fa-bell',
+  'Stream Boss': 'fas fa-gavel',
+  'Event List': 'fas fa-th-list',
+  'The Jar': 'fas fa-beer',
+  'Stream Labels': 'fas fa-file-alt',
+  'Donation Ticker': 'fas fa-ellipsis',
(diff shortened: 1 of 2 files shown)


app/components/SourceSelector.vue.ts
```diff
@@ -12,6 +12,42 @@ import SlVueTree, {
...
 
+const widgetIconMap = {
+  'Alert Box': 'fas fa-bell',
+  'Stream Boss': 'fas fa-gavel',
+  'Event List': 'fas fa-th-list',
+  'The Jar': 'fas fa-beer',
+  'Stream Labels': 'fas fa-file-alt',
+  'Donation Ticker': 'fas fa-ellipsis',
+  'Chat Box': 'fas-fa-comments',
+  'Viewer Count': 'fas fa-eye',
```
holiber: we use the same icon for the visible/invisible scene items, how it's looking now?

app/components/SourceSelector.vue.ts
```diff
@@ -126,6 +168,7 @@ export default class SourceSelector extends Vue {
   }
 
   makeActive(treeNodes: ISlTreeNode<ISceneItemNode>[], ev: MouseEvent) {
+    console.log(this.nodes);
```
holiber: console.log

app/components/SourceSelector.vue.ts
gettinToasty: it's actually a slightly different icon, let me post a screenshot

holiber: Oh, I see. We used the same icon before redesign but now it's different

PR #698, opened 2018-08-07
Various fixes for the design system CSS refactor

mbiemiller:
- add source scroll fix
- source filters side menu improvement
- updater window style update
- performance mode style fix
- more typography fixes
- more general settings fixes
- changed media query so top nav links keep text showing for longer
- handful of various UI improvements/fixes


app/components/LiveDock.vue
```diff
@@ -196,33 +197,49 @@
   }
 }
 
+.live-dock-viewer-count__count {
+ padding-right: 3px;
```
andycreeth: Seems like only 1 space indent here.

app/components/LiveDock.vue
```diff
@@ -38,14 +38,13 @@
...
             :class="{
-              'fa-eye': !hideViewerCount,
-              'fa-eye-slash': hideViewerCount
+              'icon-view': !hideViewerCount,
+              'icon-hide': hideViewerCount
             }"
             @click="toggleViewerCount"/>
-          <i class="fa fa-user label--icon" />
-          <span class="weight--medium">{{ viewerCount }}</span> {{ $t('viewers')}}
+          <!-- <i class="icon-team label--icon" /> -->
```
andycreeth: Are we removing this icon intentionally?  It looks like you updated the icon class but commented it out, so I wanted to make sure this was intended.

app/styles/mixins.less
```diff
@@ -74,6 +74,39 @@
   margin-right: 8px;
 }
 
+// Margins
+.margin-top() {
+  margin-top: 8px;
+}
```
andycreeth: I'm wondering if it might make sense to do this with variables and a little math:

```less
.margin-top(@x: 1) {
  margin-top: (@x * 1)px;
}
```
(not sure if that syntax is exactly right, but should be possible).
The advantages:
- Is a single mixin to remember
- Enforces only applying in 8 pixel increments

mbiemiller: yeah definitely will switch to this after this PR

holiber:
The icons for ChatBox and Donation Ticker are not shown 
image

holiber:
The filters window looks a bit weird
image

holiber:
Weird `Downmix to mono` checkbox position
image

PR #702, opened 2018-08-08
Fix bug where renamed widgets would not display the correct icon

gettinToasty:
source_icon_fix

### DIFF
--- a/app/components/SourceSelector.vue.ts
+++ b/app/components/SourceSelector.vue.ts
@@ -10,25 +10,23 @@ import SlVueTree, {
   ISlTreeNodeModel,
   ICursorPosition
 } from 'sl-vue-tree';
+import { WidgetType } from 'services/widgets';
 import { $t } from 'services/i18n';
 
 const widgetIconMap = {
-  'Alert Box': 'fas fa-bell',
-  'Stream Boss': 'fas fa-gavel',
-  'Event List': 'fas fa-th-list',
-  'The Jar': 'fas fa-beer',
-  'Stream Labels': 'fas fa-file-alt',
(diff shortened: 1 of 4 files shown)


app/components/SourceSelector.vue.ts
```diff
@@ -91,10 +90,12 @@ export default class SourceSelector extends Vue {
...
   }
 
-  determineIcon(isLeaf: boolean, data: { type: string, name: string }) {
+  determineIcon(isLeaf: boolean, sourceId: string) {
     if (!isLeaf) { return 'fa fa-folder'; }
-    if (data.type === 'browser_source') { return widgetIconMap[data.name] || 'fas fa-globe'; }
-    return sourceIconMap[data.name] || 'fas fa-file';
+    const sourceDetails = this.sourcesService.getSource(sourceId).getComparisonDetails();
+    if (sourceDetails.isStreamlabel) { return 'fas fa-file-alt'; }
+    if (sourceDetails.widgetType) { return widgetIconMap[sourceDetails.widgetType]; }
```
andycreeth: So you can double check this, but I'm pretty sure this will not work for the alert box.  Our widget type enum uses numbers instead of strings, because at the time TypeScript didn't support string enums.  `0` is the value assigned to the alert box.  Since `0` is falsy, this check will fail and it will show the browser source icon instead.

I think the best test to use here would be:
```TypeScript
if (sourceDetails.widgetType != null) { return widgetIconMap[sourceDetails.widgetType]; }
```
The `!=` makes sure it matches both `null` and `undefined` but not `0`

gettinToasty: good catch! yea the way it's implemented now does not account for the 0

PR #703, opened 2018-08-08
Fix keys in HotkeyGroup

gettinToasty:


app/components/HotkeyGroup.vue
```diff
@@ -14,7 +14,7 @@
     {{ title }}
   </h2>
   <div class="section-content" v-show="!collapsed">
-    <div v-for="hotkey in hotkeys" :key="hotkey">
+    <div v-for="hotkey in hotkeys" :key="`${hotkey.actionName}${hotkey.sceneId}${hotkey.sourceId}${hotkey.sceneItemId}`">
```
holiber: Hotkey class use `@ServiceHelper` decorator. So as an option, you can use `resourceId` as a unique id that this decorator appends to an instance.

gettinToasty: is there a reason `resourceId` isn't used in the `isSameHotkey` function in the hotkeys service?

holiber: `resourceId` was added when we exposed our API for the streamdeck team. The hotkeys code was added before

holiber: The current solution works fine. It's just FYI

PR #705, opened 2018-08-09
Widgets: custom fields editor

holiber:
image
image


app/components/widgets/CustomFieldsEditor.vue.ts
```diff
@@ -0,0 +1,224 @@
...
+    type: 'dropdown',
+    options: {
+      optionA: 'Option A',
+      optionB: 'Option B',
+      optionC: 'Option C'
+    },
+    value: 'optionB'
+  },
+
+  // TODO:
```
gettinToasty: Is there are reason these are commented out? SoundInput is in and and think `image-input` is the just the MediaGalleryInput

holiber: does it have an option to allow to select only images? Didn't see this option

gettinToasty: nope! but afaik neither does the site have that. now that `sound-input` is in place though basically all the architecture is there to add that validation. if that's blocking i can do it now, otherwise i think we would probably be ok just using `media-gallery-input`

holiber: Yes, let's add a filter option for media gallery to show only images. It's not a blocker since we can add it here in the separated PR

PR #706, opened 2018-08-09
slobs chatbot complete except validation

Nitsorn:
This PR contains the most updated work on SLOBS Chatbot.
All functionalities are complete, except for form validations.


andycreeth:
Lol there's a 10k line `yarn-error.log` in here.

Nitsorn:
lol yeah im removing @andycreeth

app/components/page-components/Chatbot/module-bases/ChatbotModToolsBase.vue.ts
```diff
@@ -0,0 +1,273 @@
...
+        return 'Symbols';
+      case 'links':
+        return 'Links';
+      default:
+        return 'unpermitted value';
+    }
+  }
+
+  // metadata
+  generalMetadata(protectionType: string) {
```
holiber: This method has a very weak type-checking. It allows returning any type of object here. There are some things you can do to improve that:

1) Add a returning type
`getMetadata(): Dictionary<IInputMetadata>`
2) Use `metadata` helper to have strong type protection
```
        duration: metadata.number({
          required: true,
          placeholder: $t('Punishment Duration in minutes'),
          min: 0
        })
```

Don't forget translations `$t()`

app/components/page-components/Chatbot/windows/ChatbotCustomCommandWindow.vue.ts
```diff
@@ -0,0 +1,160 @@
...
+
+@Component({
+  components: {
+    TextInput,
+    TextAreaInput,
+    ListInput,
+    TimePickerInput,
+    ChatbotAliases,
+  }
+})
```
holiber: Use a shorter way to import input-components:
```
import { inputComponents } from 'components/widgets/inputs';
@Component({
  components: {
    ...inputComponents,
    ChatbotAliases,
  }
})
```

app/components/page-components/Chatbot/windows/ChatbotCustomCommandWindow.vue.ts
```diff
@@ -0,0 +1,160 @@
...
+    cooldowns: {
+      global: 0,
+      user: 0
+    },
+    aliases: [],
+    platforms: 7,
+    enabled: true
+  };
+
+  tabs: { name: string; value: string }[] = [
```
holiber: that seems to be `ITab[]`

app/components/page-components/Chatbot/windows/ChatbotDefaultCommandWindow.vue.ts
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  get defaultCommandToUpdate() {
+    return this.chatbotCommonService.state.defaultCommandToUpdate;
+  }
+
+  // metadata
+  get metadata() {
+    let metadata: IDefaultCommandMetadata = {
+      command: this.stringMetadata(
```
holiber: `import { metadata } from 'components/widgets/inputs';`
`this.stringMetadata()` -> `metadata.string()`

app/components/shared/inputs/TimePickerInput.vue.ts
```diff
@@ -0,0 +1,33 @@
...
+  ss?: string;
+}
+
+@Component({
+  components: {
+    VueTimepicker
+  }
+})
+export default class TimePickerInput extends BaseInput<ITimeInputValueMetadata, ITimeMetadata> {
+  @Prop() value: ITimeInputValueMetadata;
```
holiber: That mixed me up. The value here is more `data` than `metadata`. https://en.wikipedia.org/wiki/Metadata
I think you need something like `ITimeInputValue` or `ITime`

app/components/shared/inputs/TimePickerInput.vue.ts
```diff
@@ -0,0 +1,33 @@
...
+  a?: string;
+  A?: string;
+  h?: string;
+  kk?: string;
+  k?: string;
+  m?: string;
+  mm?: string;
+  s?: string;
+  ss?: string;
+}
```
holiber: Why we need this object instead of a more traditional timestamp or Date() ?

.babelrc
```diff
@@ -2,6 +2,7 @@
   "presets": ["stage-2"],
   "plugins": [
     "transform-decorators-legacy",
-    "transform-class-properties"
+    "transform-class-properties",
+    "transform-runtime"
```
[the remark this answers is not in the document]
Nitsorn: vue2-timepicker. @andycreeth looked into it and think we will push it through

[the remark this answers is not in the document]
andycreeth: Yeah I'm not super happy about it, but this Vue library Oat is using requires this babel plugin to work properly.

app/components/page-components/Chatbot/ChatbotBase.vue.ts
```diff
@@ -1,36 +1,46 @@
...
+    // pre-load them to switch between 2 windows
+    this.chatbotApiService.fetchDefaultCommands();
+    this.chatbotApiService.fetchLinkProtection();
+  }
+
   get chatbotPermissions() {
-    let permissions = Object.keys(ChatbotPermissionsEnums)
-      .reduce((a: any[], b: string) => {
+    let permissions = Object.keys(ChatbotPermissionsEnums).reduce(
+      (a: any[], b: string) => {
```
andycreeth: You should define an interface in this file that describes the shape of that data you're returning here.  It would look something like:
```TypeScript
interface IChatbotPermissions {
  title: string;
  value: number;
}

(a: IChatbotPermissions[], b: string) => {
```

This is important because the return value of your getter is currently `any[]`, which makes it extremely difficult for other parts of the app to know how to use it, and you lose all type checking for anything that calls this.

app/components/page-components/Chatbot/ChatbotModTools.vue.ts
```diff
@@ -19,6 +18,8 @@ export default class ChatbotModTools extends ChatbotBase {
     this.chatbotApiService.fetchSymbolProtection();
     this.chatbotApiService.fetchLinkProtection();
     this.chatbotApiService.fetchWordProtection();
+
+    // this.chatbotCommonService.bindsToasted(this.$toasted);
```
andycreeth: Let's remove this if it's no longer needed.  If there's a reason to leave it in here but commented out, then let's leave a comment explaining what it is and why someone would uncomment it.

app/components/page-components/Chatbot/ChatbotTimers.vue
```diff
@@ -89,23 +96,35 @@ tbody tr {
   }
 }
 
+.chatbot-timers__timer-actions__container {
+  * > button {
```
andycreeth: I'm not a CSS or LESS expert, but this strikes me as a weird rule to write.  Doesn't this just match any `button` that is a child of literally any other element?  If that's the case, isn't the `* >` completely unnecessary?  Maybe you can explain what this is matching.

app/components/page-components/Chatbot/windows/ChatbotCustomCommandWindow.vue.ts
```diff
@@ -0,0 +1,154 @@
...
+    ChatbotAliases,
+  }
+})
+export default class ChatbotCustomCommandWindow extends ChatbotWindowsBase {
+  newCommand: ICustomCommand = {
+    command: null,
+    response: null,
+    response_type: 'Chat',
+    permission: {
+      level: 163,
```
andycreeth: Is this a number that could be constructed with `EChatbotPermissions` to make it more obvious where this magic number is coming from?

app/components/page-components/Chatbot/windows/ChatbotWindowsBase.vue.ts
```diff
@@ -13,4 +13,25 @@ Vue.use(VModal);
...
+  toggleLinkProtectionWindow() {
+    const currentWindow = this.chatbotCommonService.windowsService.getChildWindowOptions().componentName;
+
+    switch (currentWindow) {
+      case 'ChatbotDefaultCommandWindow':
+        this.chatbotCommonService.openLinkProtectionWindow();
+        break;
+      case 'ChatbotLinkProtectionWindow':
+        const linkProtectionPermitCommand =
+        this.chatbotApiService.state.defaultCommandsResponse['link-protection'].permit;
```
andycreeth: I think this should probably be indented another level since it's a line continuation.

app/components/shared/inputs/TimePickerInput.vue.ts
andycreeth: It looks like maybe this is the format expected by `vue2-timepicker`?  If that's the case I really think we should try to find another library soon.  😃

holiber: We came to a decision to use milliseconds to store time and intervals. Milliseconds have no problems with serialization like a Date object and we can convert it to any format we want via moment.js

PR #710, opened 2018-08-09
fix font selector

holiber:


holiber:
@gettinToasty we don't need fonts to use it outside this component now. It's nothing hard to change it if we will need.

PR #711, opened 2018-08-09
Disable FollowerGoal for YT users

gettinToasty:


app/services/widgets/widgets-api.ts
```diff
@@ -46,7 +46,7 @@ export interface IWidget {
 export interface IWidgetDisplayData {
   name: string;
   description: string;
-  platform?: string;
+  platforms?: Set<string>;
```
holiber: *-api.ts files contains only serializable data types. The javascript's Set is not serializable. You can check it: `JSON.stringify( new Set([1, 2, 3, 4, 5]))`

gettinToasty: hmm ok, should i just send that interface to `widget-data` then? it's never used in ipc or window communication, just for rendering

holiber: yes, that will work

PR #714, opened 2018-08-10
Face mask fixes

kuhasuki:
Use new attribute names
Disable transitions

### DIFF
--- a/app/services/facemasks.ts
+++ b/app/services/facemasks.ts
@@ -463,10 +463,8 @@ export class FacemasksService extends PersistentStatefulService<IFacemasksServic
             {
               maskFolder: this.facemasksDirectory,
               alertDuration: this.settings.duration,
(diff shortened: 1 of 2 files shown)


app/services/facemasks.ts
```diff
@@ -213,9 +213,11 @@ export class FacemasksService extends PersistentStatefulService<IFacemasksServic
   }
 
   trigger(mask: string, message: string, name: string) {
+    const clean = this.profanitize(message, name);
```
andycreeth: I know that `profanitize` is a portmanteau of `profanity` and `sanitize`, but it sounds like it's going to sprinkle some profanity into whatever string you pass it.  We should write that function some day.  `function realProfanitize(msg)`.  😄

kuhasuki: Yeah it does sound like some kind of markov profanity generator, maybe call it `deProfanitize`

PR #723, opened 2018-08-10
Add news banner

gettinToasty:
The only thing really left to be tested here is the actual API returning successfully as expected, but the issue is I'm not sure how to do that without actually making news banners via the admin tool... so I don't want nonsense test banners to show up to actual users if/when we push this up. The other outstanding issue with this is how to handle translation but i think that'll need to be handled from the site's end, since there's no way for us to predict what banner text may contain.

Edit: Other point I'm not 100% sure on is the stream end api call, would like some confirmation if that's indeed the correct place to put it


gettinToasty:
Update: Currently being tested by Eric Freytag but should be feature-complete at this point as long as everything works

app/services/news-banner.ts
```diff
@@ -0,0 +1,85 @@
...
+interface INewsBannerInfo {
+  id: number;
+  header: string;
+  subHeader: string;
+  link: string;
+  linkTitle: string;
+  thumbnail: string;
+}
+
+export class NewsBannerService extends StatefulService<INewsBannerInfo> {
```
holiber: Maybe `NewsService` or `AnnouncementsService` ? `Banner` is a good name for UI component but our services usually combine wider logic

app/services/news-banner.ts
```diff
@@ -0,0 +1,85 @@
...
+      await fetch(req);
+      this.CLEAR_BANNER();
+    } catch (e) {
+
+    }
+  }
+
+  private formRequest(endpoint: string, options: any = {}) {
+    const host = this.hostsService.streamlabs;
+    const headers = authorizedHeaders(this.userService.apiToken, options.headers);
```
holiber: Do we show news only for authorized users?

gettinToasty: I don't think the spec said so, but I'll clarify with Eric

gettinToasty: @holiber it should be able to work for non-logged in users too, is there a way to generate auth-ed headers without a logged in user?

holiber: There is no way to do that because it can cause serious security issues.
If we want to show this banner for unauthorized users we must rid of authorization for `https://streamlabs.com/api/v5/slobs/announcement/get`

However as I see you are disabling this banner by calling an API request. That makes me feel this API is not intended to work with unauthorized clients. I think you need to ask somebody from stemlabs backend team to resolve this or show this banner only for authorized users

gettinToasty: confirmed that its fine for only logged in users

app/services/announcements.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  closeBanner() {
+    this.postBannerClose();
+  }
+
+  private async fetchBanner() {
+    const endpoint = `api/v5/slobs/announcement/get?clientId=${this.userService.getLocalUserId()}`
+    const req = this.formRequest(endpoint);
+    try {
+      const newState = await fetch(req).then((rawResp) => rawResp.json());
```
andycreeth: I would recommend using the `handleErrors` helper in `util/requests` to make sure that the request actually succeeded.  The `fetch` API only rejects the promise when there is some sort of network error that prevents the request from going out.  As long as you get a response, the promise is resolved, even if the server returned a 502 for example.  The `handleErrors` helper just makes sure it's a 2xx reponse and rejects the promise otherwise.  More details here: https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch

All that said, I think your code here does do the correct thing in all situations, since in an error scenario, it will either fail to parse the JSON (promise will be rejected), or it will be a JSON object without an `id` in which case the line below will properly handle the error.  So I guess there's no action needed, this is more of an FYI in case you didn't know about that helper.

PR #734, opened 2018-08-15
Minimize our JS bundles for production

gettinToasty:
The mangling of filenames was preventing the slobs launch for some reason, by turning this off we sacrifice peak minimizing efficiency but still get a very good ROI

With mangle: 8.05 -> 3.4MB
Without mangle: 8.05 -> 5.13MB

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,5 +1,6 @@
 const CircularDependencyPlugin = require('circular-dependency-plugin');
 const path = require('path');
+const UglifyJsPlugin = require('uglifyjs-webpack-plugin')
 
 const plugins = [];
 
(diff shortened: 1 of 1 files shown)


webpack.config.js
```diff
@@ -116,7 +117,8 @@ module.exports = {
   },
 
   optimization: {
-    minimize: false
+    minimize: process.env.NODE_ENV === 'production',
```
andycreeth: This should not be needed.  As of webpack 4, production builds are automatically minimized.

gettinToasty: good point, that's just a holdover from when i was forcing minimize on dev env

PR #735, opened 2018-08-15
project: Directly access OBS in Renderer processes

Xaymar:
As we now have our own IPC framework, we no longer have to rely on electron's IPC framework to do our work for us. We can directly call things from any process, as long as that process is actually connected to the IPC server.

For OBS IPC, this means a lot of data can be directly accessed without blocking both the renderer and main process, thus freeing up resources for more responsiveness and faster rendering. As a side effect, we might see lower CPU usage too.


Xaymar:
Currently blocked by (and waiting on a tagged build containing): https://github.com/stream-labs/obs-studio-node/pull/73

app/app.ts
```diff
@@ -26,10 +25,19 @@ import VeeValidate from 'vee-validate';
...
 
 const { ipcRenderer, remote } = electron;
-
 const slobsVersion = remote.process.env.SLOBS_VERSION;
 const isProduction = process.env.NODE_ENV === 'production';
 
+window["obs"] = window['require']('obs-studio-node');
+
+{ // Set up things for IPC
+  debugger;
```
gettinToasty: debugger

Xaymar: Force push didn't go through before creating the PR, it's gone now.

app/app.ts
```diff
@@ -26,10 +25,18 @@ import VeeValidate from 'vee-validate';
 import ChildWindow from 'components/windows/ChildWindow.vue';
 
 const { ipcRenderer, remote } = electron;
-
 const slobsVersion = remote.process.env.SLOBS_VERSION;
 const isProduction = process.env.NODE_ENV === 'production';
 
+window["obs"] = window['require']('obs-studio-node');
```
gettinToasty: not sure why you're putting this on the window here? it doesn't look like you're accessing it in any other files. also, dot notation should be fine here eg `window.obs = require('obs-studio-node')`

Xaymar: I'm putting it on the window because that's where it was before (see further below for the deleted line). I did not want to change too many things in order to avoid any potential for newly added broken things.

window.obs is underlined red for me, which usually indicates that typescript will complain about it. I fixed the " and ' though now.

Xaymar:
The changes in actual behavior seem to be very subtle, barely noticeable here. If everything worked, it should be almost identical for most things, but the main process is no longer blocked for any renderer OBS requests. In the ideal case, the app should feel like it responds quicker to user input.

Renderer-only IPC:
image

Main-Process IPC:
image

No IPC:
image

app/app.ts
```diff
@@ -26,10 +25,18 @@ import VeeValidate from 'vee-validate';
...
 
+window['obs'] = window['require']('obs-studio-node');
+
+{ // Set up things for IPC
+  // Connect to the IPC Server
+  window['obs'].IPC.connect(process.env.SLOBS_IPC_PATH);
+  document.addEventListener('close', (e) => {
+    window['obs'].IPC.disconnect();
+  });
+}
```
andycreeth: What's the purpose of this block here?

app/app.ts
```diff
@@ -26,10 +25,18 @@ import VeeValidate from 'vee-validate';
...
 const { ipcRenderer, remote } = electron;
-
 const slobsVersion = remote.process.env.SLOBS_VERSION;
 const isProduction = process.env.NODE_ENV === 'production';
 
+window['obs'] = window['require']('obs-studio-node');
+
+{ // Set up things for IPC
+  // Connect to the IPC Server
+  window['obs'].IPC.connect(process.env.SLOBS_IPC_PATH);
```
andycreeth: Theoretically our child windows shouldn't need connections to the OBS server.  That said there may be a spot or two where we accidentally call directly into OBS from the child window.  If you move this into `app/services/app.ts` (so only the main window has a connection) do things start breaking?

app/services/sources/sources.ts
```diff
@@ -314,7 +299,20 @@ export class SourcesService extends StatefulService<ISourcesState> implements IS
       activeItems.forEach(activeItem => {
         sourcesNames.push(activeItem.sourceId);
       });
-      ipcRenderer.send('requestSourceAttributes', sourcesNames);
+
+      const sizes: obs.ISourceSize[] = obs.getSourcesSize(sourcesNames);
```
andycreeth: Is this synchronous?  The main reason we did the the weird thing with events to the main process is because we do this every second in the background and it's very disruptive to do it synchronously.

main.js
```diff
@@ -56,27 +57,24 @@ let appShutdownTimeout;
...
 
+  { // Initialize obs-studio-server
+    // Set up environment variables for IPC.
+    process.env.SLOBS_IPC_PATH = "slobs-".concat(uuid());
+    process.env.SLOBS_IPC_WORKING_DIRECTORY = path.join(app.getAppPath().replace('app.asar', 'app.asar.unpacked') + '/node_modules/obs-studio-node');
+    process.env.SLOBS_IPC_USERDATA = app.getPath('userData');
+    // Host a new IPC Server and connect to it.
+    obs.IPC.ConnectOrHost(process.env.SLOBS_IPC_PATH);
+    obs.NodeObs.SetWorkingDirectory(process.env.SLOBS_IPC_WORKING_DIRECTORY);
+  }
```
andycreeth: What's with this block?

app/app.ts
andycreeth: Also given that this environment variable is set in the main process, I don't think you can guarantee that it will always be here this early.  It's safer to use `electron.remote.process.env` instead I think.

main.js
```diff
@@ -56,27 +57,24 @@ let appShutdownTimeout;
...
-}
-
-
 function startApp() {
   const isDevMode = (process.env.NODE_ENV !== 'production') && (process.env.NODE_ENV !== 'test');
 
+  { // Initialize obs-studio-server
+    // Set up environment variables for IPC.
+    process.env.SLOBS_IPC_PATH = "slobs-".concat(uuid());
+    process.env.SLOBS_IPC_WORKING_DIRECTORY = path.join(app.getAppPath().replace('app.asar', 'app.asar.unpacked') + '/node_modules/obs-studio-node');
```
andycreeth: It looks like you're calling `path.join` with only a single argument.  I think you meant to use a `,` instead of a `+`.  Also it's best to just pass each individual component `node_modules`, `obs-studio-node` etc as individual arguments to `path.join` and let it worry about separators.

app/services/app/app.ts
```diff
@@ -68,6 +69,9 @@ export class AppService extends StatefulService<IAppState> {
   load() {
     this.START_LOADING();
 
+    // Initialize OBS
+    obs.NodeObs.OBS_API_initAPI('en-US', process.env.SLOBS_IPC_USERDATA);
```
andycreeth: Same comment about environment variable availability.

main.js
```diff
@@ -56,27 +57,24 @@ let appShutdownTimeout;
...
+    process.env.SLOBS_IPC_USERDATA = app.getPath('userData');
+    // Host a new IPC Server and connect to it.
+    obs.IPC.ConnectOrHost(process.env.SLOBS_IPC_PATH);
```
andycreeth: Does this mean that the main process still maintains an OBS connection?  Are we eventually trying to move away from that?

obs-api/index.js
```diff
@@ -1,25 +1,8 @@
...
-exports.VideoFactory = obs.VideoFactory;
-exports.ModuleFactory = obs.ModuleFactory;
-exports.addItems = obs.addItems;
-exports.createSources = obs.createSources;
-exports.getSourcesSize = obs.getSourcesSize;
+var obs = window['require']('obs-studio-node');
+
+/* Use for...in operator to perfectly mirror the osn module */
+for (let entry in obs) {
+    exports[entry] = obs[entry];
```
andycreeth: Nice!  2 space indent btw

app/app.ts
andycreeth: Yeah we've had this here exclusively for testing in the dev environment.  It's mainly been used by Eddy to test new NodeObs methods he's been working on in the context of SLOBS from a dev console.  It may be no longer needed, not really sure.

main.js
Xaymar: Ideally yes, but the main process is the only one that seems to be able to survive an uncaught v8 exception. All renderer processes basically freeze when they have an uncaught exception, so I wasn't able to have code execute after it.

app/services/sources/sources.ts
Xaymar: It would previously block the main process. Now it blocks the renderer for ~20µs or so. I have a branch for obs-studio-node which actually makes the function asynchronous, if we want to go for that.

app/app.ts
Xaymar: > If you move this into app/services/app.ts (so only the main window has a connection) do things start breaking?

Yes, because the child window has no connection to the obs ipc server. I did not write a layer that talks to the main window renderer for this, instead I have the child windows also connect to the ipc server.

This should not have any negative impact at all, it should actually improve things. In the ideal case, I want to avoid having to block any other v8 context other than the one the command is ran on. OBS handles almost all synchronization needs, so there is not much to worry about there.

> It's safer to use electron.remote.process.env instead I think.

You are correct, I will fix that.

main.js
Xaymar: It starts and connects to the IPC server, then sets the proper working directory.

main.js
Xaymar: Yup you're right, this is a leftover from the rebase mess that happened where I had a conflict in every single file. Will fix it.

main.js
Xaymar: Also it is the one running the crash handler, so it can safely shut down the server if that happens. Except if the main process crashes that is, then it can't.

app/app.ts
andycreeth: I was more curious about why the unnamed block `{` `}`.  It isn't serving any purpose for the code.  Is this just meant to indicate these statements are a logical group?  I can't say I've ever seen anyone use unnamed blocks in javascript like this before.

app/services/sources/sources.ts
andycreeth: Yeah I was talking to Eddy, and I think eventually it would be awesome if the IPC client supported calling functions both sync and async.  Then in the frontend we can make that decision whether we want to block or not.

For now I think it's fine as long as we test this with a very large scene collection (50 - 100 sources) on a moderate speed computer and see how it compares to our current live release.

app/app.ts
Xaymar: Yep, it's just to separate the logical execution of the code. I've recently started using this notation after learning rust (language) and at least for me it makes understanding the code much easier if i come back to it later. It serves no other purpose, so I can remove it if you want.

Xaymar:
The issues with the original commit have now been fixed. It should no longer crash on shut down and the env variables should now be read from the remote process.

However there still seems to be an issue with the onboarding step. For me the onboarding step froze while adding widgets, which is not something it should do to my knowledge.

Xaymar-Streamlabs:
The onboarding step is fixed in the latest obs-studio-node version.

computerquip-streamlabs:
Closed in favor of #882

PR #737, opened 2018-08-15
Slobs Chatbot V1.0

Nitsorn:
slobs chatbot v1.0 to preview build


app/services/chatbot/chatbot-interfaces.ts
```diff
@@ -0,0 +1,338 @@
...
+  result: string;
+  tags: string[];
+}
+
+// timers
+export interface ITimersData {
+  [id: number]: ITimer;
+}
+
+export interface ITimer {
```
holiber: this name is too general, let's rename it to something like IChatBotTimer

app/styles/inputs.less
```diff
@@ -218,6 +218,28 @@ textarea {
   }
 }
 
+.time-picker {
```
holiber: we don't need timepicker now, right?

app/styles/mixins.less
```diff
@@ -156,6 +156,118 @@
...
+}
+
+.margin-vertical--10 {
+  margin: 10px 0;
+}
+
+
+// widths
+.width--100 {
+  width: 100% !important;
```
holiber: do we really need `!important` here?

PR #738, opened 2018-08-16
fix flicker issues (first pass)

andycreeth:
This is a first pass at overhauling our child window and removing 99% of the flicker and replacing it with a spinner.


main.js
```diff
@@ -363,19 +359,6 @@ ipcMain.on('window-showChildWindow', (event, windowOptions) => {
...
-  new Promise(resolve => {
-    if (childWindowIsReadyToShow) {
-      resolve();
-      return;
-    }
-    ipcMain.once('window-childWindowIsReadyToShow', () => resolve());
-  }).then(() => {
-    // The child window will show itself when rendered
-    childWindow.send('window-setContents', windowOptions);
-  });
```
andycreeth: We're able to remove this now because of the new system of updating the child window contents by setting it in the main window and then waiting for it to get synced to the child window.

app/components/windows/ChildWindow.vue
```diff
@@ -1,12 +1,46 @@
 <template>
-<div style="height: 100%">
-    <!--
-        we use 2 components for situation when we need to open a new window and
-        save the previous window-component state
-    -->
-    <component v-if="components[0].name" v-show="components[0].isShown" :is="components[0].name"/>
-    <component v-if="components[1].name" v-show="components[1].isShown" :is="components[1].name"/>
+<div style="height: 100%" :class="{'night-theme': nightTheme, 'day-theme': !nightTheme}">
```
gettinToasty: is there a reason we set the theme explicitly here?

app/components/windows/ChildWindow.vue.ts
```diff
@@ -64,10 +59,15 @@ export default class ChildWindow extends Vue {
...
-    });
+    // This is essentially a race condition, but make a best effort
+    // at having a successful paint cycle before loading a component
+    // that will do a bunch of synchronous IO.
+    setTimeout(() => {
+      this.$set(this.components, this.activeComponentInd, {
+        name: this.options.componentName,
+        isShown: this.options.isShown
+      });
+    }, 50);
```
gettinToasty: would `requestAnimationFrame` work better here? i think that waits like exactly one paint cycle, but i might be wrong about that

app/components/windows/ChildWindow.vue
andycreeth: It's needed by the blank slate, which is above the ModalLayout (where we currently set it) in the hierarchy.  I tried removing it from the ModalLayout after this change, but it broke a ton of stuff and I'm not entirely sure why at the moment.

app/components/windows/ChildWindow.vue.ts
andycreeth: You're totally right, it does get called on every paint cycle.  However, it gets called right before the paint cycle as opposed to after.  I still think maybe there's something I could with it, but I'm not really sure.  Electron also has a paint event I'd like to play around with a bit more.  However I don't really have time to mess around with it more at the moment.

andycreeth: Haha interesting.  Theoretically that should work I think?  I'll test some more when I have time.

holiber: `Vue.$nextTick()` ?

andycreeth: @holiber `Vue.$nextTick()` unfortunately does not work.  `$nextTick()` only guarantees that the changes have been made in the DOM, but it doesn't guarantee that the browser has actually painted the changes to the screen.  My experimentation using `$nextTick()` often resulted in us blocking the pain to the screen with synchronous IO.

PR #739, opened 2018-08-16
Add ImageMediaInput

gettinToasty:
Just a filetype filtered version of the media gallery, @holiber i think this will unblock the rest of custom fields for you :)

### DIFF
--- a/app/components/shared/inputs/index.ts
+++ b/app/components/shared/inputs/index.ts
@@ -62,6 +62,7 @@ export interface IListOption<TValue> {
 export interface IMediaGalleryMetadata extends IInputMetadata {
   fileName: string;
   clearImage: string;
+  imageOnly?: boolean;
 }
 
 
--- a/app/components/shared/inputs/index.ts
+++ b/app/components/shared/inputs/index.ts
@@ -60,9 +60,8 @@ export interface IListOption<TValue> {
 }
 
 export interface IMediaGalleryMetadata extends IInputMetadata {
-  fileName: string;
   clearImage: string;
-  imageOnly?: boolean;
+  filter?: 'audio' | 'image';
 }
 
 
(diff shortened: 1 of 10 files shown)


app/components/shared/inputs/index.ts
```diff
@@ -62,6 +62,7 @@ export interface IListOption<TValue> {
 export interface IMediaGalleryMetadata extends IInputMetadata {
   fileName: string;
```
holiber: Why we need `fileName` here? We already have `value` property

app/services/media-gallery/media-gallery.ts
```diff
@@ -28,7 +28,8 @@ export interface IMediaGalleryInfo extends IMediaGalleryLimits {
 }
 
 interface IMediaGalleryProps {
-  audioOnly: boolean;
+  audioOnly?: boolean;
+  imageOnly?: boolean;
```
holiber: You can't set both, right? Maybe use something like `filter: {images: true, audio: true }` ?

app/components/shared/inputs/index.ts
gettinToasty: `fileName` is a display value whereas `value` is the href, but i suppose we could move the filename parsing into the input component that probably makes more sense

app/services/media-gallery/media-gallery.ts
gettinToasty: yea that's true, i'll look into making a better pattern here

app/components/shared/inputs/index.ts
holiber: Yeah, we probably just need a getter here.

holiber: To be clear

holiber: To be clear, the `metadata` property intended to store only values that are not going to be changed when you change the `value` property.

PR #742, opened 2018-08-16
Don't autoscroll when switching settings tabs

gettinToasty:


app/components/windows/Settings.vue.ts
```diff
@@ -79,6 +79,8 @@ export default class SceneTransitions extends Vue {
   @Watch('categoryName')
   onCategoryNameChangedHandler(categoryName: string) {
     this.settingsData = this.settingsService.getSettingsFormData(categoryName);
+    const settingsWindow = document.getElementsByClassName('settings-container')[0];
```
holiber: don't us global selectors. Use $el instead https://vuejs.org/v2/api/#vm-el

gettinToasty: @holiber how would i grab a subsection of that?

holiber: Just looked at what are you trying to do. Since `.settings-container` is a part of Settings.vue template you should use `refs` to access it. This is how Vue recommends accessing DOM elements. https://vuejs.org/v2/guide/components-edge-cases.html#Accessing-Child-Component-Instances-amp-Child-Elements

gettinToasty: the issue with using `$refs` in this instance is that TS blows up:
```
[tsl] ERROR in C:\Users\seanb\slobs\app\components\windows\Settings.vue.ts(82,34)
      TS2339: Property 'scrollTop' does not exist on type 'Vue | Element | Vue[] | Element[]'.
```

i could use `this.$el.querySelector` but that doesn't seem to have any great gains over `getElementsByClassName`

app/components/windows/Settings.vue.ts
```diff
@@ -79,6 +79,8 @@ export default class SceneTransitions extends Vue {
   @Watch('categoryName')
   onCategoryNameChangedHandler(categoryName: string) {
     this.settingsData = this.settingsService.getSettingsFormData(categoryName);
+    const el = this.$refs.settingsContainer as HTMLElement;
```
holiber: define `refs` as a component property
```
$refs: {
 settingsContainer: HTMLElement;
}
```

gettinToasty: i'm not sure where i would put that snippet in the code

holiber: https://github.com/vuejs/vue-class-component/issues/94

PR #744, opened 2018-08-16
Add uglify-es

eddyStreamlabs:


andycreeth:
This is not the correct way to add a dependency.  This should be a `devDependency` not a `resolution`.

andycreeth:
Nvm confirmed that we're using this to force a downgrade to work around an open compatibility issue.  All good.

PR #761, opened 2018-08-21
Design system

mbiemiller:
- overhaul of mixins and variables
- continued swapping out mixins and reducing helper classes
- moving to new padding & margin defaults throughout the app for a more condensed feel
- animation of collapsible sections
- small polish fixes - transitions, active states, hover states
- input updates - changed styles to differentiate inputs & dropdowns + more contrast on dropdowns
-  removed 'Essential' tag from everything except Alert Box
- removed old twitchalerts.css from vendor folder
- couple button color changes in Settings window and ChatBot
- styles for dropdowns that are against section bg
- moved a colorpicker and multiselect rules into global sheet so we don't have 2 style sets for OBS and regular


package.json
```diff
@@ -160,6 +160,7 @@
     "vue-property-decorator": "7.0.0",
     "vue-slider-component": "2.2.7",
     "vue-template-compiler": "2.5.16",
+    "vue-toasted": "^1.1.24",
```
andycreeth: It looks like this is in here twice now.  It got added out of alphabetical order at the end of the list.  Can you delete the one at the end of the list?

PR #762, opened 2018-08-21
Fix onboarding dropdown styling

gettinToasty:


app/components/pages/onboarding_steps/ObsImport.vue
```diff
@@ -40,11 +40,25 @@
...
 <script lang="ts" src="./ObsImport.vue.ts"></script>
 
+<style lang="less">
+@import '../../../styles/index';
+// 3rd Party Component
+.multiselect__content-wrapper {
+  left: -50%;
+  border-radius: 0 0 3px 3px;
+}
+</style>
```
andycreeth: I just noticed that this is an unscoped style block, which means it will apply globally to all multiselects.  If that's the case maybe it should be put somewhere more general in our of our app css files.

gettinToasty: i did some local testing and found our other multiselects seem to be unaffected by this for whatever reason (nor should they be), but making it a scoped style doesn't apply the styles to the unscoped classname

PR #764, opened 2018-08-21
add studio mode event and audio source updated event

andycreeth:


app/services/audio/audio.ts
```diff
@@ -197,6 +195,48 @@ export class AudioService extends StatefulService<IAudioSourcesState> implements
     });
   }
 
+  setSettings(sourceId: string, patch: Partial<IAudioSource>) {
+    const obsInput = this.sourcesService.getSourceById(sourceId).getObsInput();
+
+    // Fader is ignored by this method.  Use setFader instead
+    const newPatch = omit(patch, 'fader');
```
holiber: Does `omit` check that the second argument is `keyof Partial<IAudioSource>`?
If not then it looks like it's better to use just `delete patch.fader`;

andycreeth: It does not.  That said I was trying not to use `delete` because it mutates the object that was passed to the function with can lead to weird bugs.  That said, I could always clone first, and then delete from the cloned object.

PR #766, opened 2018-08-21
Slobs chatbot V1.4

Nitsorn:
Slobs chatbot V1.4
- Refactor to use VFormGroup instead of individual components
- Added tooltips
- Added CollapsibleSection shared component
- Added chatbot-api for restricting api from frontend
- update css table style
- update eventHandlers naming
- Abbreviate cooldowns in table headers


app/services/chatbot/chatbot-api.ts
```diff
@@ -0,0 +1,87 @@
...
+  updateWordProtection(data: IWordProtectionResponse): Promise<any | void>;
+
+  // delete
+  deleteCustomCommand(id: string): Promise<any | void>;
+  deleteTimer(id: string): Promise<any | void>;
+}
+
+export interface IChatbotCommonApi {
+  state: IChatbotCommonServiceState;
+  windowsService: any;
```
andycreeth: Generally an injected service wouldn't be considered part of the public API of anything.  And class that wants access to the window service should inject it itself.  Injected services should be considered private.

app/services/chatbot/chatbot-api.ts
```diff
@@ -0,0 +1,87 @@
...
+} from './index';
+
+export interface IChatbotServerApi {
+  // state + auth
+  state: IChatbotApiServiceState;
+  logIn(): Promise<any | void>;
+  logOut(): void;
+
+  // get
+  fetchChatbotGlobalEnabled(): Promise<any | void>;
```
andycreeth: All of these `any | void` types are effectively meaningless.  In fact the large majority of this API interface doesn't provide any type checking value because all of these functions don't have their actual return types specified.

app/services/chatbot/chatbot.ts
```diff
@@ -116,20 +125,20 @@ export class ChatbotApiService extends PersistentStatefulService<IChatbotApiServ
...
 
-  api(method: string, endpoint: string, data: any) {
+  private api(method: string, endpoint: string, data: any) {
     const url = this.apiEndpoint(endpoint, true);
     const headers = authorizedHeaders(this.state.apiToken);
     let options: {
-      headers: any,
-      method: string,
-      body?: string
+      headers: any;
```
andycreeth: Type should be `Headers`

PR #771, opened 2018-08-23
Handful of UI fixes

mbiemiller:
- cursor pointer fix for Sources
- extra space fix for Scenes
- few padding updates
- moved path inputs into global stylesheet for now 
- right margin fix for button within input wrappers


mbiemiller:
@gettinToasty lol yeah for sure, good call.. going to kill the branch after this PR

PR #777, opened 2018-08-24
remove NameSource window path for creating sources

andycreeth:
This removes a long-standing piece of tech debt that caused us to have 2 separate codepaths for adding a source that led to lots of replicated logic.


app/components/windows/AddSource.vue.ts
```diff
@@ -44,7 +44,7 @@ export default class AddSource extends Vue {
     return { name: source.name, value: source.sourceId };
   });
 
-  selectedSourceId = this.sources[0].sourceId;
+  selectedSourceId = this.sources[0] ? this.sources[0].sourceId : null;
```
gettinToasty: does this mean `sources` is either an array or a single source? i think we should standardize our data shapes here to always be arrays if that's the case

andycreeth: PS this is the *one* thing I miss from CoffeeScript.  In CoffeeScript you can use the existential operator to skip null checks:
```CoffeeScript
selectedSourceId = this.sources[0]?.sourceId;
```

gettinToasty: there's a proposal for safe navigation in ES; i think it's going to be added somewhat soon

andycreeth: oh nice

PR #782, opened 2018-08-28
small ui fixes for onboarding steps

mbiemiller:


app/components/pages/Onboarding.vue
```diff
@@ -79,13 +79,17 @@
...
-  padding: 0 20px;
+  .margin-bottom(3);
+  padding: 0 16px;
 }
 
 .setup-later {
-  margin-top: 20px;
-  color: @light-3;
+  .margin-top(3);
+  color: @night-paragraph;
```
gettinToasty: will this be messed up in day mode?

mbiemiller: we dont have day mode onboarding

PR #783, opened 2018-08-28
cleanup loud sentry events, fix settings

andycreeth:
This should save us a few million sentry events, and also fix the weird flashing when switching categories in settings.

### DIFF
--- a/app/services/media-backup.ts
+++ b/app/services/media-backup.ts
@@ -81,15 +81,24 @@ export class MediaBackupService extends StatefulService<IMediaBackupState> {
       return null;
     }
 
-    const stats = await new Promise<fs.Stats>((resolve, reject) => {
-      fs.lstat(filePath, (err, stats) => {
-        if (err) {
-          reject(err);
-        } else {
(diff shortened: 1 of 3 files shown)


app/services/media-backup.ts
```diff
@@ -81,15 +81,24 @@ export class MediaBackupService extends StatefulService<IMediaBackupState> {
...
-      fs.lstat(filePath, (err, stats) => {
-        if (err) {
-          reject(err);
-        } else {
-          resolve(stats);
-        }
+    let stats: fs.Stats;
+
+    try {
+      stats = await new Promise<fs.Stats>((resolve, reject) => {
```
holiber: it looks like a good candidate to be in `FileManagerService`

andycreeth: Yeah, `FileManagerService` is a bit poorly named.  It's really meant primarily for small text files, since it keeps all files in memory in order to give simpler/better read/write semantics.  It might be worth extracting some of this stuff into a utility library though.

PR #789, opened 2018-08-29
Fix volmeter slider

holiber:
Fix for the bug - slider is not working after chat collapse/expand


app/components/shared/Slider.vue.ts
```diff
@@ -22,8 +24,12 @@ export default class SliderInput extends Vue {
   mounted() {
     // Hack to prevent transitions from messing up slider width
     setTimeout(() => {
-      if (this.$refs.slider) this.$refs.slider.refresh();
+      this.onResizeHandler();
     }, 500);
+
+    new ResizeSensor(this.$el, () => {
+      this.onResizeHandler();
+    });
```
gettinToasty: instead of making an anonymous function you can just do `ResizeSensor(this.$el, this.onResizeHandler);`

also is there a reason to use this library vs `document.addEventListener('resize', this.onResizeHandler)`?

holiber: the `resize` event actually is window-resize event. When chat panel is expanding/collapsing the window size is not changing.

holiber:
- I did a really quick review of the code of this library so I'm not sure about all details but the base idea of this library is to insert an big invisible HTML element "resize-sensor" into your HTML element and track the `scroll` event that caused by resizing. It also caps the amount of this events via `requestAnimationFrame`. So it purely event-based and doesn't use something like `setTimeout`

- I can check these places

- I tried to update this vue-slider first, it didn't help =(

holiber:
The description of this lib is not so big http://marcj.github.io/css-element-queries/

andycreeth:
I don't have a list, but could probably find it just by doing a search for `setInterval` in the codebase.  It does sound like this library is more efficient and better than our current polling system.

PR #792, opened 2018-08-30
Move TitleBar into ChildWindow

gettinToasty:
Lets child windows alsways be closable


holiber:
The transitions window doesn't have a title

gettinToasty:
@holiber thanks, good catch

PR #798, opened 2018-09-04
app: Automatically install the required VCRedist 2017 version

Xaymar-Streamlabs:
This is an attempt at automatically installing the VCRedist 2017 dependencies for the obs-studio-node module. It is required that this runs before any other code, as other modules might keep a lock on files and require a system restart.

There is currently no UI showing the progress. There is a callback that can be used to send data to a very simple child window (similar to our updater), but I have not implemented that myself yet.

### DIFF
--- a/main.js
+++ b/main.js
@@ -1,509 +1,556 @@
 'use strict';
 
-////////////////////////////////////////////////////////////////////////////////
-// Set Up Environment Variables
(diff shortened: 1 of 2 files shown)


Xaymar-Streamlabs:
Failed test is the streaming test, which likely was currently being used by another build.

PR #804, opened 2018-09-05
Fix child window and widgets

holiber:


app/components/windows/ChildWindow.vue.ts
```diff
@@ -15,11 +15,13 @@ export default class ChildWindow extends Vue {
...
   components: { name: string; isShown: boolean; }[] = [];
 
+  private refreshingTimeout = 0;
+
   mounted() {
-    this.onWindowUpdatedHandler();
-    this.windowsService.windowUpdated.subscribe(params => {
-      if (params.windowId !== 'child') return;
-      this.onWindowUpdatedHandler();
+    this.onWindowUpdatedHandler(this.options);
```
gettinToasty: is this the only place we call `onWindowUpdatedHandler`? otherwise we need to include options in our other calls

holiber: `onWindowUpdatedHandler` is private. So yes it can be called only from current file

PR #807, opened 2018-09-10
Connect to current platform

AnkhHeart:
- Connect to current login platform instead of always using Twitch


app/services/chatbot/chatbot.ts
```diff
@@ -168,12 +169,12 @@ export class ChatbotApiService extends PersistentStatefulService<IChatbotApiServ
     const url = this.apiEndpoint(endpoint, true);
     const headers = authorizedHeaders(this.state.apiToken);
     let options: {
-      headers: any,
-      method: string,
-      body?: string
+      headers: any;
```
andycreeth: Type should be `Headers`

PR #809, opened 2018-09-10
Add notification when entering a reconnecting state

gettinToasty:
Currently kind of bootstrapped a "singleton" functionality to this (if i didn't it would send notifs every second or so until reconnected) but we might want to consider expanding that formally into the notification service if we find another use-case for it.

### DIFF
--- a/app/services/streaming/streaming.ts
+++ b/app/services/streaming/streaming.ts
@@ -17,6 +17,7 @@ import { UsageStatisticsService } from 'services/usage-statistics';
 import { $t } from 'services/i18n';
 import { StreamInfoService } from 'services/stream-info';
 import { AnnouncementsService } from 'services/announcements';
+import { NotificationsService, ENotificationType } from 'services/notifications';
 
 enum EOBSOutputType {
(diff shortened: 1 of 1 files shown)


app/services/streaming/streaming.ts
```diff
@@ -231,6 +233,19 @@ export class StreamingService extends StatefulService<IStreamingServiceState>
...
+    const msg = $t('Stream has disconnected, attempting to reconnect.');
+    const existingReconnectNotif = this.notificationsService.getUnread()
+      .filter((notice: INotification) => notice.message === msg);
+    if (existingReconnectNotif.length !== 0) return;
+    this.notificationsService.push({
+      type: ENotificationType.WARNING,
+      lifeTime: -1,
+      showTime: true,
+      message: $t('Stream has disconnected, attempting to reconnect.')
+    });
```
andycreeth: Do you think it's worth automatically clearing this notification if the stream successfully reconnects?

gettinToasty: wouldn't be against that idea at all

PR #811, opened 2018-09-11
Fix and Add Tests for facemask plugin

glaba13:
- Fix  Issues with old tests
- Add new Tests:
    - Basic Properties tests
    - Tracking Test
    - Performance tests
    - Mask Tests (It can be automatically tested all the masks from the folder and verified that applied correctly)
- Add more data for verifying results

Note 1:  Before Merging this PR, the facemask repository PR#27 [Add Testing Support](https://github.com/stream-labs/facemask-plugin/pull/27) **need to be merged** in order to work properly
Note 2: for testing test data (image, video, masks) are needed

### DIFF
--- a/test/facemask/index.js
+++ b/test/facemask/index.js
@@ -18,7 +18,10 @@ var receivedTestData = {
   lastPixelR : 0,
   lastPixelG : 0,
   lastPixelB : 0,
-  lastPixelA : 0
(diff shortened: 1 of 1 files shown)


glaba13:
facemask PR is merged already, so condition is cleared

PR #815, opened 2018-09-14
Add TimerInput

gettinToasty:


app/components/shared/inputs/TimerInput.vue.ts
```diff
@@ -0,0 +1,132 @@
+import { Component, Prop } from 'vue-property-decorator';
+import { BaseInput } from './BaseInput';
+import { ITimerMetadata } from './index';
+
+let holdTimeout: number = null;
+let holdInterval: number = null;
```
holiber: I'm a little nervous about these 2 variables with changeable values will be shared between all component's instances. Did you declare it here for any special case?

app/components/shared/inputs/TimerInput.vue.ts
```diff
@@ -0,0 +1,132 @@
...
+    const minute = val * 60;
+    this.updateValue(currentHrsInSecs + minute);
+  }
+
+  setSecond(val: number) {
+    const currentMinsInSecs = Math.floor((this.value % 3600) / 60) * 60;
+    this.updateValue(currentMinsInSecs + (val * 1));
+  }
+
+  updateValue(value: number) {
```
holiber: use `emitInput` from `BaseInput`. I think you took this method somewhere from `SliderInput`, I'm going to fix that in SliderInput too.

app/components/shared/inputs/TimerInput.vue
```diff
@@ -0,0 +1,212 @@
...
+
+      <div class="timer-footer">
+        <div v-if="hasHours">
+          Hr
+        </div>
+        <div>
+          Min
+        </div>
+        <div v-if="hasSeconds">
+          Sec
```
holiber: $t

holiber:
image
From the first sight arrows position should be a bit fixed. I feel if arrows will be inside the padding box that should looks beter. If it won't it's worth to ask @mbiemiller to fix

app/components/shared/inputs/TimerInput.vue.ts
gettinToasty: that was copied directly from the website's code

app/components/shared/inputs/TimerInput.vue.ts
gettinToasty: again, i think this was copied over from the site code

gettinToasty:
@holiber i tried putting the arrows inside the bounds of the padding but because of their size they looked quite cramped. for ease of click-ability and viewing this was the best solution i could come up with, but i suppose i could at least move them more inwards to the left

holiber:
I didn't mean the clickable area should be changed, it's ok. Only the place of drawing arrows looks a bit weird for my eye. Like it's somewhere from another set of UI components. Maybe the color of this input that is not the same as other components is contributing to this feeling too. However, it's my personal feeling so it's better to ask a third person.

mbiemiller:
hey I think the arrows should be 1 or 2px smaller in size and moved inward a bit - should be 8px from the right, and ~1-2px inward from top and bottom

app/components/shared/inputs/index.ts
```diff
@@ -34,6 +34,10 @@ export interface INumberMetadata extends IInputMetadata {
   placeholder?: string;
 }
 
+export interface ITimerMetadata extends INumberMetadata {
+  format?: 'hms' | 'hm' | 'ms';
+}
+
```
holiber: Add a new `metadata` helper for this type on line 74

PR #816, opened 2018-09-14
fix startup from protocol links

andycreeth:
Was a current working directory issue.  Should have been using the electron API for this anyway.


app/services/i18n/i18n.ts
```diff
@@ -165,7 +165,7 @@ export class I18nService extends PersistentStatefulService<II18nState> implement
   }
 
   private getI18nPath() {
-    return this.fileManagerService.resolve('app/i18n');
+    return path.join(electron.remote.app.getAppPath(),'app/i18n');
```
holiber: Does it really work for both packaged and unpackaged versions of the app?

andycreeth: Yup.  `getAppPath` is smart and returns different values depending on whether the app is running in an ASAR or not.  It's intended for this exact purpose.

PR #817, opened 2018-09-14
Export scenes to overlays

holiber:
- Export scenes to overlays
- rid of `getSceneByName` method


app/services/scene-collections/nodes/overlays/scenes.ts
```diff
@@ -26,14 +28,24 @@ export class ScenesNode extends ArrayNode<ISchema, IContext, Scene> {
 
     return {
       name: scene.name,
+      sceneId: scene.getSource().sourceId,
```
gettinToasty: is `sourceId` the same as `sceneId`? it seems strange to me that we would need to call `getSource` to grab the sceneId of a scene instance

PR #828, opened 2018-09-20
Redesign widget template

gettinToasty:


app/components/windows/WidgetEditor.vue
```diff
@@ -0,0 +1,367 @@
...
+      padding: 0;
+    }
+  }
+
+  .test-button .link {
+    color: @white;
+  }
+</style>
+
+<style lang="less" scoped>
```
holiber: What is the reason for having 2 style blocks here?

gettinToasty: one is unscoped

PR #833, opened 2018-09-21
Fix Windows title on ChildWindow renders

gettinToasty:
Because of the way ChildWindow loads, it is actually created and mounted before component data exists. Instead of setting the title in each component I just made it set in the update handler

### DIFF
--- a/app/components/windows/ChildWindow.vue.ts
+++ b/app/components/windows/ChildWindow.vue.ts
@@ -16,7 +16,7 @@ export default class ChildWindow extends Vue {
   @Inject() private windowsService: WindowsService;
   @Inject() private customizationService: CustomizationService;
 
-  components: { name: string; isShown: boolean; }[] = [];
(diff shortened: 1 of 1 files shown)


gettinToasty:
i'm not sure if this is an expensive operation but i can throw a guard clause in there if need be

app/components/windows/ChildWindow.vue.ts
```diff
@@ -75,6 +71,7 @@ export default class ChildWindow extends Vue {
     clearTimeout(this.refreshingTimeout);
     this.refreshingTimeout = window.setTimeout(() => {
       this.components.push({ name: options.componentName, isShown: true });
+      electron.remote.getCurrentWindow().setTitle(this.options.title);
```
holiber: Does it work for windows with `preservePrevWindow=true` like media gallery?

gettinToasty: good catch, will try to get it working on any shown window

PR #834, opened 2018-09-21
Merge stateful widgets

holiber:


app/components/shared/inputs/SliderInput.vue.ts
```diff
@@ -29,12 +30,14 @@ export default class SliderInput extends BaseInput<number, ISliderMetadata>  {
...
     setTimeout(() => {
       if (this.$refs.slider) this.$refs.slider.refresh();
+      this.isFullyMounted = true;
     }, 500);
   }
 
   @throttle(500)
   updateValue(value: number) {
-    this.$emit('input', this.roundNumber(value));
+    if (!this.isFullyMounted) return;
```
gettinToasty: so users cant interact with is the first 500ms after it loads?

app/components/windows/WidgetEditor.vue.ts
```diff
@@ -29,79 +33,81 @@ import { ToggleInput, NumberInput } from 'components/shared/inputs/inputs';
...
   }
 
   get loadingFailed() {
-    return this.requestState === 'fail' && !this.loaded;
+    return !this.loaded && this.settingsState.loadingState == 'fail';
+  }
+
+  get wData(): IWidgetData {
+    if (!this.settingsState.data) return null;
+    return cloneDeep(this.settingsState.data) as IWidgetData;
```
gettinToasty: if wData is just a clone here will it still update via the API when we enable custom code?

app/components/shared/inputs/SliderInput.vue.ts
holiber: Yes, the reason I added this condition was that VueSlider triggered an `input` event after the refresh call in `mounted` hook above. Didn't have time to investigate why it is going. However, I can't imagine so crazy fast user that can switch tabs and slider for less than 500ms

app/components/windows/WidgetEditor.vue.ts
holiber: Yes, since now all changes for widget settings are synced via websockets

app/components/shared/inputs/SliderInput.vue.ts
gettinToasty: it seems a little iffy to me since its a decently long timeout but i guess it'll work for now until we find out exactly what's going on

app/components/widgets/StreamBoss.vue.ts
```diff
@@ -22,6 +22,8 @@ import TestButtons from './TestButtons.vue';
 })
 export default class StreamBoss extends WidgetSettings<IStreamBossData, StreamBossService> {
 
+  tab = 'goal';
```
gettinToasty: what's this doing here? i don't see it map to anything in the template

holiber: This is a default tab that is using in `WidgetWindow.vue` . WidgetWindow is deprecated now but WidgetEditor uses `source` tab as default which actually shouldn't be a default one since not so many users really need it. We must decide how we determine a default tab. It looks like it can just be the first item in `navItems` but it worth to check. If it will work we can delete `tab` here.

app/components/widgets/WidgetSettings.vue.ts
```diff
@@ -144,4 +97,8 @@ export default class WidgetSettings<TData extends IWidgetData, TService extends
       }
     );
   }
+
+  protected afterFetch() {
+    // override me if you need
+  };
```
gettinToasty: what is the use-case for this?

app/components/widgets/goal/GenericGoal.vue.ts
```diff
@@ -32,6 +22,8 @@ interface IGoalCreateOptions {
 })
 export default class GenericGoal extends WidgetSettings<IGoalData, GenericGoalService> {
 
+  tab = 'goal';
```
gettinToasty: like StreamBoss, wondering why this is here

app/components/widgets/WidgetSettings.vue.ts
holiber: I found you used this method in TipJar

app/components/windows/WidgetEditor.vue
```diff
@@ -61,18 +61,22 @@
...
+              :hideConent="true"
+              className="widget-editor__top-tabs"
+              :tabs="codeTabs"
+              v-model="currentCodeTab"
+              @input="value => updateCodeTab(value)"
+            />
+            <code-editor v-if="currentCodeTab === 'HTML'" key="html" :value="wData" :metadata="{ type: 'html' }"/>
+            <code-editor v-if="currentCodeTab === 'CSS'"  key="css" :value="wData" :metadata="{ type: 'css' }"/>
+            <code-editor v-if="currentCodeTab === 'JS'" key="js" :value="wData" :metadata="{ type: 'js' }"/>
+            <custom-fields-editor  v-if="currentCodeTab === 'customFields'" key="customFields" :value="wData"/>
```
gettinToasty: i do like handling this here better, but do all of our widgets support custom code?

holiber: Just checked and realized that `RecentEents` and `TipJar` doesn't. Good catch. I'll add a flag in `apiSettings` property for that

app/components/widgets/StreamBoss.vue.ts
gettinToasty: ah ok, yea defaulting to source was mostly just a placeholder but i think defaulting to the first `navItem` should be fine

app/services/service.ts
```diff
@@ -81,6 +85,11 @@ export abstract class Service {
     if (enforcer !== singletonEnforcer) throw 'Cannot construct singleton';
   }
 
+  /**
+   * calls only once per application life
+   */
+  protected beforeInit() {}
```
gettinToasty: couldn't any of the operations that would need to be here just be included in `init` at the top of that function?

holiber: Looks like its something that I forgot to delete after my experiments with `Vuex` mutations inheritance. Will remove

app/services/widgets/widget-source.ts
```diff
@@ -0,0 +1,82 @@
...
+    this.widgetSourceState = this.widgetsService.state.widgetSources[sourceId];
+    Utils.applyProxy(this, this.widgetSourceState);
+  }
+
+  getSource() {
+    return this.sourcesService.getSource(this.sourceId);
+  }
+
+  getSettingsService(): WidgetSettingsService<IWidgetData> {
+    return this.widgetsService.getWidgetSettingsService(this.type);
```
gettinToasty: where is type actually set here? or is it set by widgets that extend this?

app/components/windows/WidgetEditor.vue.ts
holiber: Just checked, that's fine here. The socket event calls a mutation so it will be reactive here

app/services/widgets/widget-source.ts
holiber: it is in `Utils.applyProxy(this, this.widgetSourceState);`
This function transforms `type`, `sourceId` and `previewSourceId` into getters for Vuex.store
So instances of `WidgetSource ` are always up to date with store. We use the same pattern for Sources and SceneItems

That works fine because of state item  is `IWidgetSource`  and `WidgetSource implements IWidgetSource` line

holiber:
Hmm, I think WidgetWindow in not working in this branch

holiber:
It still exists in staging so if you wish to publish a new widget now it's better to create a PR for staging

PR #839, opened 2018-09-24
New updater patches

computerquip-streamlabs:
This includes the bootstrap, some very basic uploading tools (not integrated into the release script itself), and some modifications to main.js and package.json to reflect the changes.

Initial testing seems alright. Updating of the updater itself is done separately and independently of these patches (as per the design). So far, the updater is sitting at 3.5mB uncompressed which is downloaded each update.

I'm expecting some bugs, probably some complicated edge case bugs (at least a couple). Once those get out of the way, I'll finish implementing rollback support, test again, and then we should be alright to move on.


computerquip-streamlabs:
Oof, there appears to be some CSS styling changes here that I didn't notice. That'll need to be fixed before this goes in for sure.

PR #840, opened 2018-09-24
Platform  load prod app

Nitsorn:
Can load and install production apps


app/components/windows/AddSource.vue.ts
```diff
@@ -114,6 +115,7 @@ export default class AddSource extends Vue {
             propertiesManagerSettings: this.sourceAddOptions.propertiesManagerSettings
           }
         );
+
```
holiber: Please don't commit a file if it has only formatting changes and you base task is not something like refactoring or reformatting.
Motivation:
I want to always be able to check what files have been affected by the commit in git history. Files changes like this give me disinformation.

app/services/hosts.ts
```diff
@@ -48,7 +48,7 @@ export class HostsService extends Service {
   }
 
   get platform() {
-    return 'platform.streamlabs.com';
+    return 'dev-platform.streamlabs.com';
```
andycreeth: We currently release our dev builds from this branch, so I can't merge this until this change is backed out.  I'm not sure what the deploy status is on the APIs this uses, but if they're not out yet we should make sure this doesn't break.

app/services/platform-apps/index.ts
```diff
@@ -73,11 +73,26 @@ interface IAppManifest {
   authorizationUrls: string[];
 }
 
+interface IProductionApp {
```
andycreeth: I wonder if we should call this interface `IProductionAppResponse` or something, just to make it obvious this is a raw server response.

app/services/platform-apps/index.ts
```diff
@@ -88,6 +103,16 @@ interface IPlatformAppServiceState {
   loadedApps: ILoadedApp[];
 }
 
+interface ISubscription {
```
andycreeth: Same for this one.  Either `ISubscriptionResponse` or something.

app/services/platform-apps/index.ts
```diff
@@ -73,11 +73,26 @@ interface IAppManifest {
...
+
 export interface ILoadedApp {
   id: string;
   manifest: IAppManifest;
   unpacked: boolean;
   appPath?: string; // The path on disk to the app if unpacked
+  appUrl?: string; // the cdn url to the unzipped build
   appToken: string;
   devPort?: number; // The port the dev server is running on if unpacked
   poppedOutSlots: EAppPageSlot[];
```
andycreeth: I'm thinking now would be a good opportunity to split this interface to get more robust type checking.  It would look like this:

```TypeScript
// Attributes required by all loaded apps
interface IBaseLoadedApp {
  id: string;
  manifest: IAppManifest;
  unpacked: boolean;
  appToken: string;
  poppedOutSlots: EAppPageSlot[];
}

interface IUnpackedLoadedApp extends IBaseLoadedApp {
  unpacked: true;
  appPath: string;
  devPort: number;
}

interface IProductionLoadedApp extends IBaseLoadedApp {
  unpacked: false;
  appUrl: string;
}
```

This pattern is really nice because it makes it impossible to load an unpacked app but provide an appUrl instead of an appPath for example.

app/services/platform-apps/index.ts
```diff
@@ -177,10 +236,18 @@ export class PlatformAppsService extends
...
+
+  addApp(app: ILoadedApp) {
+    const { id, appPath, appToken } = app;
+    this.ADD_APP(app);
+    if (app.unpacked) {
+      // store app in local storage
+      localStorage.setItem(this.localStorageKey, JSON.stringify({
+        appPath, appToken
+      }));
+      this.appLoad.next(this.getApp(id));
```
andycreeth: I think the `appLoad` event is relevant for all types of apps, unpacked and production alike.  Is there a reason you're only sending it for unpacked apps here?

app/services/hosts.ts
Nitsorn: good call, I was about to ask Runxi to deploy to platform first before merging this. On hold for now.

app/services/platform-apps/index.ts
Nitsorn: 👍 like this. Will keep in mind to do similar pattern @andycreeth

app/services/platform-apps/index.ts
Nitsorn: I was under the impression that we were only using appLoad to keep track of unpacked apps load/unload event, but that make sense to do it for all. thanks

PR #841, opened 2018-09-24
Fix tab nav animation

gettinToasty:
Temporarily remove rendering of OBS display while in an animating state to improve smoothness of UI

editor_animation

### DIFF
--- a/app/components/windows/WidgetEditor.vue
+++ b/app/components/windows/WidgetEditor.vue
@@ -28,11 +28,11 @@
         <div class="custom-code__alert" :class="{ active: customCodeIsEnabled }" />
       </div>
 
-      <div class="content-container" ref="content">
+      <div class="content-container" :class="{ vertical: currentTopTab === 'code' }">
         <div class="display">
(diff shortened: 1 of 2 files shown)


holiber:
The animation that changes dimensions like `width` and `height` is very heavy for a CPU since it forces browser to recalculate positions off all nested elements. I think developers of chrome browser calls it `reflow`. To have a smooth animation even on a slow PC you should avoid `reflow` calls. For example, you can use css  `translate`, `translate3d` and `scale` that should not affect reflow. Also, you should use `position: absolute` or `position: fixed` for animated elements.

holiber:
Some detailed information about `reflow` https://developers.google.com/speed/docs/insights/browser-reflow

gettinToasty:
@holiber so i messed around with using absolute positions, translates, and scaling. For most of the animation it works pretty alright but there's a pretty big side-effect: the preview window gets scaled as well resulting in a bad UI

scaling

holiber:
Can you push your changes?

PR #850, opened 2018-09-26
Redesign custom code buttons

gettinToasty:
There are still minor edits to be made here such as a placeholder icon being used for Save and eventually autosaving custom code but that will be in a follow-up PR


holiber:
image

need padding here

holiber:
image
This is not enabling or disabling. This switcher completely deletes all my added fields and can make me very angry. Should be a delete-button here.

holiber:
I have a popup hint everywhere in this window
image

PR #855, opened 2018-09-27
Update electron to 3.0.1

computerquip-streamlabs:


computerquip-streamlabs:
The blocker for this were the tests. It's hard to say how hard we're getting hit with electron issues at the moment with all the reporting being in disarray but it might still be viable to look into getting this running in the near future. I'll fix this up when we want to look into trying it out. As far as I know, all the major Electron blockers are out of the way at this time.

PR #857, opened 2018-09-27
Update obs-studio-node to v0.3.21

Xaymar-Streamlabs:
This includes the following changes:
* connectOrHost behaves correctly again, switch to host if you only need the host behavior.
* QSV is included again.
* Facemask is included.
* OBS has been updated to 22.0.3.

To keep performance acceptable, please merge #837 before merging this one.


andycreeth:
I'm not sure what happened here, but the hash that got written into the `yarn.lock` does not match the archive that currently exists in github.  Perhaps this version was re-uploaded.  I am going to fix the `yarn.lock` because this doesn't install as-is.

PR #860, opened 2018-09-28
FrankerFaceZ Settings Support

SirStendec:
This pull request implements a new window for configuring FrankerFaceZ. The window can be opened with a new button in settings, or via FrankerFaceZ's link in the Twitch chat settings menu.

image
image

### DIFF
--- a/app/components/AppearanceSettings.vue.ts
+++ b/app/components/AppearanceSettings.vue.ts
@@ -16,7 +16,7 @@ export default class AppearanceSettings extends Vue {
   @Inject() private windowsService: WindowsService;
 
   settingsFormData: TObsFormData = null;
-  enableFFZEmotes: Boolean = false;
+  enableFFZEmotes = false;
 
 
   created() {
--- a/app/components/AppearanceSettings.vue.ts
+++ b/app/components/AppearanceSettings.vue.ts
@@ -2,8 +2,10 @@ import Vue from 'vue';
 import { Component } from 'vue-property-decorator';
 import { Inject } from '../util/injector';
 import GenericForm from 'components/obs/inputs/GenericForm.vue';
+import { $t } from 'services/i18n';
 import { TObsFormData } from 'components/obs/inputs/ObsInput';
 import { ICustomizationServiceApi, ICustomizationSettings } from 'services/customization';
+import { WindowsService } from 'services/windows';
 
 @Component({
   components: { GenericForm }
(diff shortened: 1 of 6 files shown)


app/components/AppearanceSettings.vue.ts
```diff
@@ -2,21 +2,26 @@ import Vue from 'vue';
...
 @Component({
   components: { GenericForm }
 })
 export default class AppearanceSettings extends Vue {
 
   @Inject() private customizationService: ICustomizationServiceApi;
+  @Inject() private windowsService: WindowsService;
 
   settingsFormData: TObsFormData = null;
+  enableFFZEmotes: Boolean = false;
```
andycreeth: The type you want here is `boolean` (lowercase "b").  You can read more about the difference here: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html

Also, since you're immediately assigning a boolean value, you can technically omit the type annotation entirely if you want and TypeScript will figure it out.

app/components/windows/FFZSettings.vue.ts
```diff
@@ -0,0 +1,43 @@
...
+})
+export default class FFZSettings extends Vue {
+  @Inject() customizationService: CustomizationService;
+  @Inject() windowsService: WindowsService;
+
+  $refs: {
+    ffzSettings: Electron.WebviewTag
+  }
+
+  nightMode: Boolean = false;
```
andycreeth: Should be `boolean` (see above).

app/components/windows/FFZSettings.vue.ts
```diff
@@ -0,0 +1,43 @@
...
+      webview.executeJavaScript(`
+        var ffzscript1 = document.createElement('script');
+        ffzscript1.setAttribute('src','https://cdn.frankerfacez.com/script/script.min.js');
+        document.head.appendChild(ffzscript1);
+      `, true);
+    })
+  }
+
+  get popoutURL() {
+    return `https://www.twitch.tv/popout/frankerfacez/chat?ffz-settings${this.nightMode ? '&darkpopout' : ''}`
```
andycreeth: So this is loading the `frankerfacez` popout chat, and then the injected javascript just replaces the content of the page with your settings?  Can this be any `twitch.tv` URL?  Just curious.

SirStendec: That can technically be any Twitch URL, yes. When FFZ loads, it checks for a `ffz-settings` query parameter and, if it's present, it displays its settings UI in exclusive mode. I use the chat pop-out URL specifically because it's a fairly lightweight page and because it initializes chat data, so certain parts of the FFZ settings can populate correctly.

As an example, the Badge Visibility settings are dependent on chat having been loaded to populate the list of available Twitch badges.

PR #869, opened 2018-10-02
get rid of source reference errors

andycreeth:


app/services/sources/properties-managers/properties-manager.ts
```diff
@@ -60,6 +66,7 @@ export abstract class PropertiesManager implements IPropertyManager {
    * properties manager.
    */
   destroy() {
+    this.destroyed = true;
```
holiber: this encapsulation doesn't look strong with the public instead of protected `destroyed `  property,

PR #871, opened 2018-10-02
Platform  app store events

Nitsorn:
- added platform app store
- added new protocol link for `paypalauth`
- added new service called `PlatformAppStoreService`, dealing with webview/app store communication from protocol link handler.
- webview injects guest-api


app/components/pages/PlatformAppStore.vue.ts
```diff
@@ -6,26 +6,49 @@ import { GuestApiService } from 'services/guest-api';
...
-      //   this.$refs.appStore.openDevTools();
-      // }
+    this.$refs.appStoreWebview.addEventListener('dom-ready', () => {
+      if (Utils.isDevMode()) {
+        this.$refs.appStoreWebview.openDevTools();
+      }
+      this.guestApiService.exposeApi(
+        this.$refs.appStoreWebview.getWebContents().id,
+        {
+          onInstallApp: this.onInstallAppHandler,
```
andycreeth: A function that starts with `on` generally implies that you are binding a callback for an event rather than performing an action.  I would expect these functions to be called something like:
`refreshApps` or `refreshAndLoadApps` maybe
`openLinkInBrowser`
`onPaypalSuccess` or smiliar.

Do those sound reasonable?

app/services/platform-app-store/index.ts
```diff
@@ -0,0 +1,30 @@
+import { StatefulService } from 'services/stateful-service';
+import { mutation } from 'services/stateful-service';
+
+interface IPlatformAppStoreServiceState {
+  paypalAuthCallback: Function;
+}
+
+export class PlatformAppStoreService extends
+  StatefulService<IPlatformAppStoreServiceState> {
```
andycreeth: You don't want to use a stateful service here.  Stateful services are only useful if you need the UI to react to changes in the store.  Furthermore, they have lots of drawbacks including boilerplate, and the fact that they are synced between our windows which means anything you put into a store needs to be serializable (functions are not).  You should use a normal service instead, and just store it as an attribute on the class instance.

app/services/protocol-links.ts
```diff
@@ -69,4 +71,12 @@ export class ProtocolLinksService extends Service {
       });
     }
   }
+
+  @protocolHandler('paypalauth')
+  private updateUserBillingInfo(info: IProtocolLinkInfo) {
+    // 1. navigate to store tab
+    // 2. trigger callback to store webview to continue checkout
+    this.navigationService.navigate('PlatformAppStore');
```
andycreeth: If the user has navigated to a different page, the webview will have been deloaded, and the binding on paypal success will have been lost.  This effectively makes this a failure state and I'm not sure if it would be possible for us to easily recover from that.

app/components/pages/PlatformAppStore.vue.ts
Nitsorn: yeah makes sense, dont know what I was thinking. Thanks Andy

PR #872, opened 2018-10-02
Import Script and Buildpath fix

Nitsorn:
can import scripts in production apps, and append build to filepath only when unpacked

### DIFF
--- a/app/components/PlatformAppWebview.vue
+++ b/app/components/PlatformAppWebview.vue
@@ -1,6 +1,6 @@
 <template>
 <webview
-  v-if="renderWebview && appUrl"
+  v-if="renderWebview"
   class="platform-app-webview"
   ref="appView"
   :src="appUrl"
--- a/app/components/PlatformAppWebview.vue
+++ b/app/components/PlatformAppWebview.vue
@@ -1,6 +1,6 @@
 <template>
 <webview
-  v-if="renderWebview"
+  v-if="renderWebview && appUrl"
   class="platform-app-webview"
   ref="appView"
   :src="appUrl"
--- a/app/components/PlatformAppWebview.vue.ts
+++ b/app/components/PlatformAppWebview.vue.ts
@@ -42,6 +42,8 @@ export default class PlatformAppWebview extends Vue {
   }
 
   attachWebviewListeners() {
+    if (!this.$refs.appView) return;
+
     this.$refs.appView.addEventListener('dom-ready', () => {
       if (this.platformAppsService.state.devMode) {
         this.$refs.appView.openDevTools();
@@ -67,6 +69,10 @@ export default class PlatformAppWebview extends Vue {
   }
 
   get appUrl() {
+    console.log('appUrl: ', this.platformAppsService.getPageUrlForSlot(
+      this.appId,
+      this.pageSlot
+    ));
     return this.platformAppsService.getPageUrlForSlot(
       this.appId,
       this.pageSlot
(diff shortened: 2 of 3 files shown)


app/components/PlatformAppWebview.vue.ts
```diff
@@ -67,13 +69,18 @@ export default class PlatformAppWebview extends Vue {
   }
 
   get appUrl() {
+    console.log('appUrl: ', this.platformAppsService.getPageUrlForSlot(
+      this.appId,
+      this.pageSlot
+    ));
```
andycreeth: can we remove

app/components/PlatformAppWebview.vue
```diff
@@ -1,6 +1,6 @@
 <template>
 <webview
-  v-if="renderWebview"
+  v-if="renderWebview && appUrl"
```
andycreeth: Is there any valid scenario where we wouldn't have a url here?

app/services/platform-apps/index.ts
```diff
@@ -16,6 +16,7 @@ import { HostsService } from 'services/hosts';
 import { handleErrors, authorizedHeaders } from 'util/requests';
 import { UserService } from 'services/user';
 import { trim, compact } from 'lodash';
+import { cd } from 'shelljs';
```
andycreeth: This looks like an automatic import lol

app/components/PlatformAppWebview.vue.ts
```diff
@@ -67,13 +69,18 @@ export default class PlatformAppWebview extends Vue {
...
     return this.platformAppsService.getPageUrlForSlot(
       this.appId,
       this.pageSlot
     );
   }
 
   get appPartition() {
+    console.log('App partition: ', this.platformAppsService.getAppPartition(this.appId));
```
andycreeth: Can be removed?

app/services/platform-apps/index.ts
```diff
@@ -134,6 +135,7 @@ export class PlatformAppsService extends
   appUnload = new Subject<string>();
 
   private localStorageKey = 'PlatformAppsUnpacked';
+  private baseCdnUrl = 'https://s3.amazonaws.com/streamlabs-platform';
```
andycreeth: We don't want to hard code this.  We want to add `appUrl` which you added last week as part of your production app stuff: https://github.com/stream-labs/streamlabs-obs/pull/840/files#diff-5ff6bf525e7ffbac00adf915bb8e26f2R99

Essentially here's the logic:
- Is this app running unpacked?  Do nothing past what we're normally doing
- Is this app running in production?  Make sure any script being loaded has the same beginning of the URL as `appUrl` we got back from the server.  That ensures the script belongs to this app.

app/services/platform-apps/index.ts
```diff
@@ -183,6 +185,7 @@ export class PlatformAppsService extends
  */
   async installProductionApps() {
     const productionApps = await this.fetchProductionApps();
+    console.log('production apps', productionApps);
```
andycreeth: remove

app/services/platform-apps/index.ts
```diff
@@ -468,20 +471,25 @@ export class PlatformAppsService extends
         }
 
         if (details.resourceType === 'script') {
-          console.log('got details', details);
-          console.log('GOT REQUEST', details.url);
+          // for cdn scripts, resourceType is other
```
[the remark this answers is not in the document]
Nitsorn: oh no, will remove thanks

app/services/platform-apps/index.ts
```diff
@@ -254,7 +254,7 @@ export class PlatformAppsService extends
     this.appLoad.next(this.getApp(id));
   }
 
-  async validateManifest(manifest: IAppManifest, appPath: string) {
+  async validateManifest(manifest: IAppManifest, appPath: string, isUnpacked = true) {
```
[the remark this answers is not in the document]
Nitsorn: gotcha, fixing

PR #882, opened 2018-10-05
project: Directly access OBS in Renderer processes

Xaymar-Streamlabs:
As we now have our own IPC framework, we no longer have to rely on electron's IPC framework to do our work for us. We can directly call things from any process, as long as that process is actually connected to the IPC server.

For OBS IPC, this means a lot of data can be directly accessed without blocking both the renderer and main process, thus freeing up resources for more responsiveness and faster rendering. As a side effect, we might see lower CPU usage too.

This PR replaces #735


app/app.ts
```diff
@@ -27,10 +26,18 @@ import ChildWindow from 'components/windows/ChildWindow.vue';
...
 
+window['obs'] = window['require']('obs-studio-node');
+
+{ // Set up things for IPC
+  // Connect to the IPC Server
+  window['obs'].IPC.connect(remote.process.env.SLOBS_IPC_PATH);
+  document.addEventListener('close', (e) => {
+    window['obs'].IPC.disconnect();
+  });
+}
```
andycreeth: Honestly I've never seen someone use a block like this in javascript outside the context of the normal uses such as functions and conditionals.  I'm curious what the motivation here is.  Are you just grouping these statements together for the purpose of readability?

Xaymar-Streamlabs:
Merged by #906.

PR #891, opened 2018-10-09
Platform  chat slot

Nitsorn:


app/services/windows.ts
```diff
@@ -170,7 +170,7 @@ export class WindowsService extends StatefulService<IWindowsState> {
       componentName: 'Main',
       scaleFactor: 1,
       isShown: true,
-      title: `Streamlabs OBS - Version: ${remote.process.env.SLOBS_VERSION}`
+      title: 'Streamlabs OBS - App Developer Build 8'
```
holiber: it doesn't look like something that should go to staging

PR #895, opened 2018-10-10
Temporarily remove reset default button

gettinToasty:
Our functionaliyty here isn't doing what the site does so this should be removed until it's fixed

### DIFF
--- a/app/components/widgets/CodeEditor.vue
+++ b/app/components/widgets/CodeEditor.vue
@@ -1,7 +1,7 @@
 <template>
   <div>
     <div class="toolbar">
-      <i class="icon-reset" v-tooltip="$t('Restore Defaults')" @click="restoreDefaults" />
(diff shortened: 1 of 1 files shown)


app/components/widgets/CodeEditor.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div>
     <div class="toolbar">
-      <i class="icon-reset" v-tooltip="$t('Restore Defaults')" @click="restoreDefaults" />
+      <!-- <i class="icon-reset" v-tooltip="$t('Restore Defaults')" @click="restoreDefaults" /> -->
```
holiber: How to restore defaults now?

gettinToasty: we never correctly restored defaults in the first place; this button has been broken since before the redesign

PR #901, opened 2018-10-11
Platform  settings in settings window

Nitsorn:
Load all production apps in `installed apps` tab in settings. Each app can be temporarily unloaded/ reloaded for developing/testing purposes.


app/services/platform-apps/index.ts
```diff
@@ -660,6 +686,18 @@ export class PlatformAppsService extends
...
+    const app = this.getApp(appId);
+    if (app.enabled) {
+      // disabling
+      this.appUnload.next(appId);
+    }
+    else {
+      this.appLoad.next(app);
+    }
+    this.TOGGLE_ENABLE(appId);
+  }
```
andycreeth: This feels a little brittle.  At the API level, implementing things as a toggle is generally a less robust solution because it requires knowing the current state to understand what the function will do.  Toggles are best implemented at the View layer, and at the API level we should state the intended state `setEnabled(true)` or `setEnabled(false)`.

app/services/platform-apps/index.ts
```diff
@@ -686,16 +692,14 @@ export class PlatformAppsService extends
     });
   }
 
-  toggleEnable(appId: string) {
+  setEnabled(appId: string, enabled: boolean) {
     const app = this.getApp(appId);
     if (app.enabled) {
-      // disabling
       this.appUnload.next(appId);
```
andycreeth: You have a bug here.  If I call `setEnabled('foo', true)` and that app is already enabled, we will erroneously send the `appUnload` event.

Nitsorn: fixing

PR #903, opened 2018-10-11
Update face masks when going live

kuhasuki:
This will reduce the window for potential buggy behavior when we release new masks.


kuhasuki:
Didn't think of possible bugs when the streamer isn't logged in, closing this for now.

PR #904, opened 2018-10-12
added catch handlers in platform-apps.initialize

Nitsorn:
added catch handlers to recover all methods called in `platform-apps.initiailize`

### DIFF
--- a/app/services/platform-apps/index.ts
+++ b/app/services/platform-apps/index.ts
@@ -188,6 +188,7 @@ export class PlatformAppsService extends
     return fetch(request)
       .then(handleErrors)
       .then(res => res.json())
+      .catch(() => []);
   }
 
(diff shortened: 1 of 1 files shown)


app/services/platform-apps/index.ts
```diff
@@ -446,7 +447,8 @@ export class PlatformAppsService extends
     return fetch(request)
       .then(handleErrors)
       .then(res => res.json())
-      .then(json => json.id_hash);
+      .then(json => json.id_hash)
+      .catch(() => null);
```
andycreeth: Do we need to handle this one in a special place somewhere else?  We should avoid loading an app with an app id of `null`.  That will cause a lot of issues for us down the line.  The app id is pretty important.

PR #905, opened 2018-10-12
Fix scene-collections sync

holiber:


holiber:
@andycreeth Just added an exeption for this case

app/services/scene-collections/scene-collections.ts
```diff
@@ -260,12 +265,20 @@ export class SceneCollectionsService extends Service
    * Instead, it will log an error and continue.
    */
   async safeSync(retries = 2) {
+
+    if (this.syncPending) {
+      throw new Error('Unable to start the scenes-collection sync process while prev process is not finished');
```
andycreeth: This makes me nervous because a function called "safe" can now throw an exception.  If our code is correct now, and we don't call this function twice, then it shouldn't be a problem.  But if we're wrong, this exception will cause bad things to start happening and make the situation potentially worse.

holiber: This exception just going to show that we have some problems in the code like a missed `await` construction. Do you wish the `safeSync` to schedule a next synchronization after the current one if it's busy?

PR #914, opened 2018-10-15
Platform  app nav

Nitsorn:
adding apps navbar below TopNav that is only visible when there are apps loaded
also fixed morgan's comments with SourcesShowcase columns: https://vulcun.slack.com/archives/CB99DQ42Y/p1539626112000100

### DIFF
--- a/app/components/AppsNav.vue.ts
+++ b/app/components/AppsNav.vue.ts
@@ -2,7 +2,7 @@ import Vue from 'vue';
 import { Component, Prop } from 'vue-property-decorator';
 import { Inject } from 'util/injector';
 import { NavigationService } from 'services/navigation';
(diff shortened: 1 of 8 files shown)


app/components/AppsNav.vue.ts
```diff
@@ -0,0 +1,114 @@
...
+  //     x = x.concat(this.platformAppsService.enabledApps.filter(app => {
+  //       return !!app.manifest.pages.find(page => {
+  //         return page.slot === EAppPageSlot.TopNav;
+  //       });
+  //     }));
+  //     i--;
+  //   }
+
+  //   return x;
+  // }
```
andycreeth: Curious what this is

PR #922, opened 2018-10-16
Platform  prod apps bug

Nitsorn:


app/services/platform-apps/index.ts
```diff
@@ -385,22 +386,19 @@ export class PlatformAppsService extends
...
-    }
+    this.state.loadedApps.forEach(app => this.unloadApp(app));
   }
 
-  unloadApp(appId: string) {
-    this.REMOVE_APP(appId);
-    this.appUnload.next(appId);
+  unloadApp(app: ILoadedApp) {
+    this.REMOVE_APP(app.id);
+    this.appUnload.next(app.id);
```
holiber: It's better to trigger this event at the end of this method. When the app has been completely removed

andycreeth: +1

PR #926, opened 2018-10-17
removed tooltip for record button and made active state more obvious …

mbiemiller:
…to reduce confusion


gettinToasty:
i think your `yarn.lock` accidentally got deleted here

PR #929, opened 2018-10-17
Execute JSONRPC commands as part of autoconfig

holiber:


app/services/auto-config/brand-device.ts
```diff
@@ -102,25 +115,85 @@ export class BrandDeviceService extends StatefulService<IBrandDeviceState> {
...
       obs.NodeObs.OBS_service_resetAudioContext();
 
+
+      // process API additional commands, some sources can be setup here
+      if (deviceUrls.onboarding_cmds_url) {
+        const cmdsPath = `${tempDir}/onboarding_cmds.json`;
+        await downloadFile(deviceUrls.onboarding_cmds_url, cmdsPath);
+        const cmds =  JSON.parse(fs.readFileSync(cmdsPath, 'utf8'));
+        if (!newSceneCollectionCreated) await this.sceneCollectionsService.create({ name: deviceName });
+        for (const cmd of cmds) this.ipcServerService.exec(cmd);
```
andycreeth: What's our plan for making sure these APIs don't get changed?  In a world where we have tons of these living in the database, it would be pretty easy to accidentally break the API for 1 or 2 of them.  I don't really have any good ideas off the top of my head, but is this something you have thought about?

holiber: So this is something that now is used only for GameCast and only we call this method. The only right way now to keep confidence that it will not gonna be broken is writing tests.

I personally don't like that we need this weird method. I prefer to get rid of it and implement an ability to have additional sources types depending on devices in your PC. I'll try to figure out some better solution. For now, I just isolated these weird things in `BrandDeviceService`

PR #943, opened 2018-10-19
Disable top-nav clicking while loading.

holiber:
`startLoadingOperation();` and `endLoadingOperation();` work good, but only in the case when you have them in the same method.

Example with issue:
```ts
async loadA() {
 startLoadingOperation();
 await loadB()
// ..ups we already stoped loading in the loadB() method
// weird things can be happen here
 endLoadingApplication();
}

async loadB() {
  startLoadingOperation();
  await doSomeExtraThings();
  endLoadingApplication();
}
```

Solution:
```ts
@RunInLoadingMode()
async loadA() {
 await loadB()
// we asre still in loading mode here, yahoo!
}

@RunInLoadingMode()
async loadB() {
  await doSomeExtraThings();
}
```


app/services/app/app-decorators.ts
```diff
@@ -0,0 +1,15 @@
+import { getResource } from '../../util/injector';
+import { AppService } from './app';
+
+export function RunInLoadingMode(): any {
+  return function (target: any, methodName: string, descriptor: PropertyDescriptor) {
+    const originalMethod = descriptor.value;
+    descriptor.value = function (...args: any[]) {
+      const appService = getResource<AppService>('AppService');
```
andycreeth: This is cool.  I have wondered before the best way to use a service outside of a class.

app/services/app/app-decorators.ts
```diff
@@ -0,0 +1,15 @@
+import { getResource } from '../../util/injector';
```
andycreeth: Import path can just be `util/injector`

app/services/app/app.ts
```diff
@@ -26,6 +26,7 @@ import { OutageNotificationsService } from 'services/outage-notifications';
 import { CrashReporterService } from 'services/crash-reporter';
 import { PlatformAppsService } from 'services/platform-apps';
 import { AnnouncementsService } from 'services/announcements';
+import uuid from 'uuid';
```
andycreeth: We tend to use v4 uuids because they are purely random:
`import uuid from 'uuid/v4';`

app/services/app/app.ts
```diff
@@ -144,14 +146,43 @@ export class AppService extends StatefulService<IAppState> {
...
+    }
+
+    const result = fn();
+    let returningValue = result;
+    if (result instanceof Promise) {
+      const promiseId = uuid();
+      this.loadingPromises[promiseId] = result;
+      returningValue = await result;
+      delete this.loadingPromises[promiseId];
+    }
```
andycreeth: We should add some error handling here.  Currently, any exceptions raised during the execution of `fn` (or if it returns a promise that rejects), we will get an infinite loading screen.  The scene collections service is extremely careful to never raise exceptions though its public API, but as this is meant to be more general purpose, we should handle errors in a better way than infinite loading.

holiber: good catch

app/services/app/app.ts
```diff
@@ -144,14 +146,50 @@ export class AppService extends StatefulService<IAppState> {
...
+  async runInLoadingMode(fn: () => Promise<any> | void) {
+
+    if (!this.state.loading) {
+      this.START_LOADING();
+      this.windowsService.closeChildWindow();
+      this.windowsService.closeAllOneOffs();
+      this.sceneCollectionsService.disableAutoSave();
+    }
+
+    const result = fn();
```
andycreeth: There needs to be a try/catch around this too, right?

PR #950, opened 2018-10-23
Widgets bug fixes

holiber:


app/components/shared/inputs/BaseInput.ts
```diff
@@ -16,16 +16,51 @@ export class BaseInput<TValueType, TMetadataType extends IInputMetadata> extends
...
-    if (this.$parent instanceof ValidatedForm) {
-      this.$parent.emitInput(eventData, event);
-    }
+
+    const needToSendEventToForm = (
+      this.form &&
+      !this.parentInput || (this.parentInput && !this.parentInput.delegateChildrenEvents)
+    );
+
+    if (needToSendEventToForm) this.form.emitInput(eventData, event);
```
andycreeth: I'm not entirely sure what the intention is here, but this whole things feels incredibly hacky.

holiber: We have a ValidationForm component with an `input` event that should be called on each `input` event of children inputs

```html
<validated-form @input="onFormInputHandler">
  <TextInput/>
  <ListInput/>
  <FontInput/>
</validated-form>
```

All nested inputs send an input-event to the ValidatedForm
But FontInput already has a ListInput child. So to prevent emitting 2 input-events from one component the rule here is to not send an event to the form component if it is not a root-input component.

But there is one exception when we use FormGroup as a template for a bunch of components. Since FormGroup is also an Input component. 

```

<validated-form>
  <v-form-group>
     <BoolInput/>
     <SliderInput/>
  </v-form-group>
</validated-form>

```

BWT the suggestions about how to implement this better are appreciated

PR #959, opened 2018-10-24
Integrate crash-handler

eddyStreamlabs:


app/services/app/app.ts
```diff
@@ -125,6 +130,7 @@ export class AppService extends StatefulService<IAppState> {
   @track('app_close')
   private shutdownHandler() {
     this.START_LOADING();
+    obs.NodeObs.StopCrashHandler();
```
andycreeth: I'm a little confused why we need to call this when we didn't call a method to start the crash handler in node-obs anywhere.

eddyStreamlabs: We spawn the crash handler in main.js. The reason I'm doing it here is because I don't want the crash-handler to catch any crashes that are happening on shutdown (quiet a few from what I have seen in both the renderer and OBS). So I'm making sure the OBS server, unregister and terminates it safely (this is what this function does).

PR #960, opened 2018-10-25
Fix chatbot issues

holiber:


holiber:
Looks like a GitHub diff bug

PR #963, opened 2018-10-25
Add Alert Box

gettinToasty:
On hold until some testing can get done


app/components/widgets/AlertBox.vue
```diff
@@ -0,0 +1,298 @@
...
+        <div class="variation-tile__image-box">
+          <img v-if="variation.settings.image.href" :src="variation.settings.image.href" />
+          <div class="variation-tile__name">
+            <input
+              type="text"
+              :value="variation.name"
+              :disabled="editingName !== variation.id"
+              :ref="`${variation.id}-name-input`"
+              @input="nameInputHandler($event.target.value)"
+              @blur="nameBlurHandler(variation.id)"
```
gettinToasty: This input is a deliberate antipattern to what we've been doing regarding widget inputs since there is some added functionality here that wasn't possible using our currently system, like programatically focusing the input component. Currently I have this input save on blur but I'm not sure if that should be the way to do it or a debounced on input. My worry with a debounce is they could select another input quickly and it may be overwritten since they all use the same handler.

app/components/widgets/AlertBox.vue.ts
```diff
@@ -0,0 +1,155 @@
...
+    this.selectedId = 'default';
+    this.wData.settings[this.selectedAlert].variations = this.wData.settings[this.selectedAlert].variations.filter(
+      (variation: IAlertBoxVariation) => variation.id !== id
+    );
+    this.save();
+  }
+
+  editName(id: string) {
+    this.editingName = id;
+    this.selectedId = id;
```
gettinToasty: This is done here with a stop propagation in the input to avoid possible race conditions which would lead to `this.selectedVariation` potentially being incorrect

app/components/windows/WidgetEditor.vue.ts
```diff
@@ -156,7 +162,20 @@ export default class WidgetEditor extends Vue {
...
+        const found = variations && variations.find(
+          (vari: IAlertBoxVariation) => this.selectedVariation.id === vari.id
+        );
+        if (found) {
+          found.settings.customHtmlEnabled = true;
+        }
+      });
+    }
+    console.log(newSettings);
+    this.widget.getSettingsService().saveSettings(newSettings);
```
gettinToasty: This is a pretty inelegant solution but i'm not sure of a better way to do it while not passing in like 100 props.

holiber:
Good job 👍

holiber:
Click `Add Alert`, try to switch `toggle input` : 
image

holiber:
I tried to toggle items on the left bar multiple times and got failed POST requests:
image

holiber:
Enable/DIsable custom code doesn't work for me

app/components/widgets/AlertBox.vue
```diff
@@ -0,0 +1,298 @@
...
+    <div v-if="selectedId !== 'default'">
+      <v-form-group :title="$t('Variation Condition')" type="list" v-model="selectedVariation.condition" :metadata="{ options: conditions }" />
+      <v-form-group :title="$t('Variation Frequency')" v-if="selectedVariation.condition === 'RANDOM'">
+        <frequency-input v-model="selectedVariation.conditionData" />
+      </v-form-group>
+      <v-form-group :title="$t('Variation Amount')" type="number" v-model="selectedVariation.conditionData" v-if="selectedVariation.condition !== 'RANDOM'" />
+    </div>
+  </validated-form>
+</widget-editor>
+</template>
```
holiber: Extracting metadata to the service probably the best way how we should implement forms.
Motivation 1:
As a third-party developer, I can create my own version of alert box settings. So I would prefer to fetch all metadata for `alert box` fields via an API request to the widget service.
Motivation 2:
Metadata can contain constraints like min and max values for number inputs. I can use it in tests to check that validations work. There are no ways to fetch this data from Vue components now in our tests

So can you please extract the metadata into service like here https://github.com/stream-labs/streamlabs-obs/pull/968/files#diff-161edac13f57f342c71dfd4a733621d0R52

I'm doing this for each widget that I'm writing tests for

gettinToasty:
@holiber for the first comment can you be more specific? which toggle were you trying to use? i can't reproduce with the information you've given

for the failed post, could you share the POST body that you sent? i'm guessing that for some reason your alert box got into a bad state unrelated to the toggles

holiber:
I feel like '"TypeError: Cannot read property 'settings' of undefined"' is related to failed POST request that I had before this error message

app/components/windows/WidgetEditor.vue.ts
holiber: It worth to move this method to `widget-settings` service and create an overridden version of it for `alertbox` service

holiber:
Not sure how it was in sketches but the left panel should be at least on the same level like `WidgetEditor` and `HTML CSS` tab, because the HTML/CSS tab depends on what is selected in the left panel and the main rule here is that we read UI from the left to the right and from the top to the bottom.
image

app/components/widgets/AlertBox.vue.ts
```diff
@@ -0,0 +1,166 @@
...
+    this.addAlertMenuOpen = !this.addAlertMenuOpen;
+  }
+
+  addAlert(type: string) {
+    const newVariation = this.alertBoxService.newVariation(type);
+    this.wData.settings[type].variations.push(newVariation);
+    this.selectedAlert = type;
+    this.addAlertMenuOpen = false;
+    this.save();
+    this.$nextTick(() => this.editName(newVariation.id));
```
holiber: why we need nextTick here? `editName()` already has it

gettinToasty: we push in a new variation here and have to wait till it renders on the dom before we can select said variation

gettinToasty:
@mbiemiller alex's screenshot above refers to the design change that wasn't in the given designs. do you have any thoughts on the above?

app/components/widgets/AlertBox.vue
gettinToasty: looks like the `formMetadata` helper was added in that branch so i won't be able to do that until that branch gets merged in

holiber:
Error in render: "TypeError: Cannot read property 'find' of undefined" when I click `RESUBS`

app/components/widgets/AlertBox.vue
holiber: `formMetadata` just adds a `name` property for each input. You can use a simple object instead for now. I can wrap it later

app/components/widgets/AlertBox.vue.ts
mbiemiller: @holiber thanks for pointing out that UX issue
I am going to think of a solution

andycreeth:
Can you switch base branch to `staging2`

holiber:
When I'm trying to change Font Size in HOSTS settings I have NaN
image

holiber:
Still have too many items in left menu:
image
How to reproduce:
- Open the alertbox settings
- Switch all items in left menu to `off` position
- Close alertbox settings
- Open alertbox settings

app/components/widgets/AlertBox.vue.ts
andycreeth: Is this worth leaving a comment in the code?

app/components/widgets/inputs/AlertLayoutInput.vue.ts
```diff
@@ -0,0 +1,33 @@
...
+import { Inject } from 'util/injector';
+import { IListMetadata } from '../../shared/inputs';
+import { BaseInput } from 'components/shared/inputs/BaseInput';
+import ImagePickerInput from '../../shared/inputs/ImagePickerInput.vue';
+import { CustomizationService } from 'services/customization';
+
+@Component({
+  components: { ImagePickerInput }
+})
+export default class ImageLayoutInput extends BaseInput<string, IListMetadata<string>> {
```
andycreeth: This class name doesn't match the file name.  This results in very confusing error messages from Vue when this component raises an exception.

app/services/widgets/settings/alert-box/alert-box-api.ts
```diff
@@ -0,0 +1,571 @@
...
+
+// SLOBS GENERAL SETTINGS
+export interface IAlertBoxSetting {
+  enabled: boolean;
+  showMessage: boolean;
+  showResubMessage?: boolean;
+  variations: IAlertBoxVariation[]
+}
+
+export interface IAlertBoxSettings extends IAlertBoxGeneralSettings {
```
andycreeth: This is a beastly file.  Props for defining an interface around this whole thing.

PR #966, opened 2018-10-29
Add Debounce for number inputs

holiber:


gettinToasty:
Won't this cause a pretty significant UI lag? Or is the UI updated before `updateValue` actually fires?

holiber:
@gettinToasty UI is updated before it fires updateValue

PR #968, opened 2018-10-29
Release tests for some widgets

holiber:
Tests for ChatBox, Goals Widgets and SteamBooss those works 100% time for me. Other tests are coming soon


app/styles/_colors.less
```diff
@@ -119,3 +121,5 @@
 @night-dropdown-border: @dark-4;
 @night-hover: #0F1F27;
 @night-shadow: rgba(1, 2, 2, 0.16);
+@night-editor-border: #274959;
+@night-editor-accent: #10222c;
```
gettinToasty: this is a regression

app/services/utils.ts
```diff
@@ -142,4 +142,5 @@ export default class Utils {
       });
     });
   }
+
```
gettinToasty: extra space

test/helpers/form-monkey.ts
```diff
@@ -0,0 +1,261 @@
...
+
+  async getInputs(formName: string): Promise<IUIInput[]> {
+    const formSelector = `form[name=${formName}]`;
+
+    if (!await this.client.isExisting(formSelector)) {
+      throw new Error(`form not found: ${formName}`);
+    }
+
+    const result = [];
+    const $inputs = await this.client.$$(`${formSelector} [data-role=input]`);
```
gettinToasty: is there a reason for prefixing `inputs` with `$` here? i get that it's necessary to interact with the api using the stringified format but it feels a little awkward when you do something like `${$inputs.length}`

holiber: $ at the beginning of the variable is just a codestyle that means that it's a reference to the input element in DOM and not an input value. Not a big deal here btw.

PR #973, opened 2018-10-30
Platform display embed API

andycreeth:


app/services/platform-apps/api/modules/display.ts
```diff
@@ -0,0 +1,118 @@
+import { Module, apiMethod, EApiPermissions, IApiContext } from './module';
+import { Display } from 'services/video';
+import uuid from 'uuid/v4';
+import electron from 'electron';
+import { Subscription } from 'rxjs/Subscription';
+
+interface IDisplayCreateOptions {
+  position: IVec2;
```
holiber: when I see `position` I always have a question, is it position of the left-top corner of the center of the screen?

app/services/platform-apps/api/modules/display.ts
```diff
@@ -0,0 +1,118 @@
...
+  size: IVec2;
+  sourceId?: string;
```
holiber: Why `sourceId` is optional?

app/services/platform-apps/api/modules/display.ts
andycreeth: Everything in SLOBS uses top-left as the origin currently.

app/services/platform-apps/api/modules/display.ts
andycreeth: If you omit `sourceId` the display will be the entire video output.

PR #979, opened 2018-10-30
add sources window

mbiemiller:
- added rule so app names cant wrap in add sources window
- made whole window wider

### DIFF
--- a/app/components/windows/SourcesShowcase.vue
+++ b/app/components/windows/SourcesShowcase.vue
@@ -221,21 +221,10 @@ h2 {
   display: flex;
   flex: 1 0 auto;
 
-  &.sources--has-platform-apps {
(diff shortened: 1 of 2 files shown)


app/components/windows/SourcesShowcase.vue
```diff
@@ -262,6 +251,11 @@ h2 {
   background-color: @day-section;
   width: 49%;
   .radius();
+  white-space: nowrap;
+  text-overflow: ellipsis;
+  max-width: 100%;
+  display: inline-block;
+  overflow: hidden;
```
gettinToasty: i thought width needed a pixel measurement here to work with ellipsis, is it ok like this?

andycreeth:
This fixes all my issues, thanks!

PR #989, opened 2018-11-02
Add the release of all fader objects for the backend

RodrigoHolztrattner:
This will cause the relase of all faders before calling the shutdown for the obs lib


RodrigoHolztrattner:
That is happening because the backend PR that implements this new method (OBS_Fader_ReleaseFaders()) needs to be merged first.

andycreeth:
Ah, makes sense

PR #993, opened 2018-11-05
Help page

mbiemiller:
- refactor of help page
- changed to thumbnails that link to YT and removed all webviews


app/components/pages/Help.vue.ts
```diff
@@ -5,6 +5,63 @@ import electron from 'electron';
 @Component({
 })
 export default class Help extends Vue {
+  videos = [
+    {
+      readMore: false,
+      name: 'Streamlabs OBS Quickstart',
```
gettinToasty: i think any user-visible copy should be wrapped in `$t()`

mbiemiller: ok - even though YT videos/titles are english only?

gettinToasty: oh hm 🤔 im not actually sure in that case lol

mbiemiller: yeah I don't know either, don't want to be misleading

PR #995, opened 2018-11-06
callout to warn about adding a source if an app has one

mbiemiller:


app/components/PlatformAppWebview.vue
```diff
@@ -11,7 +11,11 @@
     :partition="appPartition"
     preload="bundles/guest-api" />
   <div v-else class="platform-app-popped-out">
-    This app is currently popped out into an external window.
+    {{ $t("This app is currently popped out into an external window.") }}"
```
andycreeth: Is the quotation mark at the end of this line supposed to be there?  I don't see it matched with anything.

andycreeth:
I'm closing this for now, since I think if we are going to implement this, we need a callout that doesn't pop up and need to be closed every single time the app is opened.

PR #996, opened 2018-11-06
Use a non-async pipe instead of separate process

computerquip-streamlabs:
This changes the default node install on appveyor to x64
This should speed up any API-based tests

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,7 @@
 build: off
 
 install:
-  - ps: Install-Product node 8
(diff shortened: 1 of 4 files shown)


computerquip-streamlabs:
On average, it seems to cut down at least 3 minutes in the tests, maybe more. It's hard to measure completely but the time given by ava tests is objectively lower by at least a couple minutes.

computerquip-streamlabs:
Oh, just realized I used snake case out of habit. I'll modify that real quick.

PR #1000, opened 2018-11-08
Fix tipjar test

holiber:


holiber:
`--nosync` means do not download or upload scenes to the server, so each test starts with a clean cache and no need to delete any sources

PR #1005, opened 2018-11-09
Add facebook

gettinToasty:
If FB provides the proper API we can easily switch to the list input for game setting but until then text should be ok


app/components/windows/EditStreamInfo.vue.ts
```diff
@@ -81,16 +88,19 @@ export default class EditStreamInfo extends Vue {
 
   encoderProfile: IMultiSelectProfiles;
 
+  facebookPages: { pages: any[], page_id: string };
```
andycreeth: Ideally we should define an interface for the pieces of data we use in a page.  It looks like there already might be one defined in the facebook platform service?

app/components/windows/EditStreamInfo.vue.ts
```diff
@@ -107,6 +117,10 @@ export default class EditStreamInfo extends Vue {
         value: this.streamInfoService.state.channelInfo.game
       }
     ];
+    this.pageModel.value = this.facebookPages.page_id;
+    this.pageModel.options = this.facebookPages.pages.map((page: any) => (
```
andycreeth: Use the facebook page interface?

app/services/streaming/streaming.ts
```diff
@@ -226,7 +228,20 @@ export class StreamingService extends StatefulService<IStreamingServiceState>
    * its current state.
    */
   get formattedDurationInCurrentStreamingState() {
-    return this.formattedDurationSince(this.streamingStateChangeTime);
+    const formattedTime = this.formattedDurationSince(this.streamingStateChangeTime);
+    if (formattedTime === '03:50:00' && this.userService.platform.type ===  'facebook') {
+      const msg = $t('You are 10 minutes away from the 4 hour stream limit');
```
andycreeth: Ouch.  I forgot about this.  Glad we are warning users.

PR #1014, opened 2018-11-14
Fix app deloading

holiber:


app/services/app/app.ts
```diff
@@ -145,14 +145,14 @@ export class AppService extends StatefulService<IAppState> {
...
       this.performanceMonitorService.stop();
       this.transitionsService.shutdown();
       this.windowsService.closeAllOneOffs();
       await this.fileManagerService.flushAll();
       this.crashReporterService.endShutdown();
+      obs.NodeObs.OBS_Volmeter_ReleaseVolmeters();
+      obs.NodeObs.OBS_Fader_ReleaseFaders();
       obs.NodeObs.OBS_service_removeCallback();
       obs.NodeObs.OBS_API_destroyOBS_API();
       obs.IPC.disconnect();
```
andycreeth: Can we please move `this.crashReporterService.endShutdown()` to the bottom of this function right before `electron.ipcRenderer.send('shutdownComplete');`.  This is yet more stuff we are doing after telling the crash reporter we are done, which is keeping our shutdown crash rate artificially low.  Thanks!

PR #1016, opened 2018-11-14
feat(init): display error message & exit if OBS API fails to init

blackxored:
Display error messages if OBS API fails to initialize. Relies on `EVideoCodes` being returned from initialization function, which should be updated shortly. 

Pending:
- [x] Main window still displays. Ideally we wouldn't render anything but the dialog, correct?

CR:
- [x] Ensure we're not leaking memory
- [x] Is this the correct way to trigger a shutdown

EDIT: English 🙈

### DIFF
--- a/app/services/app/app.ts
+++ b/app/services/app/app.ts
@@ -1,10 +1,10 @@
 import uuid from 'uuid/v4';
-import { StatefulService, mutation } from 'services/stateful-service';
+import { mutation, StatefulService } from 'services/stateful-service';
 import { OnboardingService } from 'services/onboarding';
 import { HotkeysService } from 'services/hotkeys';
 import { UserService } from 'services/user';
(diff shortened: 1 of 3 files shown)


app/services/app/app.ts
```diff
@@ -86,7 +87,20 @@ export class AppService extends StatefulService<IAppState> {
...
+      'en-US',
+      electron.remote.process.env.SLOBS_IPC_USERDATA
+    );
+
+    if (apiResult !== EVideoCodes.Success) {
+      const message = apiInitErrorResultToMessage(apiResult);
+      showDialog(message);
+
+      crashHandler.unregisterProcess(this.pid);
+      electron.ipcRenderer.send('shutdownComplete');
```
andycreeth: Looking at the way the main process handles this IPC message, I think this should work fine.

PR #1019, opened 2018-11-15
feat(platform): add get source by ID to API

blackxored:
This adds a `getSource` method to the sources module, enabling getting a source by ID. 
Null-check curses are casted on the caller. We explicitly acknowledge that `getSource` in the service might return undefined, so we return `null` prior to serializing if that were to be the case.


app/services/platform-apps/api/modules/sources.ts
```diff
@@ -91,8 +98,8 @@ export class SourcesModule extends Module {
     const source = this.getAppSourceForApp(sourceId, ctx.app.id);
 
     source.setPropertiesManagerSettings({
-      appSettings: settings
-    });
+                                          appSettings: settings
```
computerquip-streamlabs: This probably isn't consistent with style guidelines.

blackxored: It isn't. Fixed, thanks for noting.

PR #1022, opened 2018-11-15
New badge

mbiemiller:
- 'New' badge for App Store in top nav
- hiding top nav text at smaller size to make nav less congested 
- added max width for help page

### DIFF
--- a/app/components/TopNav.vue
+++ b/app/components/TopNav.vue
@@ -13,7 +13,7 @@
       class="tab-button"
       :class="{ active: page === 'Dashboard' }"
       :disabled="!isUserLoggedIn || locked">
-      <i class="icon-dashboard"/> <span>{{ $t('Dashboard') }}</span>
+      <i class="icon-dashboard"/> <span class="tab-button__text">{{ $t('Dashboard') }}</span>
     </button>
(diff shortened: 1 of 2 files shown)


app/components/TopNav.vue
```diff
@@ -208,6 +217,24 @@
...
+  vertical-align: text-bottom;
+  border: 1px solid transparent;
+  line-height: 16px;
+  font-weight: 500;
+  display: inline-block;
+}
+
+.badge--new {
+  background: rgba(135, 54, 224, 0.08);
+  color: #8736e0;
```
gettinToasty: if these are new colors i think they should probably be defined in `_colors.less` and referenced as variables here

app/components/TopNav.vue
```diff
@@ -243,5 +270,10 @@
   .theme-toggle__icon--sun {
     display: none;
   }
+
+  .badge--new {
+    background: #8736e0;
+    color: #ffffff;
+  }
```
gettinToasty: same as above

mbiemiller: oops - we have these colors already
just pushed an update

PR #1031, opened 2018-11-16
Add messaging when there are no FB pages

gettinToasty:
Don't allow users to go live if they have no pages, and filter out pages that are not streaming enabled for this api. Need to hear back from FB about precisely what types of pages can be streamed with before updating this and moving forward

### DIFF
--- a/app/components/windows/EditStreamInfo.vue
+++ b/app/components/windows/EditStreamInfo.vue
@@ -3,7 +3,7 @@
   :show-controls="false"
   :customControls="true">
   <div slot="content">
-    <div v-if="infoLoading">
+    <div v-if="infoLoading || populatingModels">
       <i class="fa fa-spinner fa-pulse" />
     </div>
     <div v-if="infoError && !infoLoading" class="warning">
@@ -13,7 +13,7 @@
       <a @click="goLive">{{ $t('just go live.') }}</a>
       {{ $t('If this error persists, you can try logging out and back in.') }}
     </div>
-    <div v-if="!infoLoading && !infoError">
+    <div v-if="!infoLoading && !infoError && !populatingModels">
       <ObsTextInput v-model="streamTitleModel" />
       <ObsTextInput  v-if="isYoutube || isFacebook" v-model="streamDescriptionModel" />
       <ObsListInput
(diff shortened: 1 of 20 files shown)


app/components/windows/EditStreamInfo.vue.ts
```diff
@@ -279,6 +284,10 @@ export default class EditStreamInfo extends Vue {
     this.userService.postFacebookPage(model.value);
   }
 
+  openFBPageCreateLink() {
+    shell.openExternal('https://www.facebook.com/pages/creation/')
```
kuhasuki: Let's replace this with the link to make a gaming video creator page https://www.facebook.com/gaming/pages/create

app/components/windows/EditStreamInfo.vue
```diff
@@ -25,7 +25,16 @@
         :loading="searchingGames"
         @search-change="debouncedGameSearch"
         @input="onGameInput"/>
-      <ObsListInput v-if="isFacebook" :value="pageModel" @input="(pageId) => setFacebookPageId(pageId)" />
+      <div class="warning" v-if="isFacebook && !hasPages">
+        {{ $t('It looks like you don\'t have any Pages. Head to ') }}
+        <a @click="openFBPageCreateLink">{{ $t('Facebook Page Creation') }}</a>
+        {{ $t(' to create a page, and then trying again.') }}
```
andycreeth: Typo "trying" -> "try"

app/services/platforms/facebook.ts
```diff
@@ -79,61 +76,48 @@ export class FacebookService extends StatefulService<IFacebookServiceState> impl
...
     // FB Doesn't have token refresh, user must login again to update token
-    const host = this.hostsService.streamlabs;
-    const url = `https://${host}/api/v5/slobs/facebook/refresh`;
-    const headers = authorizedHeaders(this.userService.apiToken);
-    const request = new Request(url, { headers });
-    return fetch(request)
-      .then(handleErrors)
-      .then(response => response.json())
-      .then(response => this.userService.updatePlatformToken(response.access_token));
+    return new Promise((resolve) => (resolve()))
```
holiber: `return Promise.resolve();`

PR #1035, opened 2018-11-17
remove race conditions sending api ready

andycreeth:


kuhasuki:
Fixes the bug for face masks. Nice investigation.

PR #1037, opened 2018-11-18
Collect API usage analytics

holiber:


app/services/usage-statistics.ts
```diff
@@ -114,4 +129,30 @@ export class UsageStatisticsService extends Service {
...
+      event,
+      value,
+      product: 'SLOBS',
+      version: this.version,
+      count: 1
+    });
+    this.sendAnalytics();
+  }
+
+  @Debounce(2 * 60 * 1000)
```
andycreeth: Hey, so I missed this the first time I reviewed this, but I think you want a "throttle" here, not a "debounce".  The issue with using a debounce here is that it actually requires that this function isn't called for 2 minutes before executing.  As long as there is at least 1 event every 2 minutes, this function will never execute.  This isn't the behavior you want.  Instead you just want to rate limit this function, which is what throttle is for.

PR #1042, opened 2018-11-21
Twitch tags & after go live hook

blackxored:
I'm thinking I should probably split this PR. Submitting for comments regardless.

## `afterGoLive` hook 

Provide a hook that could be invoked by platform services,
after finishing going live. We're notified of live status
through signal handling, we chose to invoke such hook
directly if exists, as opposed to letting the client subscribe,
since there are many layers of indirection.

`startStreaming` and `toggleStreaming` now accept an optional
context, which is passed to the aforementioned `afterGoLive`
hook. This is necessary if the caller wants to use some data
that was passed prior to the stream starting (for example:
tags are passed through the `EditStreamInfo` component prior
to the stream being live, Twitch requires that the stream is
live in order to update tags, so we pass the tags before starting
and receive them on the hook).

## Twitch tags

* Fetch all Twitch tags (not quite, 100 since cursor pagination seems to be broken on their API).
* Update tags (requires re-auth, discussion required).

## Streams API

* Minimal support of the new Streams API, in support of tags.

Overall, this is very early. There's some UI for tags but it's pretty broke and by no means final.


app/components/windows/EditStreamInfo.vue.ts
```diff
@@ -107,6 +125,10 @@ export default class EditStreamInfo extends Vue {
       // If the stream info pre-fetch failed, we should try again now
       this.refreshStreamInfo();
     }
+
+    if (this.isTwitch) {
+      this.allTwitchTags = await this.twitchService.getAllTags();
```
gettinToasty: there's no need to inject the twitch service directly here, we can use the `getPlatformService` command to use it

app/services/platforms/twitch.ts
```diff
@@ -35,20 +49,26 @@ export class TwitchService extends Service implements IPlatformService {
...
     const headers = new Headers();
 
-    headers.append('Client-Id', this.clientId);
-    headers.append('Accept', 'application/vnd.twitchtv.v5+json');
-    headers.append('Content-Type', 'application/json');
-
-    if (authorized) headers.append('Authorization', `OAuth ${this.oauthToken}`);
+    Object.entries(this.getRawHeaders(authorized)).forEach(([key, value]) => {
+      headers.append(key, value);
+    });
```
gettinToasty: is there a reason you iterate over the raw headers here instead of just dropping them in the constructor?

app/services/platforms/twitch/pagination.ts
```diff
@@ -0,0 +1,3 @@
+export type TwitchPagination = {
+  cursor: string;
+};
```
gettinToasty: i'm not certain if this deserves it's own file, for interfaces if the main file gets crowded we typically split those out into a `{service}-api` file

app/services/platforms/twitch/streams.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export type StreamsResponse = {
+  data: Array<Stream>;
+  pagination: TwitchPagination;
+};
+
+export const getUserStreams = (userId: string, headers: TwitchRequestHeaders) =>
+  ajax
+    .getJSON<StreamsResponse>(`https://api.twitch.tv/helix/streams?user_id=${userId}`, headers)
+    .map(response => response.data);
```
gettinToasty: same as above, i feel like this function could be in the class file (if not a part of the class) and the interfaces could go in an api file

app/services/platforms/twitch/tags.ts
```diff
@@ -0,0 +1,157 @@
...
+  };
+
+  return ajax
+    .put(
+      `https://api.twitch.tv/helix/tags/streams?broadcaster_id=${streamId}`,
+      JSON.stringify(params),
+      headers,
+    )
+    .toPromise();
+};
```
gettinToasty: we traditionally use the `fetch` api to handle our HTTP requests, and although i'm a personal fan of ajax it seems like an antipattern here

app/components/shared/inputs/TwitchTagsInput.vue.ts
```diff
@@ -0,0 +1,34 @@
...
+    return prepareOptions(
+      this.i18nService.state.locale || this.i18nService.getFallbackLocale(),
+      this.tags,
+    );
+  }
+
+  onInput() {
+    this.$emit('input', this.value);
+  }
+}
```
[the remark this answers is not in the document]
blackxored: I believe `prepareOptions` and taking `tags` as prop make it kind of specific, but open to suggestions. Not sure if we'd be abstracting too early, specially since the UI state is not good at the moment.

app/components/windows/EditStreamInfo.vue.ts
blackxored: We need to typecast if we do it that way. Is this what we want?

app/services/platforms/twitch/pagination.ts
blackxored: We will do pagination code here, seemly unrelated to anything else but tags, once we receive feedback from Twitch on their API.

app/services/platforms/twitch/tags.ts
blackxored: The rest of this file uses Rx, and we need that since we're going to be doing recursive requests. This particular case is for consistency. We also have proper error handling (as opposed to `fetch`) when a status code is error or JSON parsing fails, for free.
We might also move the delay here if it turns out there's no way we can invoke `updateTags` reliably without one, we'll need Rx here too. Hopefully makes sense?

app/components/shared/inputs/TwitchTagsInput.vue.ts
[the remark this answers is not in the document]
gettinToasty: it could just be something like `TaggableInput`, i just bring it up because facebook and other platforms have tags we may want to implement as well

app/services/platforms/twitch/pagination.ts
gettinToasty: ohhh ok i didn't realize this was still WIP when commenting

app/services/platforms/twitch/streams.ts
blackxored: Would love to hear your thoughts on this, if we do add it to the same file then we're likely adding it after the class and then we'll have "use before define" warnings. Above the class would pollute it. 
As for the types in an API file, isn't that the same that having a separate file as is now (I couldn't find any, apologies if there's is one).

app/services/platforms/twitch.ts
blackxored: `TwitchHeaders` isn't compatible with `HeadersInit`, nor does it seem to like if we cast. We'd have to do something like
```javascript
return new Headers({ ...this.getRawHeaders(authorized) })
```
which might actually not be that bad. Up to you.

app/components/shared/inputs/TwitchTagsInput.vue.ts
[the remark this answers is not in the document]
blackxored: I think this is still too coupled to Twitch. `prepareOptions` is Twitch specific, and so are the types used in the component. In addition, the way that tags work on Twitch (e.g you can't add a new one, thus that specific placeholder) is engraved in here. So is the fact that it doesn't have search (for now), while the other platforms might. I don't think in it's current state this would be reusable at all. But I'm definitely happy to generalize once we're confident with it (at this moment I am not).

app/components/shared/inputs/TwitchTagsInput.vue.ts
```diff
@@ -0,0 +1,34 @@
+import Vue from 'vue';
```
blackxored: Just a test

app/components/shared/inputs/TwitchTagsInput.vue
```diff
@@ -0,0 +1,113 @@
+<template>
```
blackxored: Just testing

app/services/platforms/twitch/pagination.ts
blackxored: After taking a closer look at this, it doesn't seem like we're gonna be able to decouple pagination enough to not tie it to the specific resource, so I'm having second thoughts about having this file as well.

app/services/platforms/twitch/tags.ts
```diff
@@ -0,0 +1,151 @@
...
+
+const byName = contramap((tag: TwitchTagWithLabel) => tag.name, ordString);
+const sortByName = sortBy1(byName, []);
+
+export const prepareOptions = (
+  locale: string,
+  tags: TwitchTag[] | undefined,
+): TwitchTagWithLabel[] =>
+  fromNullable(tags)
+    .map(
```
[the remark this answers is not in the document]
blackxored: This code is most likely gonna have to change, please ignore for now. In any case, I believe there are readability benefits as having functions in their own line, similar to how Promises `then` call works, which is also composition (in a way).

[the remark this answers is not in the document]
andycreeth: I'm like 99% sure this error message is coming from ts-lint.

app/services/platforms/twitch/tags.ts
```diff
@@ -0,0 +1,151 @@
...
+    );
+
+/**
+ * Fetch all available tags that Twitch provides that are not
+ * automatically generated. This will use the provided pagination
+ * to request the whole dataset of tags.
+ *
+ * @param headers Headers including OAuth Token and App ID
+ */
+export const getAllTags = (headers: TwitchRequestHeaders): Promise<TwitchTag[]> =>
```
holiber: We use `function` declarations instead of `const`. In addition to being more consistent with the current codebase, it's a more flexible way in some cases. For example, when you declare the function with `function`  keyword you don't have to care about the order of declaration.

app/components/shared/inputs/TwitchTagsInput.vue.ts
```diff
@@ -0,0 +1,55 @@
...
+import { Component, Prop } from 'vue-property-decorator';
+import VSelectPage from 'v-selectpage';
+import { Inject } from 'util/injector';
+import { $t, I18nService } from 'services/i18n';
+import { prepareOptions, TwitchTag, TwitchTagWithLabel } from 'services/platforms/twitch/tags';
+
+Vue.use(VSelectPage);
+
+@Component({})
+export default class TwitchTagsInput extends Vue {
```
holiber: Components inside `shared/inputs` must be consistent with other components.  It must extend `BaseInput` and have all props beside `value` in the `metadata` property. For example, you can take a look on `TextInput`.

holiber: If you don't want your component to be consistent with other Inputs please rename it to something like `TwitchTagsEditor` and place outside the `shared/inputs` folder

app/components/shared/inputs/TwitchTagsInput.vue.ts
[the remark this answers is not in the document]
holiber: That is sad I can't reuse this component for other purposes but I have it in the `shared` folder. If it's a very specific component for `Twitch` can you split it into 2 components - `TagsInput` and `TwitchTagsInput`

app/components/windows/EditStreamInfo.vue.ts
```diff
@@ -13,21 +13,25 @@ import { StreamingService } from 'services/streaming';
...
 @Component({
   components: {
     ModalLayout,
     HFormGroup,
     BoolInput,
     ListInput,
+    TwitchTagsInput,
   },
 })
 export default class EditStreamInfo extends Vue {
```
holiber: I feel like this component needs tests now... I don't mean in this PR

app/services/platforms/twitch/tags.ts
blackxored: Hmm, do we have a strong argument against using arrow functions for this? I think not having to care about the order of declaration is actually a detriment and not a benefit, we also don't have `arguments` which is also a source of confusion sometimes. As for better stack traces, I don't believe that's true anymore, JS engines do support this case, you can check with `function.name`.

app/services/platforms/twitch.ts
```diff
@@ -35,14 +53,27 @@ export class TwitchService extends Service implements IPlatformService {
...
+      'Content-Type': 'application/json',
+    };
 
-    headers.append('Client-Id', this.clientId);
-    headers.append('Accept', 'application/vnd.twitchtv.v5+json');
-    headers.append('Content-Type', 'application/json');
+    return authorized
+      ? {
+          ...map,
+          Authorization: `${isNewApi ? 'Bearer' : 'OAuth'} ${this.oauthToken}`,
```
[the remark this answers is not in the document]
blackxored: Yeah, check this: https://dev.twitch.tv/docs/authentication/#sending-user-access-and-app-access-tokens

app/services/platforms/twitch/tags.ts
```diff
@@ -0,0 +1,151 @@
...
+  data: TwitchTag[];
+  pagination: TwitchPagination;
+};
+
+export interface TwitchRequestHeaders {
+  Accept: 'application/vnd.twitchtv.v5+json';
+  Authorization?: string;
+  'Client-Id': string;
+  'Content-Type': 'application/json';
+}
```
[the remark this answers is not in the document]
blackxored: Extracted out.

app/components/shared/inputs/TwitchTagsInput.vue.ts
blackxored: Work on this, including extracting/generalizing components is now on a separate branch. I don't think we should include on this one.

PR #1043, opened 2018-11-21
Quality of Life Improvements for Development

blackxored:
This PR is an initial step towards providing general improvements for developers and contributors of SLOBS. 

* Ensure consistency between settings (match Prettier config with TSLint rule for example).
* Gets ESLint working. 
* Adds [EditorConfig](https://editorconfig.org/) so editors can easily pickup our settings, and aid in cross-platform development (e.g. line feeds).
* Upgrades TypeScript to latest (~3.1.6~ 3.2.1). Updating selection API (which was apparently outdated) and pinning `@types/jquery` (used from our deps) were the only required changes.
* The more important parts are opt-in validations and automation. Details follow.
* Since `rimraf` was already a dependency, I don't believe there's a need to use `shx` anymore (we're only using it to `rm -rf` files). In addition, there's less typing and more clarify in our scripts section of our `package.json`. And yes, I've looked at #693 😂. Still unclear to me why `rimraf` is on `dependencies` instead of `devDependencies` BTW.
* Specify minimum node version, and prevent running on future versions. See commit message for details.

### Opt-in 

The following is opt-in behavior, meaning it's not enabled by default, as to not hinder the team and new contributors. However, I'd strongly recommend everyone to do opt in.

Here's how:

Create a file called `.opt-in` at the root of the repository, with the options you wish to enable (one per line).

- `pre-commit`:

  * Source code files are formatted with Prettier automatically.
  * Source code files are checked with TSLint and ESLint automatically.

- `commit-msg`:
  * Commit messages are validated to follow a convention.
  See [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0-beta.2/) and [commitlint-config-conventional](https://github.com/marionebl/commitlint/tree/master/@commitlint/config-conventional) for details. The benefits of a commit convention include, but not limited to, consistency in commit types (with the side-effect of shorted and more focused commits, e.g fixes vs features), automatic changelog generation, semantic versioning once we reach 1.0, etc).

Please note, if you do opt-in, that since these are new practices, not enforced before, there's a good chance you might want a workaround if something gets in your way coming from unrelated areas.
While we encourage you to fix as many as you can, you can always opt-out in an case by case basis by running `git commit -n`.

Documenting this in something like a CONTRIBUTING.md is perhaps something we'd like to do at some point in the future, for visibility.

PR #1046, opened 2018-11-23
Fix tslint/style issues

blackxored:
Fix all formatting issues across the codebase. TSLint config was (and remains for now), pretty much style-checks. However, lots of files do not adhere  to that, plus the couple of additions I've introduced lately. 
This sort of depends on #1043, hence might need rebasing so don't repeat commits.


app/components/page-components/Chatbot/shared/ChatbotAliases.vue.ts
```diff
@@ -1,7 +1,7 @@
 import { Component, Prop } from 'vue-property-decorator';
 import ChatbotBase from 'components/page-components/Chatbot/ChatbotBase.vue';
 
-import { ITextMetadata } from 'components/shared/inputs/index';
+import { ITextMetadata } from '../../../shared/inputs';
```
gettinToasty: i think for a lot of these we prefer the aliased `components/shared..` convention, it's a lot clearer to see where files are coming from that way

PR #1047, opened 2018-11-23
Fix "duplicated keys" error in Settings

holiber:
That was pretty annoying


holiber:
Unfortunately, these duplicated keys come from the OBS. We are going to refactor settings anyway, I just don't want to have a mess in errors logs for that

PR #1054, opened 2018-11-27
WIP: OBS as source of truth for hotkeys

blackxored:
Initial support for using OBS as the source of truth for hotkeys.


app/services/hotkeys.ts
```diff
@@ -519,31 +537,65 @@ export class Hotkey implements IHotkey {
...
+const isSupportedHotkey = (hotkey: OBSHotkey) =>
+  hotkey.ObjectType === obs.EHotkeyObjectType.Source && getActionFromName(hotkey.HotkeyName);
+
+const isSceneItem = (hotkey: OBSHotkey) =>
+  !!hotkey.HotkeyName.match(/^(SHOW|HIDE)_SCENE_ITEM.(.+)$/);
+
+const isSource = (hotkey: OBSHotkey) => !!hotkey.ObjectName.match(/(source|input|output)/);
+
+const idPropFor = (hotkey: OBSHotkey) =>
+  isSource(hotkey) ? 'sourceId' : isSceneItem(hotkey) ? 'sceneItemId' : 'sceneId';
```
andycreeth: IMO nested ternaries make the logic difficult to follow.  We don't have a strict style guide (yet) but started out loosely following AirBnB, which recommends against this: https://github.com/airbnb/javascript#comparison--nested-ternaries

blackxored: Ideally we would not need this at all. I think ternaries are a common `if` replacement (which we should hate more 😂) but I'm happy to remove them.

app/services/hotkeys.ts
```diff
@@ -240,166 +284,179 @@ export class HotkeysService extends StatefulService<IHotkeysServiceState> {
...
   private updateRegisteredHotkeys() {
     const hotkeys: IHotkey[] = [];
+    /*
+     * Since we're hybrid at this point, track already-added hotkeys so OBS
+     * hotkeys don't duplicate them
+     */
+    const addedHotkeys = new Set<string>();
 
-    HOTKEY_ACTIONS.GENERAL.forEach(action => {
+    Object.values(GENERAL_ACTIONS).forEach(action => {
```
gettinToasty: AFAIK this is the only place you're using the `*_ACTIONS` objects so i'm kind of wondering why you refactored them into objects? if you only use them to turn back into arrays it just seems like it's adding in an extra step, but i might be missing something here

app/services/hotkeys.ts
```diff
@@ -240,166 +284,179 @@ export class HotkeysService extends StatefulService<IHotkeysServiceState> {
...
 
-
   applyHotkeySet(hotkeySet: IHotkeysSet) {
     const hotkeys: IHotkey[] = [];
     hotkeys.push(...hotkeySet.general);
     Object.keys(hotkeySet.scenes).forEach(sceneId => hotkeys.push(...hotkeySet.scenes[sceneId]));
-    Object.keys(hotkeySet.sources).forEach(sourceId => hotkeys.push(...hotkeySet.sources[sourceId]));
+    Object.keys(hotkeySet.sources).forEach(sourceId =>
+      hotkeys.push(...hotkeySet.sources[sourceId]),
+    );
```
gettinToasty: are these all automatic linter changes? i feel like we should change the config because we use single line ifs and anonymous functions all the time

app/services/hotkeys.ts
```diff
@@ -504,46 +561,86 @@ export class Hotkey implements IHotkey {
...
+const getMigrationMapping = (actionName: string) => {
+  return {
+    MUTE: 'TOGGLE_MUTE',
+    UNMUTE: 'TOGGLE_UNMUTE',
+  }[normalizeActionName(actionName)];
+};
+
+const getActionFromName = (actionName: string) => {
+  return ACTIONS[actionName] || ACTIONS[getMigrationMapping(actionName)];
+};
```
gettinToasty: not sure why this and the above are using manual returns

app/services/hotkeys.ts
```diff
@@ -504,46 +561,86 @@ export class Hotkey implements IHotkey {
...
+  } else if (isScene(hotkey)) {
+    return 'sceneId';
+  } else if (isSceneItem(hotkey)) {
+    return 'sceneItemId';
+  } else {
+    return null;
+  }
+};
+
+const normalizeActionName = (actionName: string) => actionName.split('.')[0];
```
gettinToasty: i feel like utility functions should be defined above the class, we already have several examples of that in this file. open to putting them after but either way i think it should be consistent.

app/services/scenes/scenes.ts
```diff
@@ -1,31 +1,31 @@
...
-  Scene,
+  ISceneCreateOptions,
   ISceneItem,
-  SceneItem,
+  IScenesServiceApi,
   IScenesState,
-  ISceneCreateOptions,
-  IScenesServiceApi
+  Scene,
+  SceneItem,
```
gettinToasty: not sure why these imports change order? it just seems like dirtying the diff for no actual reason

app/services/scenes/scenes.ts
```diff
@@ -189,67 +184,57 @@ export class ScenesService extends StatefulService<IScenesState> implements ISce
...
       queryParams: { sceneToDuplicate: sceneName },
       size: {
         width: 400,
-        height: 250
-      }
+        height: 250,
+      },
     });
   }
 }
```
gettinToasty: it doesn't look like anything was actually changed in this file apart from style changes? i think we should keep those in separate PRs to reduce diff size and cognitive load when reviewing

app/services/hotkeys.ts
```diff
@@ -40,148 +53,177 @@ export interface IBinding {
...
+      streamingService.toggleStreaming();
     },
-    {
-      name: 'ENABLE_STUDIO_MODE',
-      description: () => $t('Enable Studio Mode'),
-      down: () => getTransitionsService().enableStudioMode(),
-      isActive: () => getTransitionsService().state.studioMode
+    isActive: () => {
+      const streamingService = getStreamingService();
+      return !streamingService.isStreaming;
```
holiber: Don't see a reason why you replaced one line expression to the 2 lines

app/services/hotkeys.ts
blackxored: We will peel them off, one by one. Easier to have them separate to differentiate logic (such as general being run first and not having to duplicate-check), does that make sense?

app/services/hotkeys.ts
```diff
@@ -204,33 +249,34 @@ export interface IHotkeysSet {
   scenes: Dictionary<IHotkey[]>;
 }
 
-
 interface IHotkeysServiceState {
   hotkeys: IHotkey[]; // only bound hotkeys are stored
 }
 
+type OBSHotkey = {
```
holiber: We usually use `T` prefix for types
```
type TOBSHotkey
```

app/services/hotkeys.ts
blackxored: For style issues, I believe we should focus our effort on #1129. I believe it'd be wasteful to change something back to just change it again (or not) in the other branch. This might've been before we merged #1043, so I apologize if it was a result of my editor and switching branches.

app/services/hotkeys.ts
gettinToasty: i understand breaking them up into chunks, but the data transformation from `array` into `object -> array` is where i'm lost

app/services/hotkeys.ts
blackxored: I agree 100% with this, however I've mixed feelings about it. There wasn't an easy choice, as I believe the most important thing for us looking at this file would be
Types -> Consts -> Class -> Utils
Unfortunately, the functions you see above Class are there because of no-use-before-define. The rest is supporting and I believe should stay out of the way at the end. Unless we really want them in our way before we get to read the class, that is.

app/services/hotkeys.ts
```diff
@@ -40,148 +53,179 @@ export interface IBinding {
...
+  up?(entityId: string, hotkeyId?: number): void;
 
   // These are injected dynamically
   downHandler?(): void;
+
   upHandler?(): void;
 }
 
+type HotkeyGroup = {
+  [x: string]: IHotkeyAction;
```
holiber: what is `x`? I assume it must be `actionName` here

app/services/scenes/scenes.ts
holiber: Changes of this file have to be excluded from this PR

app/services/hotkeys.ts
```diff
@@ -1,31 +1,44 @@
...
+const getStreamingService = (): StreamingService => StreamingService.instance;
 
-function getTransitionsService(): TransitionsService {
-  return TransitionsService.instance;
-}
+const getTransitionsService = (): TransitionsService => TransitionsService.instance;
 
-type THotkeyType = 'GENERAL' | 'SCENE' | 'SCENE_ITEM' | 'SOURCE';
+const isAudio = (sourceId: string) => {
+  const source = getSourcesService().getSource(sourceId);
```
holiber: Is it really possible to get  `source = null` in this function?

app/services/hotkeys.ts
blackxored: We need these indexed for `getActionFromName` and normalization, in the future, this could be a way to improve performance in lookups as well.

app/services/hotkeys.ts
blackxored: I've changed `getActionFromName`. I believe `getMigrationMapping` should stay this way due to a somewhat complex expression and the fact we're returning an object.

app/services/scenes/scenes.ts
blackxored: How do you mean? We need the scene dedup that's in here.

app/services/hotkeys.ts
blackxored: Yes, in my testing

app/services/hotkeys.ts
blackxored: This is conventional indexer syntax in TS, do you think having `actionName` would really increase readability?

app/services/hotkeys.ts
blackxored: Can you please paste the line? I'm not entirely sure what you're referring to in this particular case.

app/services/hotkeys.ts
blackxored: Only to disambiguate, no?

app/services/scenes/scenes.ts
```diff
@@ -189,9 +188,7 @@ export class ScenesService extends StatefulService<IScenesState> implements ISce
   }
 
   get scenes(): Scene[] {
-    return this.state.displayOrder.map(id => {
-      return this.getScene(id);
-    });
+    return uniqBy(this.state.displayOrder.map(id => this.getScene(id)), x => x.id);
```
holiber: That's looks like a bug symptom fix. Can you please describe this bug in the TODO comment before this line and add a task to Asana with steps how to reproduce this issue?

app/services/hotkeys.ts
holiber: yes

app/services/scenes/scenes.ts
```diff
@@ -1,4 +1,6 @@
 import Vue from 'vue';
+import { uniqBy } from 'lodash';
+import { Subject } from 'rxjs';
```
holiber: Unused import

blackxored: Which one exactly? Both `uniqBy` and `Subject` are used.

app/services/scenes/scenes.ts
blackxored: I've added it closer to where the bug happens.

app/services/hotkeys.ts
blackxored: Changed

PR #1056, opened 2018-11-27
Release multiple buckets

computerquip-streamlabs:
This supersedes both #1053 and #1052. 
Still need to do some basic testing, probably by just making a preview release and hoping for the best.

### DIFF
--- a/bin/release.js
+++ b/bin/release.js
@@ -16,7 +16,7 @@ const cp = require('child_process');
 /**
  * CONFIGURATION
  */
-const s3Bucket = 'streamlabs-obs';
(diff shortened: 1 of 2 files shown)


computerquip-streamlabs:
Actually, this obviously won't work since you can only read one set of credentials from the env at a time. Need to think on how to solve that if possible...

computerquip-streamlabs:
Never mind, just need to make sure the credentials given can upload to both buckets. Should be fine. Overthinking things lol.

computerquip-streamlabs:
I can't change permissions at all. You or an admin will afaik.

andycreeth:
Np I will change permissions now and then merge this.  Thanks.

andycreeth:
Going to test this out with today's preview and see how it goes.

PR #1064, opened 2018-11-29
Widget Bugfixes

gettinToasty:
Also apparently merging staging into staging2 lol


app/components/page-components/Chatbot/module-bases/ChatbotAlertsBase.vue.ts
```diff
@@ -1,5 +1,5 @@
 import { cloneDeep } from 'lodash';
-import { Component, Prop } from 'vue-property-decorator';
+import { Component, Prop, Inject } from 'vue-property-decorator';
```
holiber: no need for `Inject` here

gettinToasty: That's not actually one of my changes

PR #1069, opened 2018-11-29
persist disabled apps in localstorage

andycreeth:


andycreeth:
Btw this PR brings up an interesting issue... Currently you have to choose between `StatefulService` or `PersistentStatefulService`.  There is no good solution for a service that wants some of its state persisted and some of its state not persisted.

PR #1071, opened 2018-11-29
Add IntelliJ settings, including inspection profile

blackxored:
This modifies our `.gitignore` so it's more narrow as to what we exclude from the `.idea` folder. 
The main motivation is that Upsource allows us to run IntelliJ inspections while part of its automated code review features (including running TSLint :O) but it needs an inspection profile to be part of the repository.
Since code style settings would be imported from prettier I don't believe this would introduce any conflicts for Jetbrains users.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 .DS_Store
-.idea
 node_modules/
 build/
 npm-debug.log
@@ -18,3 +17,4 @@ electron.VC.db
 plugins/
 yarn-error.log
 .opt-in
+
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ electron.VC.db
 .vs/
 plugins/
 yarn-error.log
+.opt-in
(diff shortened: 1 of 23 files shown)


blackxored:
Just testing

.gitignore
```diff
@@ -16,4 +15,82 @@ electron.sln
...
+### Intellij Patch ###
+# Comment Reason: https://github.com/joeblau/gitignore.io/issues/186#issuecomment-215987721
+
+# *.iml
+# modules.xml
+# .idea/misc.xml
+# *.ipr
+
+# Sonarlint plugin
+.idea/sonarlint
```
holiber: There are too many in this file just for Idea
Thoughts about how to handle it better:
remove `.idea` folder from .gitignore here
add `.gitignore` with these rules in the `.idea` folder

blackxored: Great idea! Just moved them down.

PR #1075, opened 2018-11-30
Remove source on removal of widget preview

computerquip-streamlabs:
This fixes a major leak when you open and close a widget settings menu.
There's another leak here where if you start slobs and open the widget settings menu, it will leak the first time only. I'm still tracking that down.


computerquip-streamlabs:
This was fixed in a separate commit by coincidence.

PR #1080, opened 2018-12-03
Update Typescript, Spectron, Ava, Webdriver, Babel, RxJS, Webpack

holiber:
- TypeScript 2.9 -> 3.2 // new Ava uses d.ts files with TS 3.0 syntax
- Spectron 3.8 -> 4.0
- Ava 0.25 -> 1.0 rc2
- RxJS 5 -> 6
- Babel 6 -> 7
- WebdriverIO
- Webpack, WebpackCLI

Also, the annoying @types/jQurey dependency has been removed because of conflicts with the  webdriver typings


blackxored:
|`package.json:104`|
|---|
I wonder why we removed `babel-eslint`? We probably need for eslint even if we use babel's new TypeScript parser.

app/components/windows/Projector.vue.ts
```diff
@@ -8,7 +8,7 @@ import { ISourcesServiceApi } from 'services/sources';
 import electron from 'electron';
 import Util from 'services/utils';
 import { $t } from 'services/i18n';
-import { Subscription } from 'rxjs/subscription';
+import { Subscription } from 'rxjs';
```
blackxored: I've seen this imported as `rxjs/Subscription` in other places. Could we keep it case consistent?

holiber:
> `package.json:104`
> I wonder why we removed `babel-eslint`? We probably need for eslint even if we use babel's new TypeScript parser.

`babel-eslint` is still in deps here, what do you mean?

app/services-manager.ts
```diff
@@ -617,8 +615,10 @@ export class ServicesManager extends Service {
             }
 
             if (result.emitter === 'STREAM') {
-              const subject = new Subject<any>();
-              this.windowSubscriptions[result.resourceId] = subject;
+              let subject = this.windowSubscriptions[result.resourceId];
```
blackxored: Could this be an `or` statement, and return `this.windowSubscriptions[result.resourceId]`?

test/helpers/spectron/index.ts
```diff
@@ -1,6 +1,5 @@
 /// <reference path="../../../app/index.d.ts" />
-import 'rxjs/add/operator/first';
-import test from 'ava';
+import avaTest, { ExecutionContext, TestInterface } from 'ava';
 import { Application } from 'spectron';
```
blackxored: |`line 12`|
|---|
Relaying Upsource here. Unnecessary semi.

blackxored:
Shouldn't we include a target, now that we've migrated to preset-env? Also, can we double check we have class properties?

blackxored:
That makes sense, I thought it was used in some other places like updater, but if there are only isolated instances or none at all, then we’re good. 
—
Best Regards,

Adrian Perez 
https://adrianperez.codes

PR #1087, opened 2018-12-04
Allow YT and FB users to schedule streams from SLOBS

gettinToasty:
This also sets the stream up correctly when going live if a scheduled stream exists, making sure the stream being broadcast to is the one that was scheduled.

### DIFF
--- a/app/components/windows/EditStreamInfo.vue.ts
+++ b/app/components/windows/EditStreamInfo.vue.ts
@@ -397,7 +397,7 @@ export default class EditStreamInfo extends Vue {
       hours = hours < 10 ? `0${hours}` : hours;
       let minutes: string | number = (this.startTimeModel.time % 3600) / 60;
       minutes = minutes < 10 ? `0${minutes}` : minutes;
-      return `${dateArray[2]}-${dateArray[1]}-${dateArray[0]}T${hours}:${minutes}:00.0${moment().format('Z')}`;
+      return `${dateArray[2]}-${dateArray[0]}-${dateArray[1]}T${hours}:${minutes}:00.0${moment().format('Z')}`;
     } catch {
       this.$toasted.show(
         $t('Please enter a valid date'),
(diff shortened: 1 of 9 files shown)


kuhasuki:
Looks good to me so far. I'm gonna try it out and finish the review.

app/components/windows/EditStreamInfo.vue.ts
```diff
@@ -335,4 +389,25 @@ export default class EditStreamInfo extends Vue {
...
+
+  private formatDateString() {
+    try {
+      const dateArray = this.startTimeModel.date.split('/');
+      let hours: string | number = Math.floor(this.startTimeModel.time / 3600);
+      hours = hours < 10 ? `0${hours}` : hours;
+      let minutes: string | number = (this.startTimeModel.time % 3600) / 60;
+      minutes = minutes < 10 ? `0${minutes}` : minutes;
+      return `${dateArray[2]}-${dateArray[0]}-${dateArray[1]}T${hours}:${minutes}:00.0${moment().format('Z')}`;
+    } catch {
```
holiber: Use the `Date` object or `Timestamp` number to store and operate with dates. That allows avoiding all complications with parsing. If you prefer to store date and time in different formats you can use getter like this:
```js
get startTime() {
  return new Date(this.startTimeModel.date).setMinutes(this.startTimeModel.time)
}

get formatDateString() {
 return moment(this.startTime).format(yourFormat);
}

```

gettinToasty: are `Date` objects serializable? that's the main reason i avoided it in this case

holiber: Not sure what case are you trying to handle? date.getTime() will return the same number on the client side and server side even if there are different time zones.

gettinToasty: when you create a `new Date()` with a string i believe it defaults to UTC unless a timezone is included in the datestring, but i may be wrong here

PR #1098, opened 2018-12-06
Fix widget properties

holiber:


app/services-manager.ts
```diff
@@ -470,39 +470,22 @@ export class ServicesManager extends Service {
    *
    * @example
    * source = getResource('Source[12]')
-   *
-   * @example
-   * source = getResource('ScenesService.activeScene')
```
holiber: ^ I rolled back this feature here

app/services/widgets/widgets.ts
```diff
@@ -220,6 +218,26 @@ export class WidgetsService extends StatefulService<IWidgetSourcesState> impleme
     });
   }
 
+  private register(sourceId: string) {
+    const source = this.sourcesService.getSource(sourceId);
+    const widgetType = source.getPropertiesManagerSettings().widgetType;
+    const isWidget = typeof widgetType === 'number';
```
andycreeth: Realize this is copied from above, but it's probably safer / more correct to just use `source.getPropertiesManagerType()` instead.

PR #1099, opened 2018-12-06
Fix browser source URL

holiber:


gettinToasty:
I don't really understand this fix, it seems like you're just moving all of the same stuff to another place and then using that new function?

andycreeth:
It looks like he moved the defaults from a place where they would apply on create/update to a place where they would only apply on create.

holiber:
Exactly

andycreeth:
Confirming this resolves the issue.  Thanks Alex.

PR #1103, opened 2018-12-07
use scroll instead buttons for apps nav

holiber:
Using scrolling seems much more convenient for the big list of app names. Also, this PR fixes the issues with disappearing nav buttons in the `app-nav`


andycreeth:
@holiber It looks like this got stale.  I think this would be a good improvement.  Can you resolve conflicts so we can get this tested/merged in.

andycreeth:
Looks like TSLint failures

PR #1104, opened 2018-12-07
Sentry fixes

andycreeth:


app/services/selection/selection.ts
```diff
@@ -97,7 +97,11 @@ export class SelectionService
 
   @shortcut('Delete')
   remove() {
-    const name = this.getLastSelected().name;
+    const lastSelected = this.getLastSelected();
+
+    if (!lastSelected) return;
+
+    const name = lastSelected.name;
```
holiber: Just realized we always show only one item name even if you are going to remove several items

andycreeth: Yeah good call.  We should probably fix that too at some point.

PR #1114, opened 2018-12-07
Layout fixes

mbiemiller:
- fixes layout issue when window size is below 1100px (min width at which we allow live dock to show)
- added back footer performance labels at small width per Eric's request but reduced font size
- pulled in Alex's fix that fixes left chat toggle issue

### DIFF
--- a/app/components/windows/Main.vue.ts
+++ b/app/components/windows/Main.vue.ts
@@ -159,7 +159,7 @@ export default class Main extends Vue {
 
   mainMiddle: HTMLDivElement;
   compactView = false;
-  mainMiddleWidth: Number;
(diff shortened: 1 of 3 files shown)


app/components/windows/Main.vue.ts
```diff
@@ -176,4 +170,36 @@ export default class Main extends Vue {
...
+
+  created() {
+    window.addEventListener('resize', this.windowSizeHandler);
+  }
+
+  destroyed() {
+    window.removeEventListener('resize', this.windowSizeHandler);
+  }
+
+  windowWidth: Number;
```
andycreeth: this should be `number` (lowercase)

PR #1117, opened 2018-12-10
Validate end date for goals

gettinToasty:
Also expandable if we have other API type of errors we want to handle

### DIFF
--- a/app/components/shared/inputs/ValidatedForm.vue.ts
+++ b/app/components/shared/inputs/ValidatedForm.vue.ts
@@ -52,4 +52,8 @@ export default class ValidatedForm extends Vue {
   emitInput(data: any, event: Event) {
     this.$emit('input', data, event);
   }
+
+  get apiErrors() {
+    return this.$validator.errors.first('api');
+  }
 }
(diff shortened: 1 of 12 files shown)


app/components/shared/inputs/ValidatedForm.vue.ts
```diff
@@ -52,4 +52,8 @@ export default class ValidatedForm extends Vue {
   emitInput(data: any, event: Event) {
     this.$emit('input', data, event);
   }
+
+  get apiErrors() {
+    return this.$validator.errors.first('api');
+  }
```
holiber: Not sure it's the best way to handle errors since `ValidatedForm.vue.ts` is an abstraction that shouldn't know about API. Also, this field doesn't use any validations. You actually use it to just show a error-message. I think it's better to choose one of the next solutions:
1) Handle such errors that are not related to any field as somethink like `global` or `wildcard` errors instead `api-error`
2) Handle such errors directly in WidgetEditor or  Widget template as we did before. There is  no need for `nextTick` call in this way.

PR #1120, opened 2018-12-10
Scene Transitions API

blackxored:
Add a new module to Platforms API, Scene Transitions, responsible for providing an API for platform developers to create scene transitions programmatically. 
Currently, only stinger transitions are supported, through the facade involved in the translation between a developer-friendly API and what OBS expects it's pretty generic. 
The video files used in stinger transitions are expected to be provided as part of the app (i.e a relative path). 
There's an option for app developers to lock specific scene transitions, preventing the user from editing them. Changes to the Manage Scene Transitions dialog were implemented accordingly.


app/services/platform-apps/api/index.ts
```diff
@@ -74,30 +75,37 @@ export class PlatformAppsApi {
...
-          api[moduleName][eventName] = this.modules[moduleName][eventName];
-        } else {
-          api[moduleName][eventName] = async () => {
-            throw new Error('This app does not have permission to access this API. ' +
-              `Required permissions: ${this.modules[moduleName].permissions}`);
-          }
-        }
-      });
+        },
+      );
```
andycreeth: All of this looks like just a prettier pass.  Just want to make sure you didn't change any logic here, right?

app/services/platform-apps/api/modules/scene-transitions.ts
```diff
@@ -0,0 +1,189 @@
...
+            appId,
+            locked: shouldLock,
+          },
+          settings: {
+            ...settings,
+            path: this.platformAppsService.getAssetUrl(appId, path),
+          } as Dictionary<TObsValue>,
+        }),
+        toObsTransitionOptions(stingerTransitionDefaultOptions),
+      )(settings);
```
andycreeth: This is hard to grok.  Let's avoid `compose` until the entire team is bought it, understands the behavior, and agrees on its usefulness.

app/services/platform-apps/api/modules/scene-transitions.ts
```diff
@@ -0,0 +1,189 @@
...
+  });
+
+  return result;
+};
+
+const isVideo: (x: string) => boolean = compose(
+  (type: string | null) => /^video\/.*$/.test(type),
+  getType,
+  path.basename,
+);
```
andycreeth: This whole module is written in a highly functional style that I'm finding very difficult to parse.  I understand this is the way you like to write code, but it's not the way the rest of this codebase is written.  For now I think we should refactor this to use a more OO style that matches the rest of the codebase.

package.json
```diff
@@ -182,6 +184,7 @@
     "terser-webpack-plugin": "^1.1.0",
     "traverse": "0.6.6",
     "ts-loader": "4.4.2",
+    "ts-node": "^7.0.1",
```
andycreeth: what is this providing for us?

app/services/media-backup.ts
```diff
@@ -220,7 +224,9 @@ export class MediaBackupService extends StatefulService<IMediaBackupState> {
     let downloadedPath: string;
 
     try {
-      downloadedPath = await this.withRetry(() => this.downloadFile(data.url, serverId, data.filename));
+      downloadedPath = await this.withRetry(() =>
+        this.downloadFile(data.url, serverId, data.filename),
```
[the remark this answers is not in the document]
blackxored: Prettier, trailing commas.

app/services/platform-apps/api/modules/scene-transitions.ts
```diff
@@ -0,0 +1,189 @@
+import { flowRight as compose } from 'lodash';
+import { getType } from 'mime';
+import { apiMethod, EApiPermissions, IApiContext, Module } from './module';
+import {
+  ETransitionType,
+  ITransitionCreateOptions,
+  TransitionsService,
+} from '../../../transitions';
```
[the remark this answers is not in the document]
blackxored: Should be changed now.

app/services/platform-apps/api/modules/scene-transitions.ts
```diff
@@ -0,0 +1,189 @@
...
+import { Inject } from '../../../../util/injector';
```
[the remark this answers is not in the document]
blackxored: Same

app/services/platform-apps/api/index.ts
blackxored: We did not.

package.json
blackxored: Since we can't get rapid feedback from tests, I'm using Quokka for a REPL occasionally.

app/services/platform-apps/api/modules/scene-transitions.ts
```diff
@@ -0,0 +1,161 @@
...
+  shouldLock?: boolean;
+  /** Whether to monitor audio **/
+  shouldMonitorAudio?: boolean;
+  /** A transition point in milliseconds **/
+  transitionPoint?: number;
+  /** Type of transition point, frame or time **/
+  transitionPointType?: TransitionPointType;
+};
+
+type TransitionOptions = StingerTransitionOptions;
```
andycreeth: I'm curious why this and `StingerTransitionOptions` are written as types instead of interfaces.  I think I see what you're trying to do, but I think this would make much more sense if `TransitionOptions` were an interface, and then `StingerTransitionOptions` extended that interface.

app/services/media-backup.ts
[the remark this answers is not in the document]
andycreeth: Even for function arguments?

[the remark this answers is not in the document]
blackxored: For most things multiline, this ended up being multiple lines, so yes. It makes easier diffs, and easier if we add a new argument.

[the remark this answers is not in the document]
andycreeth: Understood.  Was having trouble understanding why it was inserted here but not elsewhere.  The multi-line thing makes sense.

app/services/platform-apps/api/modules/scene-transitions.ts
blackxored: TransitionOptions is a sum type. The fact that we only support stinger transitions at the moment doesn't make it obvious, but I believe we've talked about how we'll differentiate based on the `type` property.

PR #1121, opened 2018-12-11
fix store race condition

holiber:


andycreeth:
Closing until we can come up with a solution that addresses the above.

andycreeth:
Is this being actively worked on?  If not I am closing to reduce noise.

PR #1129, opened 2018-12-12
Big refactor: clean TSLint + AirBnB config

blackxored:
This adds a stricter TSLint config and fixes all errors. Marked as WIP because there are still issues in the codebase that TSLint doesn't catch, will update accordingly. I will close #1046 as this supersedes it.

These are some, but not all, of the changes:

* add lint commands, including CI, to `package.json`.
* remove eslint, we will invest on tslint instead.
* add `readonly` modifier to fields that deserve it.
* shorten imports.
* fix chatbot templates having invalid syntax.
* add AirBnB config through tslint and apply automatic fixes.
* manual fixes for the rest.


app/components/AppearanceSettings.vue.ts
```diff
@@ -36,14 +33,17 @@ export default class AppearanceSettings extends Vue {
...
-      componentName: 'FFZSettings',
-      title: $t('FrankerFaceZ Settings'),
-      queryParams: {},
-      size: {
-        width: 800,
-        height: 800
-      }
-    }, 'ffz-settings');
+    this.windowsService.createOneOffWindow(
+      {
```
gettinToasty: this is an example of a change i don't really like, imo it just adds lines unnecessarily and just looks kinda gross but its probably not configurable :/

app/components/ExtraSettings.vue.ts
```diff
@@ -61,7 +59,11 @@ export default class ExtraSettings extends Vue {
     this.cacheUploaderService.uploadCache().then(file => {
       electron.remote.clipboard.writeText(file);
       alert(
-        $t('Your cache directory has been successfully uploaded.  The file name %{file} has been copied to your clipboard.', { file })
+        $t(
+          // tslint:disable-next-line:max-line-length
```
gettinToasty: also this, we typically just use string concatenation in these cases which i think is a lot cleaner/more sustainable than selectively disabling rules

app/components/LiveDock.vue.ts
```diff
@@ -297,7 +290,9 @@ export default class LiveDock extends Vue {
     const minEditorWidth = 860;
     const minWidth = 290;
     const maxWidth = Math.min(appRect.width - minEditorWidth, appRect.width / 2);
+    // tslint:disable-next-line:no-parameter-reassignment TODO
     width = Math.max(minWidth, width);
+    // tslint:disable-next-line:no-parameter-reassignment
```
gettinToasty: this also seems indicative of an actual issue that should be handled rather than ignoring rules here

app/components/AppearanceSettings.vue.ts
andycreeth: Yeah this is one of those where I agree but could live with either option.  Neither option significantly impacts readability IMO.  @blackxored do you know which rule this is by any chance?

app/components/ExtraSettings.vue.ts
andycreeth: +1 lets just concat.  I wish there was a tool that did this automatically.  Maybe there is.

app/components/LiveDock.vue.ts
andycreeth: This code is pretty old, but I think it should be safe to just use a new local variable here.

app/components/custom-source-properties/StreamlabelProperties.vue.ts
```diff
@@ -77,22 +77,20 @@ export default class StreamlabelProperties extends Vue {
     }
 
     this.streamlabelsService.setSettingsForStat(
-      this.currentlySelected.settings.settingsStat ?
-        this.currentlySelected.settings.settingsStat :
-        this.currentlySelected.name,
-      this.labelSettings
+      this.currentlySelected.settings.settingsStat
+        ? this.currentlySelected.settings.settingsStat
```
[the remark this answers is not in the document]
andycreeth: I generally also tend to leave operators dangling on the previous line rather than leading the next line.  That said, I think a lot of the javascript community prefers the latter.  Ultimately this isn't one I feel very strongly about but lets see what the rest of the team things.

app/components/obs/inputs/ObsInput.ts
```diff
@@ -8,31 +8,31 @@ import {
...
-  'OBS_PROPERTY_COLOR' |
-  'OBS_PROPERTY_DOUBLE' |
-  'OBS_PROPERTY_FLOAT' |
-  'OBS_PROPERTY_SLIDER' |
-  'OBS_PROPERTY_FONT' |
-  'OBS_PROPERTY_EDITABLE_LIST' |
-  'OBS_PROPERTY_BUTTON' |
-  'OBS_PROPERTY_BITMASK' |
-  'OBS_INPUT_RESOLUTION_LIST';
+  | 'OBS_PROPERTY_BOOL'
```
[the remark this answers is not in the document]
andycreeth: TypeScript is actually technically fine it.  I think the idea here is that every line is formatted the same way, and it's easy to insert anywhere in the list, similar to the trailing comma rule.  I also somewhat cynically believe most people like this because they like the straight line all the `|`s make on the left.

app/components/AppearanceSettings.vue.ts
blackxored: I do not believe this one is configurable. I'm ambivalent about it as well, I guess there's a small point to be made about having the argument on its line (similar to multiline i.e, easy add/removal). But other than that and cosmetics I'm not sure. 
Oh, and it reads better if you're into parameter hints 😂 

capture

app/components/LiveDock.vue.ts
blackxored: About parameter reassignment, there were too many instances of this for me to feel comfortable rewriting at this moment, I've added TODO comments but can look into it in a smaller PR (like the commits I've split from it already).

app/components/custom-source-properties/StreamlabelProperties.vue.ts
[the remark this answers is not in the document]
blackxored: Hey, so... ternaries, I agree with you all, it's a non-issue if it's just a single line. When it's multiple though, I believe having it at the beginning of next line, makes it obvious the expression isn't done (otherwise you'd have to scan back to the EOL after you look at the next one and you're like "this ain't valid javascript"), also there are some examples from AirBnB here: https://github.com/airbnb/javascript#comparison--nested-ternaries

app/components/AppearanceSettings.vue.ts
andycreeth: K.  I'm assuming we can all live with this.  If anyone disagree LMK, but for now this one is resolved.

app/components/LiveDock.vue.ts
andycreeth: k

PR #1132, opened 2018-12-12
Banner Improvements

gettinToasty:
Adds nav params, fully clickable banner, and closeOnLink flag functionality

### DIFF
--- a/app/services/announcements.ts
+++ b/app/services/announcements.ts
@@ -12,7 +12,9 @@ interface IAnnouncementsInfo {
   linkTitle: string;
   thumbnail: string;
   link: string;
-  link_target: 'external' | 'slobs';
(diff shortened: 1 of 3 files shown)


app/services/announcements.ts
```diff
@@ -47,6 +51,13 @@ export class AnnouncementsService extends StatefulService<IAnnouncementsInfo> {
     const req = this.formRequest(endpoint);
     try {
       const newState = await fetch(req).then((rawResp) => rawResp.json());
+      const queryString = newState.link.split('?')[1];
```
andycreeth: Can you maybe add a comment showing what the expected link should look like with params?  I'm assuming you're parsing params out of the link field mainly so we don't have to make any backend changes?

PR #1135, opened 2018-12-13
FB Bugfixes

gettinToasty:
Catch API errors and fix game selection error not appearing

### DIFF
--- a/app/services/platforms/facebook.ts
+++ b/app/services/platforms/facebook.ts
@@ -174,13 +174,12 @@ export class FacebookService extends StatefulService<IFacebookServiceState> impl
       .then(response => response.json())
       .then(json => {
         const info = json.data.find((vid: any)=> vid.status === 'SCHEDULED_UNPUBLISHED') || json.data[0];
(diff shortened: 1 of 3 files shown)


app/services/platforms/facebook.ts
```diff
@@ -179,7 +179,8 @@ export class FacebookService extends StatefulService<IFacebookServiceState> impl
           this.SET_STREAM_URL(info.stream_url);
         }
         return info;
-      });
+      })
+      .catch(() => null);
```
holiber: What it's catching here?

gettinToasty: in cases where users have never streamed on the platform i think the data is unpopulated for prefill, i thought `json.data[0]` would just be undefined but it's possible it's throwing an exception instead

holiber: The `.catch` literally wraps the all lines of code above into `try-catch` construction. We should strive use try/catch only for parts of the code that we suspect can throw exceptions. Otherwise, all this code looks unconfident.

PR #1136, opened 2018-12-13
Fix the fallback locale

holiber:


app/services/i18n/i18n.ts
```diff
@@ -187,6 +189,10 @@ export class I18nService extends PersistentStatefulService<II18nState> implement
     return dictionary;
   }
 
+  private localeIsSupported(locale: string) {
+    return WHITE_LIST.includes(locale) && fs.existsSync(`${this.getI18nPath()}/${locale}`)
```
gettinToasty: if this is executed synchronously do we need to add an await when it's invoked or does fs handle that for us?

holiber: `fs.existsSync` is a synchronous method, Adding `await` before synchronous method makes this method asynchronous

andycreeth:
Please resolve conflicts so we can get this merged in.

andycreeth:
Looks like TSLint failures.

PR #1137, opened 2018-12-13
Fix conflicting upsource settings

holiber:
My IDE is always trying to do this change


holiber:
This branch contains only IDE settings, no tests needed

PR #1142, opened 2018-12-14
Notify users who have no face mask plugin

kuhasuki:
Show warning to users who do not have face mask plugin that face masks will not work. Should hopefully help support team with categorizing user issues.

### DIFF
--- a/app/services/facemasks.ts
+++ b/app/services/facemasks.ts
@@ -114,16 +114,24 @@ export class FacemasksService extends PersistentStatefulService<IFacemasksServic
   }
 
   startup() {
-    this.fetchFacemaskSettings().then(response => {
-      this.checkFacemaskSettings(response);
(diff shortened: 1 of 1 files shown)


app/services/facemasks.ts
```diff
@@ -144,6 +152,18 @@ export class FacemasksService extends PersistentStatefulService<IFacemasksServic
     );
   }
 
+  notifyPluginMissing() {
+    this.SET_ACTIVE(false);
+    const ok = electron.remote.dialog.showMessageBox(
+      electron.remote.getCurrentWindow(),
+      {
+        type: 'warning',
+        message: 'Unable to find face mask plugin. You will not be able to use Face Masks',
```
gettinToasty: this warning message should be wrapped in `$t()`

PR #1147, opened 2018-12-15
Selection resizing

holiber:
resizer4


app/util/rect.ts
```diff
@@ -0,0 +1,66 @@
+import { v2, Vec2 } from './vec2';
+import { ServiceHelper } from '../services/stateful-service';
+
+/**
+ * Class for a simple rectangle
+ */
+@ServiceHelper()
+export class Rect implements IRectangle {
```
holiber: the reason I created this class is because of I want to have a simple class for operations with rectangles that has no deal with fields like `croppedWidth` and `scaledWidth`

app/util/vec2.ts
```diff
@@ -0,0 +1,53 @@
+import { Vector2 } from '../../vendor/threejs/vector2';
+
+type TVecConstructorArgs = [number, number] | [IVec2] | [];
+
+/**
+ * Math for 2d vectors
+ * Use a THREE.Vector2 class as a base class
+ * @see https://threejs.org/docs/#api/en/math/Vector2
+ */
+export class Vec2 extends Vector2 {
```
holiber: I suggest using a Vector class for the geometry Math. In the ideal world it's good to have some library with classes like `Rect`, `Line`, `Vec`. Unfortunately, I didn't find any good and lightweight library for JS. So I just added a Vector2 class from the THREE.js lib, cuz it doesn't have any dependencies and well known in the JS world.

app/services/selection/selection.ts
```diff
@@ -548,6 +550,15 @@ export class Selection implements ISelection {
     this.getItems().forEach(item => item.resetTransform());
   }
 
+  scale(scale: IVec2, origin: IVec2 = AnchorPositions[AnchorPoint.Center]) {
+    const originPos = this.getBoundingRect().getOffsetFromOrigin(origin);
+    this.getItems().forEach(item => item.scaleWithOffset(scale, originPos));
+  }
+
+  scaleWithOffset(scale: IVec2, offset: IVec2) {
```
holiber: Not sure I picked the best name to explain this method. It scale the items and use an offset in x,y coords as an anchor point

holiber:
Most of the lines in this PR are code from external THREEjs library, Feel free to skip the stuff in the `vendor` directory it while reviewing

gettinToasty:
i'm not sure if we want default resize behavior to anchor to the center or the opposite corner, currently for a single item we anchor to the opposite corner correct?

holiber:
Yeah, I didn't change the anchor position, it's still the opposite corner of the resizing item. Probably it worth to change.

andycreeth:
Looks like this will need to be rebased after the tslint merge.

app/util/vec2.ts
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * shortcut for the Vec2 constructor
+ */
+export function v2(): Vec2;
+export function v2(x: number, y: number): Vec2;
+export function v2(model: IVec2): Vec2;
+export function v2(...args: [number, number] | [IVec2] | []): Vec2
+export function v2(...args: TVecConstructorArgs) {
+  return new Vec2(...args);
+}
```
andycreeth: Can you help me understand what's going on here?  Are these 5 different signatures for the same function implementation?  And TypeScript allows it because all the signatures are effectively compatible?  What's the advantage here?  More clear intellisense?

app/util/rect.ts
```diff
@@ -0,0 +1,66 @@
+import { v2, Vec2 } from './vec2';
+import { ServiceHelper } from '../services/stateful-service';
+
+/**
+ * Class for a simple rectangle
+ */
+@ServiceHelper()
```
andycreeth: I think we need to take a long hard look at `ServiceHelper` and figure out what it's future is.  We haven't really used it for much since its initial implementation.

It's strange to me that this random utility class would be using a decorator from `StatefulService` when this has nothing to do with mutations or the store.  Reading the comment on the `ServiceHelper` decorator doesn't help much, as it just talks about the fact that it can have mutations.

I'm assuming you're using this because you need to return an instance of this class from a service method and access the result in another window.  However, I think the `ServiceHelper` in its current form inherently confusing and difficult to use.  It's unclear what is actually required of classes that use this decorator.  For example, it looks like all `ServiceHelper` classes need to define a `getModel` method, but it's unclear what this should do and what it is used for.  It's also unclear to me what happens to any state on the instance that isn't recreated by constructor arguments.  I assume that just gets subtly lost when moving to a new window.  There's also a lot that happens with proxies with ServiceHelpers which seem to not be used or needed here.

Essentially what I'm saying is... until we've spent more time looking into the future of `ServiceHelper` and educating the team about it, I don't think we should be using it.  In this case you seem to only be using it to share some very simple behavior.  I don't think you gain much from having a `ServiceHelper` here.  You could simply have your service method return an `IRectangle` and instantiate this utility class in the child window.

app/util/vec2.ts
holiber: This is how Typescript recommends usage of function overloads in the docs https://www.typescriptlang.org/docs/handbook/functions.html (see Overloads)

I realized that this one is excess `export function v2(...args: [number, number] | [IVec2] | []): Vec2` because it's the same as `...args: TVecConstructorArgs`

That is mostly for better error messages and intellisense
image

app/util/rect.ts
holiber: ServiceHelper allows to safely use this class in different windows
Also, it allows to construct this class via API calls and call all its methods.
You, right I need to add more clarification for this pattern for other developers in our team.

I just double checked the code, we can omit this decorator for the `Rect` class without bad consequences

app/util/rect.ts
```diff
@@ -0,0 +1,61 @@
...
+  height: number;
+
+  constructor(options: IRectangle) {
+    this.x = options.x;
+    this.y = options.y;
+    this.width = options.width;
+    this.height = options.height;
+  }
+
+  getModel(): IRectangle {
```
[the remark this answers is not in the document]
holiber: Added this for serialization when used a ServiceHelper decorator. There is no need for this method right now

holiber:
Test passed locally, merging this

PR #1154, opened 2018-12-17
change analytics server and save events from authorized users

holiber:


app/services/usage-statistics.ts
```diff
@@ -129,7 +142,7 @@ export class UsageStatisticsService extends Service {
       product: 'SLOBS',
       version: this.version,
       count: 1,
-      uuid: this.userService.state.auth ? this.userService.state.auth.platform.id : void 0,
+      uuid: this.userService.state.auth ? this.userService.state.auth.platform.id : this.state.uuid,
```
andycreeth: What's the purpose of this?  Can we just send `slobs_user_id` in this field instead?  It's literally just a persistent uuid, which seems to be what this is.

app/services/usage-statistics.ts
```diff
@@ -142,14 +155,16 @@ export class UsageStatisticsService extends Service {
...
+      headers,
+      method: 'post',
+      body: JSON.stringify(data || {}),
+    });
     fetch(request).then(handleResponse);
   }
+
+  @mutation()
+  private SET_UUID() {
+    this.state.uuid = uuid();
```
andycreeth: I think we should back this out per above comment, but I just wanted to let you know this is unsafe.  Mutations need to deterministic, so running them in multiple windows yields the same result.  This will result in child windows having a different UUID than the main window.

app/services/usage-statistics.ts
holiber: It's for combining analytics events from unauthorized users who don't have `slobs_user_id` .
I'll add a comment here

app/services/usage-statistics.ts
holiber: Good catch

PR #1157, opened 2018-12-18
Fix audio settings

holiber:


app/services/hardware/hardware.ts
```diff
@@ -31,17 +31,25 @@ export class HardwareService extends StatefulService<IHardwareServiceState> {
...
   }
 
+  getDshowDevices() {
+    return this.state.dshowDevices;
+  }
+
+  getDshowDevice(id: string) {
+    return this.state.dshowDevices.find(device => device.id == id);
+  }
+
```
eddyStreamlabs: I would remove these if not in use as it introduces confusion on what is used or not.

PR #1158, opened 2018-12-18
Updater bootstrap improvements

computerquip-streamlabs:
This fixes and improves a couple issues with the bootstrap. Some highlights:
* Use electron-log for a consistent output of what's going on. We should now get update output in log.log instead of the console so we don't need to have the user run in a command console to see why the updater bootstrap failed.

* Use an explicit radix for parseInt because good practice.

* Specify 'utf8' encoding with readFile since it defaults to return a Buffer type even though we treat it as a string.

### DIFF
--- a/updater/bootstrap.js
+++ b/updater/bootstrap.js
@@ -71,7 +71,7 @@ async function getPreviousRoll(cacheFile, version) {
                 return;
             }
 
-            let roll = parseInt(data);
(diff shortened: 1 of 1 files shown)


computerquip-streamlabs:
Yeah, I don't see anything else that sticks out here. I've removed the WIP tag in the title. These changes are good to have regardless of any other bug present.

PR #1159, opened 2018-12-18
Scene transitions API v2

blackxored:
- [x] Ability to set a transition as default
- [x] Ability to list existing transitions
- [x] Ability to delete a transition
- [ ] Ability to update a transition
- [ ] Ability to list connections
- [x] Ability to create connections
- [x] Ability to delete connections

### DIFF
--- a/app/services/platform-apps/api/modules/scene-transitions.ts
+++ b/app/services/platform-apps/api/modules/scene-transitions.ts
@@ -23,6 +23,20 @@ enum ETransitionPointType {
   Frame = 1,
 }
 
+interface ITransition {
+  id: string;
+  name: string;
+  type: ETransitionType;
+  duration: number;
+}
+
+interface ITransitionConnection {
+  id: string;
+  fromSceneId: string;
+  toSceneId: string;
+  transitionId: string;
+}
+
 interface StingerTransitionOptions {
   /** Transition type **/
   type: 'stinger';
@@ -59,7 +73,7 @@ const stingerTransitionDefaultOptions: Partial<StingerTransitionOptions> = {
 };
 
 /**
- * This module can be used to create scene transitions.
+ * This module can be used to manage scene transitions.
  * It's useful for apps to provide both editable and uneditable transitions for the streamer.
  */
 export class SceneTransitionsModule extends Module {
@@ -81,9 +95,10 @@ export class SceneTransitionsModule extends Module {
    * @param options A description of transition options
    *
    * @see {TransitionOptions}
+   * @see {ITransition}
    */
   @apiMethod()
-  async createTransition(ctx: IApiContext, options: TransitionOptions) {
+  async createTransition(ctx: IApiContext, options: TransitionOptions): Promise<ITransition> {
     if (options.type === 'stinger') {
       const appId = ctx.app.id;
       const { url } = options;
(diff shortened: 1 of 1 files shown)


app/services/platform-apps/api/modules/scene-transitions.ts
```diff
@@ -109,6 +111,80 @@ export class SceneTransitionsModule extends Module {
...
+
+  /**
+   * Get a list of transitions belonging to this App.
+   *
+   * @param ctx API context
+   * @return A list of Transition objects belonging to the current App
+   */
+  @apiMethod()
+  async getAppTransitions(ctx: IApiContext): Promise<Array<ITransition>> {
+    return Object.keys(this.transitionsService.obsTransitions)
```
andycreeth: `this.transitionsService.obsTransitions` is really just meant to be a lookup table for OBS object references.  While this technically works, it would be more correct to rely on what's stored in state here:
```TypeScript
this.transitionsService.transitions.filter(...)
```
This should also be more efficient since you don't need to map to transitions at the end.

app/services/platform-apps/api/modules/scene-transitions.ts
```diff
@@ -109,6 +111,80 @@ export class SceneTransitionsModule extends Module {
...
+
+  /**
+   * Set a transition as the default transition
+   *
+   * @param transitionId ID of the transition to be set as default
+   * @return Whether setting the default succeeded
+   */
+  @apiMethod()
+  async setDefaultTransition(_ctx: IApiContext, transitionId: string): Promise<boolean> {
+    return tryCatchToBool(() => this.transitionsService.setDefaultTransition(transitionId));
```
andycreeth: What specific case are you trying to catch here?  The API system will automatically catch any exceptions your API handler raises, and reject the promise with the exception error message.  I'm worried that by reducing this to a boolean it may be less obvious to the developer why it isn't working.

app/services/platform-apps/api/modules/scene-transitions.ts
```diff
@@ -109,6 +111,80 @@ export class SceneTransitionsModule extends Module {
...
+  /**
+   * Get a list of transitions
+   *
+   * This includes all transitions the user has set up in SLOBS.
+   * For transitions managed by this App use `getAppTransitions`.
+   *
+   * @see {getAppTransitions}
+   */
+  @apiMethod()
+  async getTransitions(_ctx: IApiContext): Promise<Array<ITransition>> {
```
andycreeth: Rather than importing the interface and using it for the return value, you should redefine a compatible interface in this file.  It's essentially boilerplate but it serves a very important purpose:  A developer down the line may innocently make a change to `ITransition` in the transitions service, without realizing it will break many apps in the wild.  By having them be separate interfaces, when they become incompatible, TypeScript will complain and it will force the developer to write a compatibility layer to translate the data.

app/services/platform-apps/api/modules/scene-transitions.ts
blackxored: `transitions` is not exposed (at least in my testing), so I'd have to reach out into `state.transitions` which looked like a worse option, `obsTransitions` is. I'm happy to use that if we're good with the cosmetic loss of encapsulation.

andycreeth: Ah right.  We can just expose a getter for cosmetic reasons if we prefer.  I kind of go back and forth on whether encapsulating state is beneficial.  As long as a service has complete ownership over mutations and when they occur, it's OK for other parts of the code to get read-only access to that state.  Getters can be used to provide useful "views" into that state, but it's not essential.

blackxored: I've just used the other method. Have to pass `ctx` but it's not the end of the world. Using `transitions` is indeed more efficient, thought there's still an extra lookup for `propertiesManager`.

app/services/platform-apps/api/modules/scene-transitions.ts
blackxored: I've removed all of these including the error handler.

app/services/platform-apps/api/modules/scene-transitions.ts
blackxored: Fixed

PR #1161, opened 2018-12-18
feat(platform): scene item events (add, remove, update)

blackxored:
Add events for scene items. Includes add, remove and update.


andycreeth:
Sorry for forgetting about this PR.  Scene items events would still be useful.  There seems to be a bunch of extra stuff here though.  Not sure what happened, but this is the wrong commit.

PR #1170, opened 2018-12-19
refactor: deprecated start streaming

blackxored:
`startStreaming` is deprecated. This removes its last usage and adds a note about removing the method altogether.

### DIFF
--- a/app/components/windows/EditStreamInfo.vue.ts
+++ b/app/components/windows/EditStreamInfo.vue.ts
@@ -309,7 +309,7 @@ export default class EditStreamInfo extends Vue {
   }
 
   goLive() {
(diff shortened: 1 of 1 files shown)


app/services/streaming/streaming.ts
```diff
@@ -88,13 +88,15 @@ export class StreamingService extends StatefulService<IStreamingServiceState>
   /**
    * @deprecated Use toggleStreaming instead
    */
+  // TODO: no usages anymore, can we remove, including on interface?
```
andycreeth: This method is called via our socket API by some 3rd party applications.  Alex is working on separating out this API to make this more obvious, but we should remove these comments.

PR #1175, opened 2018-12-19
refactor: better handling of optional args

blackxored:
Fix instances were optional args were not passed, not defined as optional, or required but the implementation provided a default.

### DIFF
--- a/app/components/windows/MediaGallery.vue.ts
+++ b/app/components/windows/MediaGallery.vue.ts
@@ -98,13 +98,12 @@ export default class MediaGallery extends Vue {
     return this.formatBytes(this.maxUsage, 2);
   }
 
-  formatBytes(bytes: number, argPlaces: number) {
+  formatBytes(bytes: number, argPlaces: number = 1) {
     if (!bytes) {
       return '0KB';
     }
 
-    const places = argPlaces || 1;
-    const divisor = Math.pow(10, places);
+    const divisor = Math.pow(10, argPlaces);
     const base = Math.log(bytes) / Math.log(1024);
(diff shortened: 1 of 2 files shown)


app/components/windows/MediaGallery.vue.ts
```diff
@@ -98,13 +98,12 @@ export default class MediaGallery extends Vue {
     return this.formatBytes(this.maxUsage, 2);
   }
 
-  formatBytes(bytes: number, argPlaces: number) {
+  formatBytes(bytes: number, argPlaces: number = 1) {
```
holiber: You changed a required argument to an optional. So now the function allows calling itself without asking to specify what exactly the user wants to do. Did you analyze code to make this decision?

IMO changing a required value to an optional is worth to do only if the most of the functions calls assume to use the default value. But I don't see any call of `formatBytes` without the second argument = 1.

I'm ok if with this change if @gettinToasty confirm that `1` as a default arg makes sense here

gettinToasty: you can see directly below that we use `const places = argPlaces || 1;` and then base calculations off of `places`. there's a place in `MediaGallery.vue` that calls `formatBytes` without a second argument and i think that justifies a default argument. this was all just copied over from site code so i think we should take opportunities we find to clean it up :+1:

blackxored: It is called without an argument in the template, all other usages explicitly pass it. I believe it's better to have the handling here, than passing the magic constant in the template, which is what did before. Now we're just being explicit about it being optional, but please let me know if it shouldn't be, and what the correct value for the template would be if we decide to go that route.

gettinToasty: this is a good change

PR #1180, opened 2018-12-19
refactor: simplify conditionals

blackxored:
Simplify conditionals across the code base. Includes, but not limited, to:

* Direct assignment instead of branch assignment.
* Return boolean expressions directly.
* No unnecessary `else` after return.

### DIFF
--- a/app/components/obs/inputs/ObsSystemFontSelector.vue.ts
+++ b/app/components/obs/inputs/ObsSystemFontSelector.vue.ts
@@ -160,13 +160,10 @@ export default class ObsSystemFontSelector extends ObsInput<IObsInput<IObsFont>>
   }
 
   get selectedFont() {
-    return _.find(this.selectedFamily.fonts, font => {
-      if (this.value.value.flags !== this.getFlagsFromFont(font)) {
-        return false;
-      }
-
-      return true;
-    });
+    return _.find(
+      this.selectedFamily.fonts,
+      font => this.value.value.flags === this.getFlagsFromFont(font),
+    );
   }
 
   get fontsByFamily() {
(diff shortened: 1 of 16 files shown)


app/components/obs/inputs/ObsSystemFontSelector.vue.ts
```diff
@@ -162,11 +162,7 @@ export default class ObsSystemFontSelector extends ObsInput<IObsInput<IObsFont>>
 
   get selectedFont() {
     return _.find(this.selectedFamily.fonts, font => {
-      if (this.value.value.flags !== this.getFlagsFromFont(font)) {
-        return false;
-      }
-
-      return true;
+      return this.value.value.flags === this.getFlagsFromFont(font);
```
gettinToasty: this could be simplified further using implicit return but otherwise LGTM

blackxored: Removed braces are arrow function.

app/util/DragHandler.ts
```diff
@@ -207,11 +207,7 @@ export class DragHandler {
       return false;
     }
 
-    if (b.offset + b.length < a.offset) {
-      return false;
-    }
-
-    return true;
+    return b.offset + b.length >= a.offset;
```
andycreeth: Same thing as above, checks returning `false`, with a return `true` at the end.  Breaks the pattern, this isn't more readable.

app/services/source-filters.ts
```diff
@@ -121,11 +121,7 @@ export class SourceFiltersService extends Service {
       }
 
       /* Video filters can be applied to video sources. */
-      if (source.video && filterType.video) {
-        return true;
-      }
-
-      return false;
+      return source.video && filterType.video;
```
andycreeth: Same thing.  Why break the pattern of the function?  Why not combine this entire function into 1 big conditional?  (please don't do that)

PR #1182, opened 2018-12-19
refactor: unterminated statements

blackxored:
Missing semicolons, unneeded `continue`, assignments in conditions, etc.

### DIFF
--- a/build-utils/search-untranslated.js
+++ b/build-utils/search-untranslated.js
@@ -33,7 +33,7 @@ const recursive = require('recursive-readdir');
 
     const fileContent = fs.readFileSync(filePath).toString();
     let match;
-    while (match = searchExp.exec(fileContent)) {
+    while ((match = searchExp.exec(fileContent))) {
       let string = match[1];
       string = string.replace('\\', '');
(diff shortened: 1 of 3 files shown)


build-utils/search-untranslated.js
```diff
@@ -33,7 +33,7 @@ const recursive = require('recursive-readdir');
 
     const fileContent = fs.readFileSync(filePath).toString();
     let match;
-    while (match = searchExp.exec(fileContent)) {
+    while ((match = searchExp.exec(fileContent))) {
```
holiber: Don't understand this change

blackxored: `Fixed warning: 'match = searchExp.exec(fileContent)' used as condition`

I believe that by adding the parens, we communicate we *want* to do an assignment first and then use the truthiness of that as our condition. Otherwise, the reader might think we forgot a `=` or two.

andycreeth: That's kind of strange.  I'm not sure the extra patentheses actually communicate that.  Is this generally used convention?

PR #1185, opened 2018-12-19
Work around a node bug with shell and stdio options

computerquip-streamlabs:


computerquip-streamlabs:
This is being closed in favor of repairing the user's computer over working around the issue. This is caused essentially by a core service not being started correctly.

PR #1187, opened 2018-12-20
ci: lint

blackxored:
Run tslint on CI and output results JSON that could be used as an asset in the future.


gettinToasty:
for some reason the git diff is really messed up on this PR

blackxored:
Thanks for catching this, fixed now. Was the only branch that wasn't prefixed as refactor/* hence my shell wizardry missed it :D

PR #1189, opened 2018-12-20
ci: report test results to AppVeyor

blackxored:
Report test results to AppVeyor, use Ava's Tap reporter with `tap-appveyor` util. 

Screenshot:
capture

I couldn't get filename or duration, seems the test framework is the limitation.


package.json
```diff
@@ -22,7 +22,7 @@
     "release": "yarn install --cwd bin && node bin/release.js",
     "lint": "tslint --project tsconfig.json {app,guest-api,obs-api,updater}/**/*.ts",
     "lint:ci": "yarn lint --format json --out tslint-results.json",
-    "test": "tsc -p test && ava -v",
+    "test": "tsc -p test && ava --tap | tap-appveyor",
```
andycreeth: Will this prevent us running tests locally?  Also I think `test-flaky` is what is run on CI now, no?

blackxored: We can run tests locally without a problem, there's slightly different output but I don't believe it's worth a separate script. If that were to be the case, I still believe that tap output it's our best choice, and a better way would be for us to pick a reporter instead: https://github.com/sindresorhus/awesome-tap#reporters. As for the second issue, `test-flaky` invokes this internally.

blackxored: Provided separate command, `test:ci` and left `test` as before. `test-flaky` will invoke dynamic based on `CI` env var. Hopefully, this helps.

PR #1191, opened 2018-12-20
chore: add some TODO comments

blackxored:
Forgot to open this one (branch prefix bit me, twice).

### DIFF
--- a/app/services/file-manager.ts
+++ b/app/services/file-manager.ts
@@ -144,6 +144,7 @@ export class FileManagerService extends Service {
       await this.writeFile(filePath, file.data);
 
       if (version !== file.version) {
+        // TODO: throw here seems to be used as control flow
         throw new Error('Wrote out of date file!  Will retry...');
       }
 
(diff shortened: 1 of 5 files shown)


app/services/file-manager.ts
```diff
@@ -144,6 +144,7 @@ export class FileManagerService extends Service {
       await this.writeFile(filePath, file.data);
 
       if (version !== file.version) {
+        // TODO: throw here seems to be used as control flow
```
andycreeth: Kind of.  This is a tough one to get right.  If we wrote an out of date file it's a catastrophic failure the same as if the write call above had failed.  I'm not saying this couldn't be rewritten to avoid the throw here, but I doubt the end result would be more readable/maintainable.

obs-api/index.js
```diff
@@ -1,5 +1,6 @@
 "use strict";
 exports.__esModule = true;
+// TODO: can we replace this with const?
```
andycreeth: yes

PR #1197, opened 2019-01-02
Allow FB users to stream to other services

gettinToasty:


app/services/platforms/facebook.ts
```diff
@@ -219,7 +220,7 @@ export class FacebookService extends StatefulService<IFacebookServiceState>
 
   fbGoLive() {
     return new Promise(resolve => {
-      if (this.state.streamUrl) {
+      if (this.state.streamUrl && this.settingsService.getStreamService() === 'Facebook Live') {
```
andycreeth: You should be able to pull this directly from the store.  First you need to add it to the interface at the top of the settings service (it's an incomplete interface, we only define what we use).  Then you should be able to access it like `this.settingsService.state.Stream.service`

PR #1200, opened 2019-01-03
Remove disabling check stream info checkbox for FB users

gettinToasty:
This is currently required to be visible to go live on FB

### DIFF
--- a/app/components/ExtraSettings.vue
+++ b/app/components/ExtraSettings.vue
@@ -1,41 +1,51 @@
 <template>
-<div>
-  <div class="section">
-    <div class="section-content">
-      <p>{{ $t('If you are experiencing weird behavior, you can try deleting your cache directory.  This will result in you losing your scene configuration and settings, but can fix some stability issues.')}}</p>
-      <div class="input-container">
-        <a class="link" @click="showCacheDir">
-          <i class="icon-view" /> <span>{{ $t('Show Cache Directory')}}</span>
-        </a>
+  <div>
+    <div class="section">
+      <div class="section-content">
+        <p>{{ $t('If you are experiencing weird behavior, you can try deleting your cache directory. This will result in you losing your scene configuration and settings, but can fix some stability issues.')}}</p>
+        <div class="input-container">
+          <a class="link" @click="showCacheDir">
+            <i class="icon-view"/>
+            <span>{{ $t('Show Cache Directory')}}</span>
+          </a>
+        </div>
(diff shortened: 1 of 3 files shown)


app/components/ExtraSettings.vue
```diff
@@ -1,41 +1,51 @@
...
-      <div class="input-container">
-        <a class="link" @click="deleteCacheDir">
-          <i class="icon-trash" /> <span>{{ $t('Delete Cache and Restart') }}</span></a>
-      </div>
-      <div class="input-container">
-        <a class="link" @click="uploadCacheDir" :disabled="cacheUploading">
-          <i class="fa fa-upload" /> <span>{{ $t('Upload Cache to Developers') }}</span>
-          <i class="fa fa-spinner fa-spin" v-if="cacheUploading" />
-        </a>
+  <div>
```
andycreeth: Diff is unnecessarily large here because you added a level of indentation to everything.  It's actually intentional that the first level of HTML inside the template tag isn't indented.  It's a pattern used by at least a sizable chunk of the Vue community.  I think it's meant to save a level of indentation.  TBH I don't really care much either way, but this would be easier to review without the additional indentation, and many of our files are written this way.

gettinToasty: that was my linter autoformatting on save...

i guess i can try to configure it to ignore vue templates?

andycreeth: AFAIK we don't have any linting configured for `.vue` files.  I'm honestly fine with us ditching this convention, but we have to choose again between making the change all at once (and losing git blamability), or dirtying the diff of every PR that touches a Vue template for a while.

app/components/ExtraSettings.vue
```diff
@@ -2,27 +2,34 @@
 <div>
   <div class="section">
     <div class="section-content">
-      <p>{{ $t('If you are experiencing weird behavior, you can try deleting your cache directory.  This will result in you losing your scene configuration and settings, but can fix some stability issues.')}}</p>
+      <p>{{ $t('If you are experiencing weird behavior, you can try deleting your cache directory. This will result in you losing your scene configuration and settings, but can fix some stability issues.')}}</p>
```
andycreeth: Kids these days and their single spaces after periods.  🤔

gettinToasty: double spaces (depending on style guide) haven't been proper english since at least 2003 and as early as the 80s

PR #1201, opened 2019-01-03
[WIP] Replay buffer

andycreeth:


andycreeth:
This is good to go, but won't be merged until 0.11.9 is ready to go.

andycreeth:
Rolling app API into this PR.

app/components/windows/SourcesShowcase.vue
```diff
@@ -64,6 +64,15 @@
       </ul>
```
blackxored: Really nitpicky, but since this is used below as well, might make sense to extract method on this one.

blackxored:
I think `omit` returns a new object, the spread might not be needed.

blackxored:
|`app/services/platform-apps/api/index.ts:100`|
|---|
Can we avoid recreating this function for every module * event?

blackxored:
|`app/services/streaming/streaming.ts:368`|
|---|
This is a really good change IMHO :+1

blackxored:
Perf nitpicks: this creates a function every time.

andycreeth:
In this case, I don't think so.  This function is created once at app startup, no?

PR #1206, opened 2019-01-03
fix: game capture as source

blackxored:
Game capture was incorrectly being handled as scene item which resulted in double bindings, with the first one taking precedence and failing to process as it did not include the `hotkeyId` returned from OBS.

### DIFF
--- a/app/services/hotkeys.ts
+++ b/app/services/hotkeys.ts
@@ -163,15 +163,15 @@ const SOURCE_ACTIONS: HotkeyGroup = {
     up: sourceId => getSourcesService().setMuted(sourceId, true),
     shouldApply: isAudio,
   },
-  HOTKEY_START: {
-    name: 'HOTKEY_START',
+  GAME_CAPTURE_HOTKEY_START: {
+    name: 'GAME_CAPTURE_HOTKEY_START',
     description: () => $t('Capture Foreground Window'),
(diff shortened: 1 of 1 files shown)


app/services/hotkeys.ts
```diff
@@ -163,6 +163,20 @@ const SOURCE_ACTIONS: HotkeyGroup = {
     up: sourceId => getSourcesService().setMuted(sourceId, true),
     shouldApply: isAudio,
   },
+  HOTKEY_START: {
```
andycreeth: Btw I keep meaning to ask, why are these called something so insanely generic as `HOTKEY_START`?  I assume this is what they're coming back from OBS as right?  Is this going to cause us issues in the future?  What if some other source is using a hotkey name this generic.  Do you think we should do anything about this now, since this is technically a breaking change for this hotkey anyway (this hotkey will become unbound for users when we release this).

blackxored:
I was actually gonna ask about it shortly, as I’ve seen that Replay Buffer shared the same hotkey name. I hate to comeback to it but it feels more and more than we’d need a unique and stable ID coming from OBS. Open to discussing other solutions as well. 

—
Best Regards,

Adrian Perez
https://adrianperez.codes

andycreeth:
Yes, let's ask Rodrigo about this.  This doesn't seem sustainable.

PR #1208, opened 2019-01-04
Fixes for Intel 

holiber:
Fixes the issue when it's impossible to add multiple `CaptureCard` sources
Allow installing device overlay only for authorized users. Otherwise, widgets will not be correctly setup.

### DIFF
--- a/app/services/auto-config/brand-device.ts
+++ b/app/services/auto-config/brand-device.ts
@@ -13,6 +13,7 @@ import { ScenesService } from '../scenes';
 import { IpcServerService } from '../ipc-server';
 import { AudioService } from '../audio';
 import { PrefabsService } from 'services/prefabs';
+import { UserService } from '../user';
 
 interface IBrandDeviceUrls {
   system_sku: string;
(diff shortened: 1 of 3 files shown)


app/services/auto-config/brand-device.ts
```diff
@@ -13,6 +13,7 @@ import { ScenesService } from '../scenes';
 import { IpcServerService } from '../ipc-server';
 import { AudioService } from '../audio';
 import { PrefabsService } from 'services/prefabs';
+import { UserService } from '../user';
```
andycreeth: Nitpick: Import should be `import { UserService } from 'services/user';`

app/services/prefabs/prefabs.ts
```diff
@@ -1,20 +1,13 @@
...
+import { ISourceAddOptions, SourcesService, TSourceType, Source } from 'services/sources';
 import { ScenesService, TSceneNode, TSceneNodeType } from 'services/scenes';
 import { Inject } from '../../util/injector';
 import { mutation, ServiceHelper } from '../stateful-service';
 import uuid from 'uuid/v4';
 import { TObsValue } from 'components/obs/inputs/ObsInput';
 import Utils from '../utils';
 import Vue from 'vue';
-
+import { HardwareService } from '../hardware';
```
andycreeth: Nitpick: Should be a blank line after the last import.  I thought this was a prettier/ts-lint rule but maybe we don't have it enabled.  Also, import should be `import { HardwareService } from 'services/hardware';`

PR #1212, opened 2019-01-04
Fix padding on scene transitions

gettinToasty:


app/components/windows/SceneTransitions.vue
```diff
@@ -114,6 +114,12 @@
 
 <script lang="ts" src="./SceneTransitions.vue.ts"></script>
 
+<style lang="less">
+.scene-transition__tabs {
+  padding: 0 20px !important;
```
andycreeth: CSS pro-tip: Always use `!important` if you want your rules to be applied Kappa

PR #1216, opened 2019-01-07
feat: use promises to catch updater spawn errors

summeroff:
What changed. 
1) Added promises to catch errors, and also to catch successful-ish updater start. 

2) Spawn now is not starting updater directly. It start windows shell command `start` which start updater asynchronously. And only after `start` success the second command will be called. It is 'echo' command what send text message which we catch with promise as sign of success. 

3) remove explicit set of `stdio` . To get message from running app we need default stdio configuration. 

4) `detached: true` changed to calling unref later.  

What fails of updater start was checked:
* UAC block( user cancel start of updater or permit it )
* non existing updater filename 
* user does not have rights to execute updater or access to folder with updater
* updater not an executable file( for example: from server was downloaded not an updater file but some 404 page file ) 

To make testing with different conditions more easily a test app was created: 
https://github.com/summeroff/spawn_catch_test

### DIFF
--- a/updater/bootstrap.js
+++ b/updater/bootstrap.js
@@ -293,14 +293,11 @@ async function entry(info) {
     
     const updaterStartCommand = `start \"\"  \"${updaterPath}\" `  
     
-    updaterArgs.push('&& echo update_started_fine_app_can_be_closed');  
-
     log.info(updaterArgs);
 
     const update_spawned = cp.spawn(`${updaterStartCommand}`, updaterArgs, {
         cwd: info.tempDir,
         detached: false,
-        //stdio: 'ignore',
         shell: true
     });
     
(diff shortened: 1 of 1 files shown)


updater/bootstrap.js
```diff
@@ -290,17 +290,51 @@ async function entry(info) {
...
 
     log.info(updaterArgs);
 
-    cp.spawn(`${updaterPath}`, updaterArgs, {
+    const update_spawned = cp.spawn(`${updaterStartCommand}`, updaterArgs, {
         cwd: info.tempDir,
-        detached: true,
-        stdio: 'ignore',
+        detached: false,
+        //stdio: 'ignore',
```
andycreeth: Let's delete this rather than commenting it out.  If there's a reason to leave it commented out then please add a comment above explaining why/when we would want to uncomment this in the future.

updater/bootstrap.js
```diff
@@ -290,17 +290,51 @@ async function entry(info) {
...
+        update_spawned.stdout.on('data', resolve);
+    });
+
+    //wait for something to happen
+    const promise = await Promise.race([primiseError, primiseExit, primiseDataOut, primiseDataError]); 
+    log.info('Updater spawn promise ' + `result \"${promise}\"`);
+    
+    update_spawned.unref();
+
+    if(`${promise}`.startsWith("update_started_fine_app_can_be_closed") )  //hardcoded string "update_started_fine_app_can_be_closed"
```
andycreeth: So I get what you're doing here, but this seems a little hacky and I want to make sure it's definitely needed.

My understanding is that you need this because you need to know whether the updater command had a `0` exit code, but if you listen to the `exit` signal on the child process, you're actually listening for the shell exit instead?  Will the shell always exit with `0` even if the command you ran failed?  Why do we need to spawn this in a shell?

summeroff: Yes. I checked on test app and it looks that we always get 0 if updater started. `&&` check exactly for 0 to start 'echo' or not. So no need to echo and check for a string.

PR #1219, opened 2019-01-07
fix(windows): avoid creating window bigger than current display res

blackxored:
Avoid creating a child window that is bigger than the current display's resolution. Tested on `1024x768`.

### DIFF
--- a/app/services/windows.ts
+++ b/app/services/windows.ts
@@ -203,6 +203,22 @@ export class WindowsService extends StatefulService<IWindowsState> {
     // This prevents "snapping" behavior when navigating settings
     if (options.componentName !== this.state.child.componentName) options.center = true;
 
+    // Override `options.size` when what is passed in is bigger than the current display
+    if (options.size) {
+      const { width: screenWidth, height: screenHeight } = electron.screen.getDisplayMatching(
+        this.windows.main.getBounds(),
+      ).workAreaSize;
+
+      const SCREEN_PERCENT = 0.75;
+
+      if (options.size.width > screenWidth || options.size.height > screenHeight) {
+        options.size = {
+          width: screenWidth * SCREEN_PERCENT,
+          height: screenHeight * SCREEN_PERCENT,
+        };
+      }
+    }
+
     ipcRenderer.send('window-showChildWindow', options);
     this.updateChildWindowOptions(options);
   }


app/services/windows.ts
```diff
@@ -203,6 +203,20 @@ export class WindowsService extends StatefulService<IWindowsState> {
     // This prevents "snapping" behavior when navigating settings
     if (options.componentName !== this.state.child.componentName) options.center = true;
 
+    // Override `options.size` when what is passed in is bigger than the current display
+    if (options.size) {
+      const { screen } = require('electron');
+      const { width: screenWidth, height: screenHeight } = screen.getPrimaryDisplay().workAreaSize;
+      const SCREEN_PERCENT = 0.75;
```
andycreeth: Is the UX bad if this is `1.0`?  I'm hesitant because we're already making the windows smaller than they were designed in this case.  Making them even smaller makes me nervous.

app/services/windows.ts
```diff
@@ -203,6 +203,20 @@ export class WindowsService extends StatefulService<IWindowsState> {
     // This prevents "snapping" behavior when navigating settings
     if (options.componentName !== this.state.child.componentName) options.center = true;
 
+    // Override `options.size` when what is passed in is bigger than the current display
+    if (options.size) {
+      const { screen } = require('electron');
```
andycreeth: This file already has `electron` imported at the top of the file.

app/services/windows.ts
```diff
@@ -203,6 +203,20 @@ export class WindowsService extends StatefulService<IWindowsState> {
...
+      const { width: screenWidth, height: screenHeight } = screen.getPrimaryDisplay().workAreaSize;
```
andycreeth: Primary display actually isn't what you want here.  You want to get the display that this window will appear on.  SLOBS is often run on secondary displays.

app/services/windows.ts
blackxored: We can't import `screen` like that since it's apparently reserved. Run into this myself.

andycreeth: Access it as `electron.screen`

andycreeth: It's accessed this way elsewhere in the file

app/services/windows.ts
blackxored: Yeah, with no downscaling you get the main window's titlebar and then right below the child window's titlebar and content spawning the entire display. The number I chose it's by no means scientific but it's the best one I found for keeping a "center/dialog"-like look.

app/services/windows.ts
blackxored: Fixed

app/services/windows.ts
andycreeth: Got it, makes sense.

blackxored:
> Does it work fine with multiple screens?

Initially it didn't, but we get the current display now.

blackxored:
Gonna split the last commit into a separate PR as tests are not liking it.

andycreeth:
Looks like low-res styles PR includes this change, so closing this, since tests still aren't passing.  LMK if this is incorrect.

PR #1220, opened 2019-01-07
Migrate ObsBoolInput to base inputs

gettinToasty:


app/components/obs/inputs/ObsBoolInput.vue.ts
```diff
@@ -1,13 +1,27 @@
 import { Component, Prop } from 'vue-property-decorator';
 import { ObsInput, IObsInput, TObsType } from './ObsInput';
+import VFormGroup from '../../shared/inputs/VFormGroup.vue';
```
andycreeth: These imports should be absolute rather than relative:
`import VFormGroup from 'components/shared/inputs/VFormGroup.vue';`

app/components/obs/inputs/ObsBoolInput.vue.ts
```diff
@@ -1,13 +1,27 @@
...
 class ObsBoolInput extends ObsInput<IObsInput<boolean>> {
   static obsType: TObsType;
 
   @Prop()
   value: IObsInput<boolean>;
 
+  get metadata() {
+    return metadata.bool({
+      // Here we need to hide the description only if showDescription is explicitly false
+      // tslint:disable-next-line:no-boolean-literal-compare
```
andycreeth: This tslint rule is fairly dumb.  @blackxored is this designed to catch potential bugs or is this just meant for stylistic purposes?  Honestly I would be in favor of disabling this rule if there isn't a good bug-catching argument for it.

andycreeth: We are disabling this rule.

holiber:
I feel like we can completely delete this input because it doesn't add any extra logic, and replace it with a simple BoolInput. Source-properties use this input in `GenericForm.vue.ts` so it can be replaced there

gettinToasty:
@holiber i think the thinking behind keeping it in was that is conforms to the ObsInput interface which is fairly different than how we handle non-ObsInput data

holiber:
Since we are not going to use these inputs outside the source-properties I think we can isolate interfaces like `ObsInput` into `GenericForm.vue.ts`.  That is definitely more work for refactoring, but it's worth to think about it now. Let me know if you need help.

holiber:
The bigger problem that we're using ObsBoolInput in the places where we need a simple BoolInput. For example in `OverlaySettings.vue`

gettinToasty:
i was under the impression that settings were supposed to use ObsInputs, is that incorrect?

holiber:
It's true only for settings we fetch from OBS API

app/components/OverlaySettings.vue
```diff
@@ -31,7 +31,11 @@
   </div>
   <div class="section">
     <div class="section-content">
-      <ObsBoolInput :value="mediaBackupOptOut" @input="setMediaBackupOptOut" />
+      <bool-input
+        v-model="mediaBackupOptOut"
+        :metadata="{ title: $t('Do not back up my media files in the cloud (requires app restart)')}"
```
holiber: can be shortened to `:title=` here

PR #1242, opened 2019-01-10
fix: troubleshooter streaming status

blackxored:
Controls don't seem to react to streaming status changes (i.e going live/offline), this commit fixes it by subscribing to stream state events, see the message for details.

### DIFF
--- a/app/components/windows/Troubleshooter.vue.ts
+++ b/app/components/windows/Troubleshooter.vue.ts
@@ -1,6 +1,8 @@
 import Vue from 'vue';
 import moment from 'moment';
 import { Component } from 'vue-property-decorator';
+import { Subscription } from 'rxjs';
+import { debounceTime, tap } from 'rxjs/operators';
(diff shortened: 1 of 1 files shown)


app/components/windows/Troubleshooter.vue.ts
```diff
@@ -26,8 +28,16 @@ export default class Troubleshooter extends Vue {
...
   issueCode = this.windowsService.getChildWindowQueryParams().issueCode as TIssueCode;
 
+  private subscription: Subscription;
+
   mounted() {
     this.getSettings();
+    this.subscription = this.streamingService.streamingStatusChange
+      .pipe(
+        debounceTime(500),
+        tap(this.getSettings),
```
andycreeth: What's the purpose of `tap` here?  Just to be explicit about the fact that we're calling `getSettings` for the side effect and not for the return value?

blackxored: Exactly.

PR #1245, opened 2019-01-14
Migrate ObsListInput to ListInput

gettinToasty:


app/components/shared/inputs/index.ts
```diff
@@ -50,6 +50,9 @@ export interface ITimerMetadata extends INumberMetadata {
 export interface IListMetadata<TValueType> extends IInputMetadata {
   options: IListOption<TValueType>[];
   allowEmpty?: boolean;
+  loading?: boolean;
+  internalSearch?: boolean;
+  disabled?: boolean;
```
holiber: I think things like  `disabled`, `visible`,  or `enabled` must be inside `IInputMetadata `

PR #1252, opened 2019-01-15
refactor: add incremental strict null checks config/scripts

blackxored:
Starting branch. Selectively enable `strictNullChecks` by running custom script on `package.json`.


andycreeth:
Would it be possible to fail the CI build if the number of errors emitted by `yarn strict-null-check` is increased by the PR?

andycreeth:
Oh I just noticed you're only compiling individual files.  Got it.

andycreeth:
In that case we should fail CI if it emits any errors at all.

andycreeth:
What's the status on this?

blackxored:
Let's discuss this one next week.

andycreeth:
Per our discussion last week I think we should add this to our CI build and merge.

PR #1255, opened 2019-01-16
Setup Sentry as the target for crashes

RodrigoHolztrattner:
Basically a revert from https://github.com/stream-labs/streamlabs-obs/pull/928/files
This also applies for the `renderer` process (see `app.ts`)


andycreeth:
Closing and re-opening.  Hopefully code-climate should comment the ts-lint violation.

PR #1259, opened 2019-01-17
Fix source dragging on app-start

holiber:


andycreeth:
Looking at the history, it looks like this flag was added along with source grouping.  I'm going through the code and I honestly don't understand what the purpose of this flag is.  Can you maybe explain why we need this?

holiber:
Didn't remember. Will take a look

app/components/StudioEditor.vue.ts
```diff
@@ -87,13 +87,14 @@ export default class StudioEditor extends Vue {
       }
     }
 
+    // prevent dragging if the clicking is past the source
+    if (!this.getOverSource(event)) this.canDrag = false;
```
holiber: I found this bug, so added a comment here to not to forget

andycreeth:
Any idea why the tests are failing?  Can we fix?

PR #1266, opened 2019-01-18
Low-res styles

blackxored:
Style fixes for SLOBS in low resolution. I've confirmed these work and do not affect high res, but tests are failing.


andycreeth:
Any idea what's going on with tests here @blackxored ?

PR #1267, opened 2019-01-18
Add SpamSecurityInput

gettinToasty:
Now we'll finally have site parity with this input as far as look and feel goes, could be expanded into a TextValueSlider if such a need arises

### DIFF
--- a/app/components/widgets/inputs/SpamSecurityInput.vue
+++ b/app/components/widgets/inputs/SpamSecurityInput.vue
@@ -1,5 +1,5 @@
 <template>
-  <slider-input :metadata="spamMetadata" :value="localValue" @input="value => updateLocalValue(value)" />
+  <slider-input :metadata="spamMetadata" :value="localValue" @input="updateLocalValue" />
 </template>
 
 <script lang="ts" src="./SpamSecurityInput.vue.ts"></script>
--- a/app/components/widgets/inputs/SpamSecurityInput.vue
+++ b/app/components/widgets/inputs/SpamSecurityInput.vue
@@ -1,0 +1,5 @@
+<template>
+  <slider-input :metadata="spamMetadata" :value="localValue" @input="value => updateLocalValue(value)" />
+</template>
+
+<script lang="ts" src="./SpamSecurityInput.vue.ts"></script>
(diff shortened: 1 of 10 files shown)


app/components/widgets/inputs/SpamSecurityInput.vue
```diff
@@ -0,0 +1,5 @@
+<template>
+  <slider-input :metadata="spamMetadata" :value="localValue" @input="value => updateLocalValue(value)" />
```
blackxored: Can we pass this first-class?

app/components/widgets/inputs/SpamSecurityInput.vue.ts
```diff
@@ -0,0 +1,31 @@
...
+  localValue = this.optionData[this.value - this.idxMod];
+
+  get optionData() {
+    return [$t('Off'), $t('Low'), $t('Medium'), $t('High'), $t('Very High')];
+  }
+
+  get spamMetadata() {
+    return {
+      data: this.optionData,
+      ...this.metadata,
```
blackxored: Nitpick: Can we put spreads at the top of the object literals?

app/services/widgets/settings/alert-box/index.ts
```diff
@@ -83,20 +83,16 @@ export class AlertBoxService extends WidgetSettingsService<IAlertBoxData> {
...
+        tooltip: $t(
+          'This setting helps control text to speech spam, such as "wwwwwwwwwww." ' +
+            'By default this is turned off, but most users will want to set this to Low or Medium. ' +
+            'The higher the security, the less spam, but also a higher chance of a legitimate message being flagged ' +
+            'as spam and not being read.',
         ),
-        max: 5,
-        min: 1,
-        interval: 1,
+        data: [$t('Off'), $t('Low'), $t('Medium'), $t('High')],
```
blackxored: This set of options is used at least twice, wondering if it would make sense to extract.

gettinToasty: that's actually a different dataset than the default one.

app/components/widgets/inputs/SpamSecurityInput.vue.ts
gettinToasty: in this case: no; we want user-defined data props to overwrite the defaults

app/components/widgets/inputs/SpamSecurityInput.vue
gettinToasty: not sure what you mean by that

PR #1274, opened 2019-01-22
Fix ViewerCount settings failing to save

gettinToasty:
This is a bandaid but I really can't tell where we're appending `px` to the font_size property at all... I have a sneaking suspicion that the site API might be doing it


gettinToasty:
Confirmed this is a BE issue, closing in favor of a fix there

PR #1284, opened 2019-01-23
fix(settings): show cache dir opens actual directory

blackxored:
Instead of opening `%APPDATA%` and highlighting the `slobs-client` directory, this change actually opens the folder with the default file manager, which should help with UX and bug reports/support.


andycreeth:
The original idea here was to highlight the directory so it could be deleted or zipped up and shared.  We now have dedicated buttons for this, so this is probably a good change.

PR #1285, opened 2019-01-23
chore(onboarding): use generic msg as we support more services #trivial

blackxored:
Replaces the "Sign in with Twitch or Youtube account" with "Sign in with your service account".
Translations other than English would need updating.

### DIFF
--- a/app/components/pages/onboarding_steps/Connect.vue
+++ b/app/components/pages/onboarding_steps/Connect.vue
@@ -8,7 +8,7 @@
         {{ $t('We are improving our backend systems. As part of the migration process, we will need to you log in again. If you have any questions, you can') }}
         <a @click="contactSupport">$t('contact support.')</a>
       </div>
-      <div class="onboarding-desc" v-else>{{ $t('Sign in with your Twitch or Youtube account to get started with Streamlabs') }}</div>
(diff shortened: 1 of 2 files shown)


andycreeth:
Good catch.  I think we should replace "service account" with "streaming account".  I think that would be more in the vernacular.

andycreeth:
Also, is this the right way to handle it?  I think you just want to modify the key in the `en-US` file.  The key will be missing in all other files, so our translation system will recognize it as a new string to be translated.  By changing the key in all files, it will not be flagged as a string that needs translation.  CC: @holiber is that correct?

holiber:
That's correct. We store our translations in Crowdin. All files except en-Us folder are auto-generated.

blackxored:
Changed to "streaming". Is there any action required from me on this one, assuming we don't want to stick to old keys?

blackxored:
@gettinToasty I believe so, yes. @holiber, can you confirm?

andycreeth:
Yes to be clear @blackxored what you should do is:
- Add new key to `en-US`
- Remove old key from `en-US`
- DO NOT touch any other language JSON files.

andycreeth:
Aaaaand I just checked the diff and that's exactly what you did already :)  We can merge this.

PR #1287, opened 2019-01-23
fix: multiple deletion message

blackxored:
When deleting multiple sources at once, the message only shows one source's name (technically, the last selected source), which is incorrect and misleading as the user might think they're deleting just one source.

This changes the message to "Are you sure you want to remove these N items?" if multiple sources are selected.

### DIFF
--- a/app/services/selection/selection.ts
+++ b/app/services/selection/selection.ts
@@ -100,11 +100,17 @@ export class SelectionService extends StatefulService<ISelectionState>
     if (!lastSelected) return;
 
     const name = lastSelected.name;
+    const selectionLength = this.getSelection().getIds.call(this).length;
+    const message =
+      selectionLength > 1
+        ? $t('Are you sure you want to remove these %{count} items?', { count: selectionLength })
+        : $t('Are you sure you want to remove %{sceneName}?', { sceneName: name });
+
     electron.remote.dialog.showMessageBox(
       electron.remote.getCurrentWindow(),
       {
+        message,
         type: 'warning',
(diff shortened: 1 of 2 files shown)


app/services/selection/selection.ts
```diff
@@ -100,11 +100,17 @@ export class SelectionService extends StatefulService<ISelectionState>
     if (!lastSelected) return;
 
     const name = lastSelected.name;
+    const selectionLength = this.getSelection().getIds.call(this).length;
```
gettinToasty: is there a reason you're calling `getIds` explicitly through the `this` context? i just don't think i've ever seen a need for that before on this app

blackxored: I've no context as to why it is this way, so I'll answer just from what I can tell, we're using `Selection` below as a sort of mix-in, and calling the method directly means we don't have access to instance props (i.e state). There's another example with `getSelection` around.

andycreeth:
```
  private getSelection(): Selection {
    return Selection.prototype;
  }
```
@holiber can you please explain what is going on here.  This is a very strange pattern.

holiber: @blackxored use `.getSize` instead
@andycreeth we can't call access a `super` like for normal inheritance, so it's a way to call an original mixins' method from mixin instead of the overridden method. 

for example
```ts
 /**
   * @override Selection.select
   */
  select(items: TNodesList): ISelection {
    this.getSelection().select.call(this, items); // if we replace this line with `this.select(items)` it's gonna be a reqursion call
  ....
}
```

But for current case I don't see any need why we should use `this.getSelection().getIds.call(this).length`

PR #1289, opened 2019-01-23
fix(platform): app nav scroll with buttons

blackxored:
Make the arrow buttons work in the Platform Apps navbar work by programmatically scrolling it, with an included small animation/smoothness.

### DIFF
--- a/app/components/shared/HScroll.vue.ts
+++ b/app/components/shared/HScroll.vue.ts
@@ -34,6 +34,15 @@ export default class HScroll extends Vue {
     this.onInvalidateHandler();
   }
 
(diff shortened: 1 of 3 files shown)


app/components/shared/HScroll.vue.ts
```diff
@@ -34,6 +34,15 @@ export default class HScroll extends Vue {
     this.onInvalidateHandler();
   }
 
+  scrollBy(horizontal: number, vertical: number, animate: boolean = false) {
+    this.$el.scrollBy({
+      top: vertical,
+      left: horizontal,
+      behavior: animate ? 'smooth' : 'auto',
```
holiber: Didn't see this css-property before, that's cool! 👍

PR #1290, opened 2019-01-24
Handle slashes in the game name for the optimizer

holiber:


app/services/video-encoding-optimizations/video-encoding-optimizations.ts
```diff
@@ -116,19 +116,32 @@ export class VideoEncodingOptimizationService extends PersistentStatefulService<
...
+
+      try {
+        profiles = await fetch(
+          this.urlService.getStreamlabsApi(`gamepresets/${encodeURIComponent(game.toUpperCase())}`),
+        )
+          .then(handleErrors)
+          .then(camelize);
+      } catch (e) {
+        // probably some network error
+        // don't stop here
```
andycreeth: Please `console.error` the error.  Silently swallowing exceptions leads to us missing issues down the road.

app/services/video-encoding-optimizations/video-encoding-optimizations.ts
```diff
@@ -116,19 +116,32 @@ export class VideoEncodingOptimizationService extends PersistentStatefulService<
...
-      profiles = await fetch(this.urlService.getStreamlabsApi('gamepresets/DEFAULT'))
-        .then(handleErrors)
-        .then(camelize);
+      try {
+        profiles = await fetch(this.urlService.getStreamlabsApi('gamepresets/DEFAULT'))
+          .then(handleErrors)
+          .then(camelize);
+      } catch (e) {
+        // probably some network error
+        // don't stop here
```
andycreeth: Same comment as above

app/services/video-encoding-optimizations/video-encoding-optimizations.ts
```diff
@@ -116,19 +116,32 @@ export class VideoEncodingOptimizationService extends PersistentStatefulService<
       profiles = this.state.lastLoadedProfiles;
     } else {
       // try to fetch game-specific profile
-      profiles = await fetch(this.urlService.getStreamlabsApi(`gamepresets/${game.toUpperCase()}`))
-        .then(handleErrors)
-        .then(camelize);
+
+      try {
+        profiles = await fetch(
+          this.urlService.getStreamlabsApi(`gamepresets/${encodeURIComponent(game.toUpperCase())}`),
```
blackxored: This change isn't enough and does require a backend change. I believe we're gonna use a querystring param.

holiber: `encodeURIComponent` avoid rerouting to the wrong URL
There is nothing more I can do for the backend in this PR. The frontend works as expected. Honestly, I'm not sure I should spend any time for the backend right now because we are not going to add any preset for these 2 games with slashes in the name. I mean in the foreseeable future

PR #1298, opened 2019-01-24
Migrate ObsResolutionInput to ListInput

gettinToasty:


gettinToasty:
CC is fake news here LOL i'm wondering if we can configure it to follow campground rules or something like that

andycreeth:
Seems good :+1:

PR #1302, opened 2019-01-25
Refactor NumberListInput

gettinToasty:
This isn't used anywhere on the site currently but will be in the spin wheel when we put that in. Just getting ahead of potential tech debt down the road


gettinToasty:
@holiber the reason i can't do that is `ListInput` only accepts `ListInputOption<string>`

PR #1304, opened 2019-01-25
Guest API upgrades

andycreeth:
- Handle refreshes for pages with an injected guest API
- Refactor the `exposeApi` method => split into multiple documented/named private methods for better readability

For the sake of easy review:
`fix guest API not working after page refresh` this commit has the behavioral change
`refactor guest api service to split up a really big method` this commit is just splitting a big function into logical smaller ones


app/services/guest-api.ts
```diff
@@ -104,84 +108,117 @@ export class GuestApiService extends Service {
...
+          id: request.id,
+          error: true,
+        };
+
+        this.safeSend(contents, 'guestApiResponse', response);
+      });
   }
 
-  private safeSend(contents: Electron.WebContents, channel: string, msg: any) {
+  private safeSend(contents: Electron.WebContents, channel: string, msg?: any) {
```
holiber: Do we have limited options for `channel`? If so I would have a Enum or restricted string type;

andycreeth: Electron doesn't restrict them, but we should probably have all of our IPC messages enumerated somewhere.  Seems like a good refactor for another day.

PR #1308, opened 2019-01-25
Add more filter related tests

jamacanbacn:
The line `Adding and removing a source filter` was renamed to color correction since that is the filter it was using originally.

### DIFF
--- a/test/filters.js
+++ b/test/filters.js
@@ -5,8 +5,7 @@ import { addSource } from './helpers/spectron/sources';
 
 useSpectron();
 
-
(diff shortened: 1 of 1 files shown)


jamacanbacn:
My Scroll Filter test would show as a failure due to logging a vue warning. (By logging a warning we count it as a failed test)

This PR fixes the slider warning that would lead to a failed test:
https://github.com/stream-labs/streamlabs-obs/pull/1309

andycreeth:
Dope, thanks man!

PR #1312, opened 2019-01-28
Migrate ObsIntInput to NumberInput

gettinToasty:


app/components/shared/inputs/index.ts
```diff
@@ -41,6 +41,7 @@ export interface INumberMetadata extends IInputMetadata {
   min?: number;
   max?: number;
   placeholder?: string;
+  isInteger?: boolean;
```
holiber: If you're going to use the same input for `int` and `float`, what do you think about to have a bit more flexible option, something like `precision`?
```
1 // precision = 0
1.2 // precision = 1
1.23 // precision = 2
1.12345... // precision = -1
```

gettinToasty: i briefly considered that but i don't think we actually have validations on precision anywhere in the code base, but more specifically i didn't want to have to handle all currently roundings for every currency, which is the most likely case where we would use precision

app/components/TransitionSettings.vue.ts
```diff
@@ -3,9 +3,8 @@ import { Component, Prop } from 'vue-property-decorator';
 import { Inject } from 'util/injector';
 import { TransitionsService, ETransitionType } from 'services/transitions';
 import * as inputComponents from 'components/obs/inputs';
-import { TObsFormData, IObsInput } from 'components/obs/inputs/ObsInput';
+import { TObsFormData } from 'components/obs/inputs/ObsInput';
```
holiber: Unused import

gettinToasty: its used in `saveProperties`

PR #1317, opened 2019-01-29
don't distribute source maps in production

andycreeth:


andycreeth:
The `devtool` setting just controls what types of maps are generated.  We are still generating production source maps, since we upload them to sentry when we release.  This just prevents them also being included in our ASAR archive and taking up update bandwidth for no reason.

andycreeth:
i suck at git

PR #1318, opened 2019-01-29
fix(settings): hide stream info confirm setting while logged out

blackxored:
This setting is unintuitive given how when logged out the stream confirmation dialog would not appear regardless of it being enabled. We hide this setting for unauthenticated users now.

### DIFF
--- a/app/components/ExtraSettings.vue
+++ b/app/components/ExtraSettings.vue
@@ -26,7 +26,7 @@
   <div class="section">
     <div class="section-content">
       <bool-input
-        v-if="!isFacebook"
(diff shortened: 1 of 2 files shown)


app/components/ExtraSettings.vue
```diff
@@ -26,7 +26,7 @@
   <div class="section">
     <div class="section-content">
       <bool-input
-        v-if="!isFacebook"
+        v-if="isLoggedIn && !isFacebook"
```
andycreeth: This is a test of PR commenting.  Please ignore.  :)

PR #1325, opened 2019-01-30
fix(twitch): robustness checks for tags and fresh caches

blackxored:


app/services/stream-info.ts
```diff
@@ -5,7 +5,7 @@ import { Inject } from 'util/injector';
 import { StreamingService } from './streaming';
 import { HostsService } from 'services/hosts';
 import { authorizedHeaders } from 'util/requests';
-import { Subject } from 'rxjs';
+import { BehaviorSubject, Subject } from 'rxjs';
```
holiber: What's the reason for using `BehaviorSubject` doing this? Do you have issues with this event in the ChildWindow?

holiber: You can read the actual `stream-info` from the `state`, don't see why this must be changed here

andycreeth: Given it represents a changing value rather than an event, `BehaviorSubject` is probably the correct type to use here.  I assume he's doing this avoid having to separately query the initial value?

blackxored: Reading from the state did not work if the user has just logged in, we changed to BehaviorSubject to be able to get the last emitted value regardless of subscription time (this component would subscribe after emit).

holiber: > Reading from the state did not work if the user has just logged in
Seems like the must actualize the state when the user has logged in. We have an event for it.

If we are going to use `BehaviorSubject` we must think about some name convention for it to distinguish `BehaviorSubject` events from regular events

PR #1327, opened 2019-01-30
test: twitch tags

blackxored:
Add tests for Twitch tags.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "lint:ci": "yarn lint --format json --out tslint-results.json",
     "test": "tsc -p test && ava -v",
     "test-flaky": "node ./test/helpers/runner.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -88,7 +88,8 @@
     "files": [
       "test-dist/test/*.js",
       "test-dist/test/api/*.js",
-      "test-dist/test/widgets/*.js"
+      "test-dist/test/widgets/*.js",
+      "test-dist/test/twitch/**/*.ts"
```
holiber: I think we need a group of test for the GoLive window

andycreeth:
Still working on these?

blackxored:
Yes, trying to figure out why they fail on CI. Working locally.

PR #1332, opened 2019-01-31
Fix missing temp source on update

gettinToasty:
This sentry bug has a huge volume, but I'm not sure how we should handle the case where a temporary source isn't found.

Sentry entry: https://sentry.io/streamlabs-obs/streamlabs-obs/issues/838582567/


gettinToasty:
@holiber sounds good, i'll close this one then

PR #1333, opened 2019-01-31
Fix video encoder settings

holiber:


holiber:
This is a new PR for `staging` https://github.com/stream-labs/streamlabs-obs/pull/1334

PR #1336, opened 2019-01-31
[Sentry] [NEW] Unexpected token p in JSON

holiber:


app/util/requests.ts
```diff
@@ -12,8 +12,15 @@ import humps from 'humps';
...
-  return response.json().then(json => Promise.reject(json));
+  const contentType = response.headers.get('content-type');
+  const isJson = contentType && contentType.includes('application/json');
+
+  if (!response.ok) {
+    return isJson
+      ? response.json().then(json => Promise.reject(json))
+      : response.text().then(text => Promise.reject(text));
+  }
+  return isJson ? response.json() : response.text();
```
andycreeth: Might be a little cleaner like this.  Honestly just a nitpick though.
```TypeScript
  const contentType = response.headers.get('content-type');
  const isJson = contentType && contentType.includes('application/json');
  const result = isJson ? response.json() : response.text();

  return response.ok ? result : result.then(r => Promise.reject(r));
```

PR #1338, opened 2019-02-01
fix(twitch): avoid update tags with no permission

blackxored:
Do not attempt to update tags if the user doesn't have the `user:edit:broadcast` permission. Race conditions/general timing makes this check necessary as we might not have done a check prior to attempting to send channel info.

### DIFF
--- a/app/services/platforms/twitch.ts
+++ b/app/services/platforms/twitch.ts
@@ -160,13 +160,9 @@ export class TwitchService extends Service implements IPlatformService {
       body: JSON.stringify(data),
     });
 
-    // prettier-ignore
-    return Promise.all([
-      fetch(request).then(handleResponse),
-      this.setStreamTags(tags).catch(e => {
-        console.error('Failed to update tags', e);
-      }),
-    ]).then(_ => true);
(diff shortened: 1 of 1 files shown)


app/services/platforms/twitch.ts
```diff
@@ -160,13 +160,9 @@ export class TwitchService extends Service implements IPlatformService {
...
 
-    // prettier-ignore
-    return Promise.all([
-      fetch(request).then(handleResponse),
-      this.setStreamTags(tags).catch(e => {
-        console.error('Failed to update tags', e);
-      }),
-    ]).then(_ => true);
+    return Promise.all([fetch(request).then(handleResponse), this.setStreamTags(tags)]).then(
+      _ => true,
```
andycreeth: That single character savings over `() => true` 😄

blackxored: LOL, there's science behind it, I'd like to believe. I like to preserve the fact we do take an argument and we're ignoring it, both of which are explicit with this variation. But yeah, at the end what stands is the character length ;). At least it's not as (potentially) bad as `constTrue` and `constFalse`, happy to talk about those :D.

PR #1339, opened 2019-02-01
refactor(streaming): platform capabilities

blackxored:
Fine-grained specification of platform capabilities. Initial work, as further refactor into existing code isn't done yet. 

This provides a way for platforms to specify what capabilities they support, relieving our `IPlatformService` of optional functions as code grows and eliminating type casts when accessed dynamically through `getPlatformService`.

On the documentation side, we'd have a clearer view as to what we support for each platform, which could also tie into Platform apps.

Supported capabilities:

* `chat`: Display and interact with chat.
* `tags`: Fetch and set stream tags.
* `user-info`: Fetch and set user info.
* `viewer-count`: Fetch viewer count.
* `schedule-stream`: Schedule streams for a later date.
* `scope-validation`: Ability to check whether we're authorized to perform actions that require a given scope.

Usage example:

``` javascript
const platform = getPlatformService(...)
if (platform.supports('tags') {
  // platform is IPlatformService & IPlatformCapabilityTags on this branch
} else {
 // platform is just IPlatformService
}
```

Future abstractions such as this would be possible now:

```javascript
runIfSupports('tags', this.getAllTags)
```

Or decorators, pick your poison :D.

Capability list by platform:

* Twitch:
  - `chat`
  - `scope-validation`
  - `tags`
  - `user-info`
  - `viewer-count`
* Mixer:
  - `chat`
  - `viewer-count`
* Facebook:
  - `chat`
  - `user-info`
  - `viewer-count`
  - `stream-schedule`
* Youtube:
  - `chat`
  - `stream-schedule`

Feel free to add to this list.

### DIFF
--- a/app/services/platforms/facebook.ts
+++ b/app/services/platforms/facebook.ts
@@ -1,5 +1,12 @@
 import { StatefulService, mutation } from '../stateful-service';
-import { IPlatformService, IPlatformAuth, IChannelInfo, IGame } from '.';
+import {
+  IPlatformService,
+  IPlatformAuth,
+  IChannelInfo,
+  IGame,
+  TPlatformCapability,
+  TPlatformCapabilityMap,
+} from '.';
 import { HostsService } from '../hosts';
 import { SettingsService } from '../settings';
 import { Inject } from '../../util/injector';
@@ -35,6 +42,13 @@ export class FacebookService extends StatefulService<IFacebookServiceState>
   @Inject() hostsService: HostsService;
   @Inject() settingsService: SettingsService;
   @Inject() userService: UserService;
+
+  capabilities = new Set<TPlatformCapability>([
+    'chat',
+    'user-info',
+    'viewer-count',
+    'stream-schedule',
+  ]);
 
   authWindowOptions: Electron.BrowserWindowConstructorOptions = { width: 800, height: 800 };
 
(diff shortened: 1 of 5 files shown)


app/services/platforms/facebook.ts
```diff
@@ -36,6 +43,13 @@ export class FacebookService extends StatefulService<IFacebookServiceState>
   @Inject() settingsService: SettingsService;
   @Inject() userService: UserService;
 
+  capabilities = new Set([
+    'chat',
+    'user-info',
+    'viewer-count',
+    'stream-schedule',
+  ] as TPlatformCapability[]);
```
andycreeth: Can we avoid the typecast here?  TypeScript supports generic sets: `new Set<TPlatformCapability>(...)`

app/services/platforms/index.ts
```diff
@@ -19,9 +20,61 @@ export interface IGame {
...
+  'user-info': IPlatformCapabilityUserInfo;
+  /** Fetch viewer count **/
+  'viewer-count': IPlatformCapabilityViewerCount;
+  /** Schedule streams for a latter date **/
+  'stream-schedule': IPlatformCapabilityScheduleStream;
+  /** Ability to check whether we're authorized to perform actions under a given scope **/
+  'scope-validation': IPlatformCapabilityScopeValidation;
+};
+
+export type TPlatformCapability = keyof TPlatformCapabilityMap;
```
andycreeth: Nifty.  Didn't know about using `keyof` like this.

PR #1343, opened 2019-02-04
refactor: set Preview's error sample rate to 1

blackxored:
Set Sentry's sample rate to `1.0` if we're on Preview.

### DIFF
--- a/app/app.ts
+++ b/app/app.ts
@@ -26,6 +26,7 @@ import electronLog from 'electron-log';
 const { ipcRenderer, remote } = electron;
 const slobsVersion = remote.process.env.SLOBS_VERSION;
 const isProduction = process.env.NODE_ENV === 'production';
+const isPreview = !!remote.process.env.SLOBS_PREVIEW;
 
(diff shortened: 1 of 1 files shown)


app/app.ts
```diff
@@ -26,6 +26,7 @@ import electronLog from 'electron-log';
 const { ipcRenderer, remote } = electron;
 const slobsVersion = remote.process.env.SLOBS_VERSION;
 const isProduction = process.env.NODE_ENV === 'production';
+const isPreview = !!process.env.SLOBS_PREVIEW;
```
andycreeth: So this is an easy trap to fall into, but this early in the app initialization process, not all env vars will have been copied over from the main process, and this particular var is set in the main process.  You can make sure you have the right value here by checking directly with the main process by changing this to `remote.process.env.SLOBS_PREVIEW`

blackxored: Good one, I've updated the check. Since you said copying, there isn't any chance we'd get an empty string that'd fail this truthy check, right?

andycreeth: Don't think so.

PR #1344, opened 2019-02-04
fix rollout script

andycreeth:
- Fix error handling to properly surface errors
- Fix chance argument checking
- Accept `0` as a valid chance value
- Use positive exit codes

### DIFF
--- a/bin/aws-wrapper.js
+++ b/bin/aws-wrapper.js
@@ -20,4 +20,8 @@ args.push('--access-key', accessKey);
 
 console.log(`Wrapping ${process.argv[2]} with arguments ${args}`);
 
-cp.fork(process.argv[2], args);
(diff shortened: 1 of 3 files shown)


bin/aws-wrapper.js
```diff
@@ -20,4 +20,8 @@ args.push('--access-key', accessKey);
 
 console.log(`Wrapping ${process.argv[2]} with arguments ${args}`);
 
-cp.fork(process.argv[2], args);
+const proc = cp.fork(process.argv[2], args);
+
+proc.on('exit', code => {
```
blackxored: I'm a madman for extra chars apparently 😂, but I'd write this as:

```javascript
proc.on('exit', process.exit)
```

PR #1345, opened 2019-02-04
fix: update source has null source ID

blackxored:
While we're pretty sure the error here is that we're getting an update object which doesn't specify source ID, it's hard to track who's sending this. For now, we add a breadcrumb so we can debug further.


blackxored:
I believe we should close this one then?

PR #1353, opened 2019-02-05
Refactor app platform to use BrowserView instead of webview tags

andycreeth:
This is stage 1 of migrating from `webview` to `BrowserView`, which is the future of embedding remote content in Electron.  I started with App Platform because it is the biggest and most complex use case of webviews, and if we can make it work here then it can work anywhere.  There are also substantial benefits of using `BrowserView` for App Platform as we have far more control over the lifecycle and don't need to use CSS hacks to keep them running in the background.


app/components/PlatformAppPageView.vue.ts
```diff
@@ -0,0 +1,87 @@
+import Vue from 'vue';
+import { Component, Prop } from 'vue-property-decorator';
+import { PlatformAppsService, EAppPageSlot } from 'services/platform-apps';
+import { Inject } from 'util/injector';
+import electron from 'electron';
```
blackxored: Nitpick: can we move module imports to the top?

app/components/PlatformAppPageView.vue.ts
```diff
@@ -0,0 +1,87 @@
...
+    this.loadSub = this.platformAppsService.appLoad.subscribe(app => {
+      if (this.appId === app.id) {
+        this.unmountContainer();
+        this.mountContainer();
+      }
+    });
+  }
+
+  mountContainer() {
+    this.containerId = this.platformAppsService.mountConatiner(
```
blackxored: There's a typo in this method. Would it make sense to refactor it so it takes an object instead, unclear how the two last args differ?

app/components/PlatformAppPageView.vue.ts
```diff
@@ -0,0 +1,87 @@
...
+    );
+  }
+
+  destroyed() {
+    if (this.resizeInterval) clearInterval(this.resizeInterval);
+    this.loadSub.unsubscribe();
+    this.unmountContainer();
+  }
+
+  currentPosition: IVec2;
```
blackxored: Could we move these up, before methods?

app/components/PlatformAppPageView.vue.ts
```diff
@@ -0,0 +1,62 @@
...
+    this.platformAppsService.unmountContainer(this.containerId);
+  }
+
+  currentPosition: IVec2;
+  currentSize: IVec2;
+
+  checkResize() {
+    const rect = this.$refs.appContainer.getBoundingClientRect();
+
+    if (
```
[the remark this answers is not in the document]
blackxored: We should probably extract method on this one

app/components/pages/PlatformAppMainPage.vue.ts
```diff
@@ -0,0 +1,23 @@
...
+  @Prop() params: { appId: string };
+  @Inject() platformAppsService: PlatformAppsService;
+
+  get pageSlot() {
+    return EAppPageSlot.TopNav;
+  }
+
+  get poppedOut() {
+    return !!this.platformAppsService
+      .getApp(this.params.appId)
```
blackxored: Couldn't this fail, even if we control `params.appId`, for instance, when an app gets removed potentially?

app/services/platform-apps/api/modules/display.ts
```diff
@@ -67,40 +50,71 @@ export class DisplayModule extends Module {
...
-        displayEntry.webviewPosition.x + displayEntry.position.x,
-        displayEntry.webviewPosition.y + displayEntry.position.y,
-      );
-    } else {
-      displayEntry.display.resize(0, 0);
+  private updateDisplay(displayEntry: IDisplayEntry, transform: IBrowserViewTransform) {
+    // The container isn't mounted, so get rid of any active displays and stop
+    if (!transform.mounted) {
+      if (displayEntry.display) {
+        displayEntry.display.destroy();
```
blackxored: Could we extract method on this one into `disposeDisplay` or something like that? We could reuse below.

app/services/platform-apps/api/modules/display.ts
```diff
@@ -67,40 +50,71 @@ export class DisplayModule extends Module {
...
+      displayEntry.display = null;
+    }
+
+    // There is no active display, so create one
+    if (!displayEntry.display) {
+      // Create display
+      displayEntry.display = new Display(displayEntry.displayId, {
+        electronWindowId: transform.electronWindowId,
+        slobsWindowId: transform.slobsWindowId,
+        paddingColor: displayEntry.options.paddingColor,
```
blackxored: Just wondering why do we need style-like props when we're apparently dealing with display/window mgmt code.

app/services/platform-apps/container-manager.ts
```diff
@@ -0,0 +1,360 @@
+import { ILoadedApp, EAppPageSlot } from '.';
```
blackxored: Import order?

app/services/platform-apps/container-manager.ts
```diff
@@ -0,0 +1,360 @@
...
+
+  /**
+   * Registers an app with the container service.
+   * Any persistent pages will immediately be spun
+   * up in the background.
+   * @param app The app to register
+   */
+  registerApp(app: ILoadedApp) {
+    // Make sure this app isn't already registered
+    this.unregisterApp(app);
```
blackxored: Do we have a more reliable way of checking this? I could be concerned about side-effects that `unregisterApp` might bring. For instance, when listening for app load/unload events in platform.

EDIT: Apparently none. Still, a bit hard to reason about.

app/services/platform-apps/container-manager.ts
```diff
@@ -0,0 +1,360 @@
...
+      electronWindowId,
+      slobsWindowId,
+      mounted: true,
+    });
+
+    return containerInfo.container.id;
+  }
+
+  setContainerBounds(containerId: number, pos: IVec2, size: IVec2) {
+    const cont = this.containers.find(cont => cont.container.id === containerId);
```
blackxored: Should we rename this to `current` or similar?

app/services/platform-apps/container-manager.ts
```diff
@@ -0,0 +1,360 @@
...
+
+    if (!cont) return;
+
+    const transform = cont.transform.getValue();
+
+    const win = electron.remote.BrowserWindow.fromId(electronWindowId);
+    // This method was added in our fork
+    (win as any).removeBrowserView(cont.container);
+
+    // If these are different, it means that another window (likely the main)
```
blackxored: Nitpick: block comment

app/services/platform-apps/container-manager.ts
```diff
@@ -0,0 +1,360 @@
...
+      }
+    }
+  }
+
+  /**
+   * Refreshes all currently running containers for the provided app.
+   * @param app The app to refresh
+   */
+  refreshContainers(app: ILoadedApp) {
+    this.containers.forEach(cont => {
```
blackxored: Would prefer a filter to make it clearer.

app/services/platform-apps/container-manager.ts
```diff
@@ -0,0 +1,360 @@
...
+      if (cont.appId === app.id) cont.container.webContents.reload();
+    });
+  }
+
+  private getContainerInfoForSlot(app: ILoadedApp, slot: EAppPageSlot): IContainerInfo {
+    const existingContainer = this.containers.find(
+      cont => cont.appId === app.id && cont.slot === slot,
+    );
+
+    if (existingContainer) {
```
blackxored: We could consider shortening this conditional.

app/services/platform-apps/container-manager.ts
```diff
@@ -0,0 +1,360 @@
...
+
+    view.webContents.loadURL(this.getPageUrlForSlot(app, slot));
+
+    this.containers.push(info);
+
+    return info;
+  }
+
+  private destroyContainer(containerId: number) {
+    const cont = this.containers.find(cont => cont.container.id === containerId);
```
blackxored: Really not sure what these variable names would mean, this line and below.

app/services/platform-apps/container-manager.ts
```diff
@@ -0,0 +1,360 @@
...
+    this.containers = this.containers.filter(c => c.container.id !== containerId);
+
+    // Electron types are incorrect here.  This method exists and is documented, but
+    // does not appear in the type definitions.
+    (cont.container as any).destroy();
```
blackxored: Can we ensure we'll always get a result from the `find` call above? Otherwise, that'll blow on `undefined`.

app/services/platform-apps/index.ts
```diff
@@ -645,24 +505,31 @@ export class PlatformAppsService extends StatefulService<IPlatformAppServiceStat
...
     const app = this.getApp(appId);
-    const url = this.getAssetUrl(appId, page);
-    return `${url}?app_token=${app.appToken}`;
+    if (!app) return null;
+
+    return this.containerManager.getAssetUrl(app, asset);
   }
 
-  getAssetUrl(appId: string, asset: string) {
+  mountConatiner(
```
blackxored: Typo from before.

app/services/platform-apps/api/modules/display.ts
andycreeth: The "display" is the OBS name for the thing that shows the video output.  It's aspect ratio locked, so it includes some padding/gutter to fit into whatever size box you give it.

app/services/platform-apps/container-manager.ts
andycreeth: Yeah all unregister app does is shut down any running containers.  This is an additional layer of security to make sure that we aren't already running containers for this app before we start up new ones.  Perhaps this method doesn't have the best name, but I think the logic is sound.

PR #1357, opened 2019-02-07
Fix deflection when set to 0

gettinToasty:
Also migrates the slider to our slider input

### DIFF
--- a/app/services/audio/audio.ts
+++ b/app/services/audio/audio.ts
@@ -204,8 +204,8 @@ export class AudioService extends StatefulService<IAudioSourcesState> implements
 
   setFader(sourceId: string, patch: Partial<IFader>) {
     const obsFader = this.sourceData[sourceId].fader;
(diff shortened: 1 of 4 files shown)


app/services/audio/audio.ts
```diff
@@ -203,14 +204,11 @@ export class AudioService extends StatefulService<IAudioSourcesState> implements
 
   setFader(sourceId: string, patch: Partial<IFader>) {
     const obsFader = this.sourceData[sourceId].fader;
-
-    if (patch.deflection) obsFader.deflection = patch.deflection;
-    if (patch.mul) obsFader.mul = patch.mul;
+    if (![null, undefined].includes(patch.deflection)) obsFader.deflection = patch.deflection;
```
andycreeth: `if (patch.deflection != null)` is equivalent.  This is the one case where it's ok to use `==` instead of `===` because it checks for both `null` and `undefined` and nothing else, which usually you treat the same.

PR #1360, opened 2019-02-07
Fix SourceShowcase scroll

gettinToasty:
sourceshowcase


gettinToasty:
merging this despite CC violation because it seems like unavoidable boilerplate

PR #1364, opened 2019-02-11
Fix HFormGroup input alignments

gettinToasty:


app/components/windows/WidgetEditor.vue
```diff
@@ -119,15 +119,13 @@
   }
 
   .top-settings {
-    .input-container {
-      margin-bottom: 0;
-      width: auto;
+    .row.alignable-input {
+      width: 80px;
```
blackxored: I'd prefer if we didn't use fixed widths, but since it was here before I don't see a huge issue with it.

gettinToasty: this is a pretty specific use-case; most of our inputs won't have them but according to the design spec this is how they should look

PR #1366, opened 2019-02-11
chore(package): recursive-readdir version clash

blackxored:
Fix `recursive-readdir` warning, we don't need it on dev deps. 

```
yarn why v1.13.0
warning package.json: "dependencies" has dependency "recursive-readdir" with range "^2.2.2" that collides with a dependency in "devDependencies" of the same name with version "2.2.2"
[1/4] Why do we have the module "recursive-readdir"...?
[2/4] Initialising dependency graph...
warning slobs-client@0.11.14: "dependencies" has dependency "recursive-readdir" with range "^2.2.2" that collides with a dependency in "devDependencies" of the same name with version "2.2.2"
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "recursive-readdir@2.2.2"
info Has been hoisted to "recursive-readdir"
info This module exists because it's specified in "dependencies".
info Disk size without dependencies: "68KB"
info Disk size with unique dependencies: "112KB"
info Disk size with transitive dependencies: "216KB"
info Number of shared dependencies: 2
Done in 0.80s.
```


holiber:
It's using in the util for searching untranslated strings in the project 
`build-utils\search-untranslated.js`

holiber:
So we have to remove it from deps, not from `dev-deps`

PR #1372, opened 2019-02-12
wait on autosave disable

andycreeth:
- Resolves a race condition where we could start deloading the application state while an autosave is still in progress.
- When entering the application loading state, wait for 200ms to give the browser time to paint kevin before we start doing slow synchronous operations.

### DIFF
--- a/app/services/user.ts
+++ b/app/services/user.ts
@@ -13,7 +13,6 @@ import { PlatformAppsService } from 'services/platform-apps';
 import { getPlatformService, IPlatformAuth, TPlatform, IPlatformService } from './platforms';
 import { CustomizationService } from 'services/customization';
 import * as Sentry from '@sentry/browser';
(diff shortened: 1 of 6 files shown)


app/services/user.ts
```diff
@@ -86,7 +84,8 @@ export class UserService extends PersistentStatefulService<IUserServiceState> {
     // actually log in from integration tests.
     electron.ipcRenderer.on('testing-fakeAuth', async (e: Electron.Event, auth: IPlatformAuth) => {
       const service = getPlatformService(auth.platform.type);
-      this.login(service, auth);
+      await this.login(service, auth);
+      this.onboardingService.next();
```
holiber: what we skip here?

andycreeth: We actually don't skip anything, we just need to continue from the login step, since simply setting the login state isn't enough.

PR #1375, opened 2019-02-13
Port a few inputs to tsx

gettinToasty:
Handles boolean, toggle, and text inputs to illustrate a few different cases of how we should be handling inputs in TSX.


app/components/shared/inputs/TextInput.tsx
```diff
@@ -0,0 +1,66 @@
...
+import styles from './TextInput.m.less';
+import { BaseInput } from './BaseInput';
+import { ITextMetadata } from './index';
+import { $t } from 'services/i18n';
+
+@Component({})
+export default class TextInput extends BaseInput<string, ITextMetadata> {
+  @Prop() readonly value: string;
+
+  @Prop({ default: () => ({}) })
```
andycreeth: I'm a little confused as to what is going on here.  Why is the default value a function?

app/components/shared/inputs/TextInput.tsx
```diff
@@ -0,0 +1,66 @@
...
+    return {
+      ...super.getValidations(),
+      date_format: this.options.dateFormat,
+      max: this.options.max,
+      min: this.options.min,
+      alpha_num: this.options.alphaNum,
+    };
+  }
+
+  toggleVisibleButton(h: Function) {
```
andycreeth: Ouch so any function that returns XML needs `h` in context I guess?  I'm surprised there isn't a compiler to handle this for us.  Not a huge deal, just mildly annoying.

app/components/shared/inputs/ToggleInput.m.less
```diff
@@ -46,7 +26,7 @@
   background-color: @white;
 
 }
-.night-theme {
+:global(.night-theme) {
```
andycreeth: I'm a CSS module noob.  Can you explain what is going on here?

app/components/shared/inputs/BoolInput.tsx
```diff
@@ -0,0 +1,23 @@
+import { Component, Prop } from 'vue-property-decorator';
+import { BaseInput } from './BaseInput';
+
+@Component({})
+export default class BoolInput extends BaseInput<boolean, { name: string; title: string }> {
+  @Prop() readonly value: boolean;
+  @Prop() readonly title: string;
```
andycreeth: Is there anything else required to get type checking for these props?  I thought there was more boilerplate involved here.

andycreeth:
Thank you for converting some more complex components to test our TSX implementation.

app/components/shared/inputs/TextInput.tsx
gettinToasty: i don't remember actually but that's not a change in the code. i think it's related to the metadata helper function or something like that

app/components/shared/inputs/TextInput.tsx
gettinToasty: yea that was actually something i was planning to look into-- might give it some time today

app/components/shared/inputs/ToggleInput.m.less
gettinToasty: this puts `.night-theme` in the global scope, which means the identifying module hash won't be appended to it on compile because it's not a local style on the module

app/components/shared/inputs/BoolInput.tsx
gettinToasty: nope! `BaseInput` is where the boilerplate exists in that regard. boilerplate will always live on the (formerly) `extends Vue` layer in the inheritance chain, but once that is defined you can inherit off of it just fine.

app/components/shared/inputs/TextInput.tsx
holiber: Vue requires using a function for default values if the default value is Object

andycreeth: Huh.  Interesting.

app/components/shared/inputs/ToggleInput.m.less
andycreeth: Ah got it.  ty

PR #1377, opened 2019-02-13
Add audio settings from OBS

eddyStreamlabs:


eddyStreamlabs:
Requires an update of osn.

app/services/settings/settings.ts
```diff
@@ -303,7 +304,10 @@ export class SettingsService extends StatefulService<ISettingsState>
   }
 
   setSettings(categoryName: string, settingsData: ISettingsSubCategory[]) {
-    if (categoryName === 'Audio') return this.setAudioSettings(settingsData);
+    if (categoryName === 'Audio') {
+      this.setAudioSettings([settingsData[1]]);
+      settingsData.pop();
```
gettinToasty: why are you mutating the settingsData here? if settingsData[1] is the last element you could write
```
this.setAudioSettings([settingsData.pop()])
```

but this seems kind of weird regardless, why not just pass in the single element as the parameter here?

eddyStreamlabs: The second element of the array needs to be passed to `setAudioSettings`, then the rest of the `setSettings` function should only do treatment with the first element of the array.

Buy you"re right, I can do what you suggest.

eddyStreamlabs: Updated

PR #1379, opened 2019-02-13
Remove debounce from NumberInput

gettinToasty:
I'm not sure why this was here in the first place as I'm having no issues without it, but was causing a really unresponsive UX in int inputs

edit: fixed with a variation on the debounce concept to handle minimum values correctly whilst maintaining performance
intinput

### DIFF
--- a/app/components/shared/inputs/NumberInput.vue
+++ b/app/components/shared/inputs/NumberInput.vue
@@ -10,7 +10,7 @@
         type="text"
         ref="input"
         :placeholder="options.placeholder"
-        :value="value"
+        :value="displayValue"
         @input="handleInput($event.target.value)"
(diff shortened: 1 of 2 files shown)


andycreeth:
Generally the debounce solves a lot of issues when dealing with OBS.  For example, when you are trying to type in a value of `1000`, when the valid range is `300` to `10000`.  As soon as you type in that `1`, it will get sent to OBS and change to `300`.

andycreeth:
A whole second seems maybe a little long though

gettinToasty:
@andycreeth i was just talking about this in slobs-frontend: i don't think we have anywhere on the app that uses this input and has a min value of something greater than one. i think it may be an artifact of a time before we had slider inputs

PR #1381, opened 2019-02-14
Fix slider and arrow scroll jank

gettinToasty:
Also removes inline error field from NumberInput but that isn't really a fix, will investigate validations further


gettinToasty:
@andycreeth that error message portion was carried over in the port of ObsIntInput-- our errors are usually handled by the form group and ValidatedForm

PR #1397, opened 2019-02-19
Fix empty BoolInput label

gettinToasty:
TSX handles special characters differently than Vue templates do-- this'll be a rare case but a gotcha to look out for


app/components/shared/inputs/BoolInput.tsx
```diff
@@ -15,7 +15,7 @@ export default class BoolInput extends BaseInput<boolean, { name: string; title:
       <div class="input-wrapper" data-role="input" data-type="bool" data-name={this.options.name}>
         <div class="checkbox" onClick={this.handleClick}>
           <input type="checkbox" checked={this.value} />
-          <label>{this.options.title || '&nbsp;'}</label>
+          <label>{this.options.title || '\u00A0'}</label>
```
holiber: I suggest adding a comment here that `u00A0` is `nbsp`. Or wrap in in a variable like `Utils.nbsp`

PR #1404, opened 2019-02-20
chore: remove IDEA settings

blackxored:
If we don't chose to sync settings, watchers, code style, etc for those of us using IntelliJ
products, having these files have created more churn than benefits. We're also not using UpSource at the moment, so opting to remove them for now.


andycreeth:
https://github.com/stream-labs/streamlabs-obs/pull/1395

blackxored:
Closing in favor of the above.

PR #1408, opened 2019-02-20
Fix NumberInput crash

gettinToasty:
This is a bandaid to give more time to investigate issues with our input validator


app/components/shared/inputs/NumberInput.vue.ts
```diff
@@ -31,6 +31,7 @@ export default class NumberInput extends BaseInput<number | string, INumberMetad
 
     if (this.options.min !== void 0 && Number(value) < this.options.min) {
       this.timeout = window.setTimeout(() => this.updateValue(`${this.options.min}`), 1000);
+      return;
```
andycreeth: Does this deserve a TODO comment or something?

gettinToasty: i can add one

PR #1415, opened 2019-02-22
test: recording in every format

blackxored:
Test that we record with no errors. Also tests every format. 
Working on verifying that the files are actually valid video files, it seems everything under the sun that could do this requires binaries/bindings. I'm still exploring here, would probably go with `mmmagic`.

### DIFF
--- a/test/recording.ts
+++ b/test/recording.ts
@@ -1,0 +1,57 @@
+import * as path from 'path';
+import { tmpdir } from 'os';
+import { mkdtemp, readdir } from 'fs-extra';
+import { focusChild, focusMain, test, useSpectron } from './helpers/spectron';
+import { setFormDropdown, setFormInput } from './helpers/spectron/forms';
+import { sleep } from './helpers/sleep';
+
+useSpectron();
+
+test('Recording', async t => {
+  const tmpDir = await mkdtemp(path.join(tmpdir(), `slobs-recording-`));
+  const { app } = t.context;
+
+  /*
+   * Set recording path to a temp dir
+   */
+  await focusMain(t);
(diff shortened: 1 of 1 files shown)


test/recording.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    await focusChild(t);
+    await app.client.click('li=Output');
+    await setFormDropdown(t, 'Recording Format', format);
+    await app.client.click('button=Done');
+    await focusMain(t);
+
+    // Start recording
+    await app.client.click('.record-button');
+    await sleep(2000);
```
holiber: Use `waitForVisible` if it's possible instead `sleep`. You added `(2000 + 1000) * 6` ms to this tests with the `sleep` method

test/recording.ts
```diff
@@ -0,0 +1,57 @@
+import * as path from 'path';
+import { tmpdir } from 'os';
+import { mkdtemp, readdir } from 'fs-extra';
+import { focusChild, focusMain, test, useSpectron } from './helpers/spectron';
+import { setFormDropdown, setFormInput } from './helpers/spectron/forms';
+import { sleep } from './helpers/sleep';
+
+useSpectron({ appArgs: '--nosync' });
```
holiber: '--nosync' is not necessary here, because this test doesn't use authorization

test/recording.ts
blackxored: We're recording 2 sec videos, hence the sleep, does that make sense?

test/recording.ts
blackxored: Got it. Thanks.

PR #1416, opened 2019-02-22
test: recording file detection

blackxored:
Add file detection to recordings to ensure video files are at least created correctly according to file format. WIP as CI is probably not going to like the native bindings/postinstall commands.


blackxored:
Closed as this complicates front-end and CI and backend might be able to test this better.

PR #1417, opened 2019-02-22
test(themes): install overlays

blackxored:
Test that we can install overlay themes. 

Few problems arise:
- Embedded Streamlabs prevents us from accessing/testing search, clicking through install, and more. We currently invoke installs programmatically for a constant overlay theme.
- We can only test that the section has loaded (i.e `webview` is there).

That said, this test the result of installing by:
- Checking the new scene collection was created.
- Checking that scenes were populated.
- Checking that sources were populated.

Which should be enough for a regression test.


blackxored:
Closing in favor of #1446 which has these changes, but goes through the UI to install themes, thanks to @holiber 🙏.

PR #1420, opened 2019-02-25
feat: autostart replay buffer if enabled

blackxored:
Start replay buffer automatically if the setting is enabled.


blackxored:
I agree, further context has been since provided. I'll keep this branch for myself :D

PR #1423, opened 2019-02-25
Fix AlertBox and add facemask support in widget editing

gettinToasty:
Daniel pointed me in the right direction that the facemask settings, which were not previously editable in slobs but were included in the api requests, were getting corrupted. 7750d10 is the commit that will prevent this from happening in the future and will also unbork existing user settings


app/services/widgets/settings/alert-box/index.ts
```diff
@@ -211,6 +212,10 @@ export class AlertBoxService extends WidgetSettingsService<IAlertBoxData> {
       setting.image_href === '/images/gallery/default.gif'
         ? 'http://uploads.twitchalerts.com/image-defaults/1n9bK4w.gif'
         : setting.image_href;
+    const constrainedDuration =
+      Math.floor(setting.alert_duration / 1000) <= 300
+        ? Math.floor(setting.alert_duration / 1000)
+        : 300;
```
holiber: const constrainedDuration = Math.floor(Math.max(300, setting.alert_duration / 1000))

app/components/widgets/AlertBox.vue
```diff
@@ -13,7 +13,7 @@
       <span class="button button--default add-alert-button" @click="toggleAddAlertMenu()">{{ $t('Add Alert') }}</span>
       <div v-if="addAlertMenuOpen" class="add-alert-dropdown">
         <button
-          v-for="type in alertTypes"
+          v-for="type in alertTypes.filter(t => t !== 'facemasks')"
```
blackxored: Would prefer this kind of logic (any logic, really) on a computed prop instead.

app/components/widgets/AlertBox.vue
```diff
@@ -31,7 +31,10 @@
         <span class="left-accordion__title">{{ alertName(alert) }}</span>
       </div>
       <div class="left-accordion__input" v-if="wData.settings[alert]">
-        <validated-form  @input="save()"><toggle-input v-model="wData.settings[alert].enabled" /></validated-form>
+        <validated-form v-if="alert !== 'facemasks'" @input="save()"><toggle-input v-model="wData.settings[alert].enabled" /></validated-form>
```
blackxored: Nitpick: remove brackets.

app/components/widgets/AlertBox.vue
```diff
@@ -31,7 +31,10 @@
...
+        <validated-form v-if="alert === 'facemasks'" @input="handleFacemaskInput()">
```
blackxored: See comment about extracting above. This also impacts the previous line.

app/components/widgets/AlertBox.vue
```diff
@@ -89,7 +92,7 @@
       <v-form-group v-model="selectedVariation.settings.text.color2" :metadata="metadata.secondaryColor" />
     </div>
   </validated-form>
-  <validated-form slot="media-properties" key="media-properties" @input="save()" v-if="selectedVariation">
+  <validated-form slot="media-properties" key="media-properties" @input="handleFacemaskInput()" v-if="selectedVariation">
```
blackxored: Nitpick: remove brackets.

app/components/widgets/AlertBox.vue.ts
```diff
@@ -95,6 +98,8 @@ export default class AlertBox extends WidgetSettings<IAlertBoxData, AlertBoxServ
   editingName: string = null;
   languages: any[] = [];
 
+  facemaskEnabled = this.facemasksService.getEnabledStatus();
```
blackxored: If this doesn't change after init then it's fine as property initializer. Can you please confirm that's the case? Thanks.

app/components/widgets/AlertBox.vue.ts
```diff
@@ -203,4 +208,18 @@ export default class AlertBox extends WidgetSettings<IAlertBoxData, AlertBoxServ
...
+  handleFacemaskInput() {
+    if (this.selectedAlert === 'facemasks') {
+      const { duration } = this.selectedVariation.settings;
+      this.facemasksService
+        .updateFacemaskSettings({
+          duration,
+          enabled: this.facemaskEnabled,
+          device: this.facemasksService.getEnabledDevice(),
+        })
+        .catch(() => this.onFailHandler($t('Something went wrong updating Facemask settings')));
```
blackxored: Do we still see value in reporting to Sentry? 
`Sentry.captureException` prior to invoking `onFailHandler`? Please disregard if the handler already does this.

app/services/widgets/settings/alert-box/index.ts
gettinToasty: i dont want to hold up preview so i'll make a note of that and make this change in a future PR

app/services/facemasks.ts
```diff
@@ -399,9 +420,7 @@ export class FacemasksService extends PersistentStatefulService<IFacemasksServic
     return new Promise((resolve, reject) => {
       const asyncReads = missing.map(uuid => this.readFile(uuid));
       Promise.all(asyncReads)
-        .then(results => {
-          resolve();
-        })
+        .then(results => resolve())
```
blackxored: Nitpick: would prefer `_` as variable name to signal ignoring.

app/services/widgets/settings/alert-box/alert-box-api.ts
```diff
@@ -406,6 +406,38 @@ interface IAlertBoxRaidSettings {
   raid_variations: IAlertBoxVariation[];
 }
 
+// FACEMASKS
+
```
blackxored: Nitpick: remove empty line.

app/services/widgets/settings/alert-box/index.ts
```diff
@@ -211,6 +212,10 @@ export class AlertBoxService extends WidgetSettingsService<IAlertBoxData> {
       setting.image_href === '/images/gallery/default.gif'
         ? 'http://uploads.twitchalerts.com/image-defaults/1n9bK4w.gif'
         : setting.image_href;
+    const constrainedDuration =
```
blackxored: Can we add comments about this logic, or extract consts? Magic numbers aren't always clear on what they meant or why they were chosen. I'd assume `1000` is ms to sec, but even that might be wrong, the `300` is even more obscure.

gettinToasty: 300 is the max duration for alerts, 1000 is ms to sec yea

app/components/widgets/AlertBox.vue
gettinToasty: not sure which brackets you're referring to here

app/components/widgets/AlertBox.vue
gettinToasty: extracting doesnt work here since it's in a for loop

app/components/widgets/AlertBox.vue
gettinToasty: again not sure about which brackets you're referring to... i dont even see any here

app/components/widgets/AlertBox.vue.ts
gettinToasty: it does change

app/components/widgets/AlertBox.vue.ts
gettinToasty: i dont think we report to sentry on this yet, but would be a good thing to add in, i'll make a future PR with that

app/services/widgets/settings/alert-box/index.ts
gettinToasty: 1000 is ms to sec yea, 300 seconds is the maximum duration for alerts. would really prefer to have things like this handled on the backend but this is the api we inherited

PR #1424, opened 2019-02-25
test: output settings

blackxored:
Test output settings.


test/settings/output.ts
```diff
@@ -0,0 +1,88 @@
...
+  832,
+  864,
+  896,
+  928,
+  960,
+  992,
+  1024,
+];
+
+test('simple settings', async t => {
```
gettinToasty: this should really be more descriptive but other than that LGTM

blackxored: Changed to "simple output mode settings"

PR #1436, opened 2019-02-26
fix(settings): extra settings buttons overlay, add margin and spacing

blackxored:
Add extra margin to the top of the actionable items (Restart Streamlabels session and Run
Auto Optimizer) so they don't seem to be part of the previous option.
Add horizontal spacing as well so they look as dialog controls.

EDIT: For CR'ers: I've reformated this template. Bulk of this PR is https://github.com/stream-labs/streamlabs-obs/pull/1436/files#diff-8c7c9c9197f56c0e78171b2c36a6427dR43 and the style underneath, should be pretty trivial.

### DIFF
--- a/app/components/ExtraSettings.vue
+++ b/app/components/ExtraSettings.vue
@@ -1,54 +1,69 @@
 <template>
-<div>
-  <div class="section">
-    <div class="section-content">
-      <p>{{ $t('If you are experiencing weird behavior, you can try deleting your cache directory. This will result in you losing your scene configuration and settings, but can fix some stability issues.')}}</p>
-      <div class="input-container">
-        <a class="link" @click="showCacheDir">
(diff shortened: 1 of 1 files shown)


app/components/ExtraSettings.vue
```diff
@@ -1,54 +1,69 @@
...
 </template>
 
 <script lang="ts" src="./ExtraSettings.vue.ts"></script>
+
+<style lang="less" scoped>
+@import '../styles/mixins';
+.actions {
+  .flex();
+  .flex--space-between();
+  margin-top: 8px;
```
holiber: Use `.margin()` mixin from `mixins.less`

blackxored: Used `.margin-top()`.

PR #1440, opened 2019-02-28
test: video settings

blackxored:
Test video settings:

- Resolution
- Downscale Filter
- FPS type options
- Common FPS values

### DIFF
--- a/test/settings/video.ts
+++ b/test/settings/video.ts
@@ -1,0 +1,51 @@
+import { focusChild, focusMain, test, useSpectron } from '../helpers/spectron';
+import { addTrailingSpace, createOptionsAssertion } from '../helpers/spectron/assertions';
+import { FormMonkey } from '../helpers/form-monkey';
+import { setFormDropdown } from '../helpers/spectron/forms';
+
+useSpectron();
+
+test('Populates video settings', async t => {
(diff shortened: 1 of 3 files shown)


test/settings/video.ts
```diff
@@ -0,0 +1,51 @@
...
+  await assertOptions(
+    '[data-name=FPSType]',
+    'Common FPS Values',
+    ['Common FPS Values', 'Integer FPS Value', 'Fractional FPS Value'].map(addTrailingSpace),
+  );
+
+  await assertOptions(
+    '[data-name=FPSCommon]',
+    '30',
+    ['10', '20', '24 NTSC', '29.97', '30', '48', '59.94', '60'].map(addTrailingSpace),
```
gettinToasty: if these are static arrays i don't see why you don't just add a trailing space into the array itself with an explaining comment? the comment in the helper is a lot less clear when removed from the context

blackxored: We have plenty of these, it looks like every dropdown option coming from OBS, I figure it'd help code duplicating not having to insert (or forget to) the spaces in every test, additionally adding a comment for every instance, instead the comment is on the function itself. Hopefully that makes sense?

PR #1442, opened 2019-02-28
test(settings): bind hotkeys

blackxored:
Test that we can bind a hotkey. I don't seem to be able to actually "trigger" it so can't quite verify that it triggers the action, only that the binding persists.

EDIT: depends on some refactors made in #1440, changed PR base branch for easier review, will update after that one is in.

### DIFF
--- a/app/components/shared/Hotkey.vue
+++ b/app/components/shared/Hotkey.vue
@@ -1,32 +1,31 @@
 <template>
-<div class="hotkey">
-  <div class="Hotkey-description">
-    {{ description }}
+  <div class="hotkey" :data-test-id="description.replace(/\s+/, '_')">
+    <div class="Hotkey-description">{{ description }}</div>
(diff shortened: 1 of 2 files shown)


app/components/shared/Hotkey.vue
```diff
@@ -59,7 +58,8 @@
   .transition();
 }
 
-.bindings-enter, .bindings-leave-to {
+.bindings-enter,
+.bindings-leave-to {
```
gettinToasty: not sure why there's a diff here? i get that it's a formatting thing but it also dirties up the commit, i can't tell if there were any actual changes to this component

blackxored: Yes, sorry about the formatting. 
We've added this: https://github.com/stream-labs/streamlabs-obs/pull/1442/files#diff-452d97a23b7ef7b145ddee5b609dd550R2

PR #1443, opened 2019-02-28
chore(package): update ava to latest stable

blackxored:
Updates AVA to latest stable. Not much on their changelog, but surely getting us out of an RC would help overall.


blackxored:
@holiber Would you be able to confirm that `test-flaky` wouldn't have any issues? Both builds worked fine but wanted to be sure extra safe.

PR #1445, opened 2019-03-01
ci: test timeouts on first run

blackxored:
Add a 1 minute timeout for tests when running on CI (via `test-flaky`), test retries are not
affected by this at this moment. We should consider changing this as well.


andycreeth:
Hmm I think there are some tests that will always take longer than 1 minute.  The test where we create a source of every time generally takes a couple minutes to run in a standard use case.

blackxored:
Updated timeout to 2min, as sources test does indeed take ~1.10 sometimes. I still think this would protect us from / expose hanging tests.

PR #1448, opened 2019-03-01
test: optimized encoder

blackxored:
Test optimized encoder, default profile and game-specific profile options in the Go Live dialog.

In (and only in) CI, we test game-specific profile settings that get populated, namely:

- Output mode changes to Advanced
- Encoder: (x264)
- Bitrate: 2500

### DIFF
--- a/app/components/windows/EditStreamInfo.vue
+++ b/app/components/windows/EditStreamInfo.vue
@@ -70,7 +70,10 @@
           <h-form-group type="text" v-model="startTimeModel.date" :metadata="dateMetadata" />
           <h-form-group type="timer" v-model="startTimeModel.time" :metadata="timeMetadata" />
         </div>
-        <div v-if="selectedProfile">
+        <div
+          v-if="selectedProfile"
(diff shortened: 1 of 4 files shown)


app/components/windows/EditStreamInfo.vue
```diff
@@ -70,7 +70,10 @@
           <h-form-group type="text" v-model="startTimeModel.date" :metadata="dateMetadata" />
           <h-form-group type="timer" v-model="startTimeModel.time" :metadata="timeMetadata" />
         </div>
-        <div v-if="selectedProfile">
+        <div
+          v-if="selectedProfile"
+          :class="{ profile: true, 'profile-default': selectedProfile.game === 'DEFAULT' }"
+        >
```
gettinToasty: cant you use the form to get the data here? i don't like the idea of making classes solely for the purposes of tests

blackxored: Unfortunately not to the best of my ability, due to not having the value anywhere in the DOM, nor being able to get it out via selectors reliably.

PR #1458, opened 2019-03-04
Refactor ObsSystemFontSelector

gettinToasty:
Uses the new property values


app/components/shared/inputs/index.ts
```diff
@@ -55,6 +55,8 @@ export interface IListMetadata<TValueType> extends IInputMetadata {
   internalSearch?: boolean;
   allowCustom?: Function;
   noResult?: string;
+  labelStyle?: any;
```
holiber: don't see any usages of `labelStyle`

app/components/shared/inputs/index.ts
```diff
@@ -55,6 +55,8 @@ export interface IListMetadata<TValueType> extends IInputMetadata {
...
+  optionStyle?: Function;
```
holiber: Using scopes will be a more correct way here
```html
<list-input>
  <span slot="item" :style="myStyle" slot-scope="{ item }">
      {{ item.title }}
  </span>
</list-input>
```

gettinToasty: i'm not sure i understand what you mean here

holiber: Sorry, I mean using `slots`. Metadata should describe the model of the input component, not the appearance. For better understanding the purpose of metadata, try to imagine you're building some external UI for SLOBS. It can be a mobile application. You can fetch metadata of the list-input field via API to get the list of possible options. But CSS styles for these options are useless for you because it's not an HTML application and you don't have CSS

gettinToasty: gotcha-- my question to you here then is that ListInput is already using a slot here
```
    <template slot="option" slot-scope="props">
      <span
        :data-option-value="props.option.value"
        :style="options.optionStyle && options.optionStyle(props.option.value)"
      >{{ props.option.title }}</span>
    </template>
```
so would you prefer something like nested slots?

holiber: You can use this slot. I only think it's better to move `data-option-value` outside of the default template
````html
 <span :data-option-value="props.option.value">
   <template slot="option" slot-scope="props">
     {{ props.option.title }}
   </template>
 </span>
````

PR #1459, opened 2019-03-04
Fix initialization of default values in NumberInput

gettinToasty:


app/components/shared/inputs/NumberInput.vue.ts
```diff
@@ -21,7 +21,8 @@ export default class NumberInput extends BaseInput<number | string, INumberMetad
     input: HTMLInputElement;
   };
 
-  displayValue: number | string = this.value || this.options.min || 0;
+  defaultDisplayValue = this.options.min < 0 || this.options.min == null ? 0 : this.options.min;
+  displayValue: number | string = this.value == null ? this.defaultDisplayValue : this.value;
```
blackxored: Would prefer to have a single type, instead of `number | string` as counter-intuitive as it might look to cast to string on `defaultDisplayValue`.

PR #1466, opened 2019-03-05
Fix textarea jank

gettinToasty:
The original issue (being able to type at the beginning of an input) seems to have resolved itself and was causing some slight jank


gettinToasty:
@andycreeth no idea why that happened... but fixed now

PR #1472, opened 2019-03-07
Remove references to underscore

gettinToasty:
Properly treeshake lodash imports and remove deprecated usages


gettinToasty:
@andycreeth yup; we'll have to be vigilant about that in our CRs

PR #1479, opened 2019-03-08
ExternalAPI - Fallback

holiber:
If some method is not exposed in exernalApi call it from internalApi


app/services/api/external-api/scenes/scene.ts
```diff
@@ -133,9 +135,9 @@ export class Scene {
   /**
    * removes all nodes from the scene
    */
-  clear(): void {
-    return this.scene.clear();
-  }
+  // clear(): void {
+  //   return this.scene.clear();
+  // }
```
andycreeth: What is this commented out for?

holiber: I commented it out to check that this method will be called from internal API and tests will be passed. There is no need for this now.

PR #1481, opened 2019-03-08
Color Audit Pt 1

gettinToasty:
Implementation of scoped variables for all of our colors. Dark mode is not yet working in the twitch tags input but apart from some slight intentional differences in the name of color unification this is going well so far


app/app.g.less
```diff
@@ -327,14 +327,14 @@ select {
...
+  background: var(--section);
   border-radius: 0 3px 3px 0;
 }
 
 ::-webkit-scrollbar-thumb {
-  background-color: @white;
-  border: 2px solid @day-border;
-  border-right: 1px solid @day-border;
+  background-color: var(--scrollbar);
+  border: 2px solid var(--border);
```
blackxored: I think this one should be called `border-style`

app/app.g.less
```diff
@@ -368,7 +368,8 @@ select {
 }
 
 .v--modal {
-  background-color: @day-bg;
+  background-color: var(--background);
+  box-shadow: 0 0 0 var(--shadow);
```
blackxored: Does this do anything? Is it animated later?

gettinToasty: not sure, just trying to change it the least from staging

app/app.g.less
gettinToasty: can you explain your thoughts here? why border and not suffixing any of the other colors?

PR #1484, opened 2019-03-11
Remove cert modules

andycreeth:


Aidenqing:
When I was packaging the source code, there is a problem that  is 'Cannot find certificate Streamlabs (General Workings, Inc.), all certs:'. Could you please tell me how to solve it?

PR #1485, opened 2019-03-11
Extend FB stream limit notification

gettinToasty:
FB changed their stream limit time from 4 to 8 hours so we should adjust our messaging to not confuse users

### DIFF
--- a/app/services/streaming/streaming.ts
+++ b/app/services/streaming/streaming.ts
@@ -135,7 +135,7 @@ export class StreamingService extends StatefulService<IStreamingServiceState>
 
   private finishStartStreaming() {
     const shouldConfirm = this.settingsService.state.General.WarnBeforeStartingStream;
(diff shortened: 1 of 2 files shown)


app/services/streaming/streaming.ts
```diff
@@ -135,7 +135,7 @@ export class StreamingService extends StatefulService<IStreamingServiceState>
 
   private finishStartStreaming() {
     const shouldConfirm = this.settingsService.state.General.WarnBeforeStartingStream;
-    const confirmText = 'Are you sure you want to start streaming?';
+    const confirmText = $t('Are you sure you want to start streaming?');
```
blackxored: It seems like we didn't update the EN language file on this one.

gettinToasty: that's true for like 70% of strings in our repo lol i think that should be a separate task

PR #1498, opened 2019-03-14
move chat to browserview

andycreeth:


app/services/chat.ts
```diff
@@ -0,0 +1,190 @@
...
+    this.navigateToChat();
+  }
+
+  mountChat(electronWindowId: number) {
+    if (!this.chatView) return;
+
+    const win = electron.remote.BrowserWindow.fromId(electronWindowId);
+
+    // This method was added in our fork
+    (win as any).addBrowserView(this.chatView);
```
blackxored: Nitpick, prefer:
`// @ts-ignore: this method was added in our fork`
so we don't have to remove (or read) the typecast if we ever fix typings.

app/services/chat.ts
```diff
@@ -0,0 +1,190 @@
...
+    this.bindWindowListener();
+    this.bindDomReadyListener();
+
+    this.customizationService.settingsChanged.subscribe(changed => {
+      this.handleSettingsChanged(changed);
+    });
+  }
+
+  private deinitChat() {
+    (this.chatView as any).destroy();
```
blackxored: I assume the cast is because it can be `null` (not initialized)? If so, can we conditionally call `destroy`?

app/services/chat.ts
```diff
@@ -0,0 +1,190 @@
...
+        this.chatView.webContents.loadURL(chatUrl);
+      });
+    } else {
+      const chatUrl = await service.getChatUrl(nightMode);
+      this.chatView.webContents.loadURL(chatUrl);
+    }
+  }
+
+  private bindWindowListener() {
+    this.chatView.webContents.on('new-window', e => {
```
blackxored: Would prefer this to be `event` or `evt`. The single char var makes the following line harder to parse, at least for me.

app/services/chat.ts
```diff
@@ -0,0 +1,190 @@
...
+
+            setTimeout(loadLazyEmotes, 1000);
+          }
+
+          loadLazyEmotes();
+        `,
+          true,
+        );
+      }
+      if (settings.enableFFZEmotes && this.userService.platform.type === 'twitch') {
```
blackxored: Nitpick: Add a line break, so we separate from the bunch of inline JS above.

app/services/chat.ts
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  private handleSettingsChanged(changed: Partial<ICustomizationSettings>) {
+    if (!this.chatView) return;
+
+    if (changed.chatZoomFactor) {
+      this.chatView.webContents.setZoomFactor(changed.chatZoomFactor);
+    }
+
+    if (changed.enableBTTVEmotes != null) {
```
blackxored: Would prefer if we merge these two conditionals.

app/services/chat.ts
```diff
@@ -0,0 +1,190 @@
...
+
+export class ChatService extends Service {
+  @Inject() userService: UserService;
+  @Inject() customizationService: CustomizationService;
+  @Inject() windowsService: WindowsService;
+
+  private chatView: Electron.BrowserView;
+
+  init() {
+    this.userService.userLogin.subscribe(() => this.initChat());
```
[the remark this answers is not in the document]
andycreeth: What specifically do you think could be refactored into shared code?  This is a single statement wiring up a function to an event.  I don't really see any boilerplate or commonality here.

app/services/chat.ts
andycreeth: Did not realize this was a thing.

app/services/chat.ts
andycreeth: We're not using strict null checks, so that cast wouldn't be needed.  The reason for the cast is because electron's typings are incorrect for this API (not used by many people) so this method is not in the typings lol.

app/services/chat.ts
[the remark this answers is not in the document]
blackxored: On login -> X
On logout -> Y
Do X

Only X and Y are different here and it's the same logic. Just the cleanup and init functions change.

PR #1502, opened 2019-03-14
Add new loader from Beaker

gettinToasty:


andycreeth:
We need to chat about the dependencies this library brings in before we can think about merging this.

andycreeth:
Per our discussion, let's try moving the animation delay logic into the customization service using a store watcher.  Otherwise this is great and good to go.

PR #1510, opened 2019-03-18
Fix ListInput reselecting issue

gettinToasty:
Cosmetic issue existed when a user tried to reselect the currently selected input, displaying no selection

### DIFF
--- a/app/components/shared/inputs/ListInput.vue.ts
+++ b/app/components/shared/inputs/ListInput.vue.ts
@@ -20,10 +20,10 @@ export default class ListInput extends BaseInput<string, IListMetadata<string>> 
   readonly placeholder: string;
 
   onInputHandler(option: IListOption<string>) {
-    if (option) {
-      this.emitInput(option.value);
-      this.$nextTick();
-    }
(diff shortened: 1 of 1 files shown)


app/components/shared/inputs/ListInput.vue.ts
```diff
@@ -20,10 +20,10 @@ export default class ListInput extends BaseInput<string, IListMetadata<string>>
...
 
   onInputHandler(option: IListOption<string>) {
-    if (option) {
-      this.emitInput(option.value);
-      this.$nextTick();
-    }
+    // Fixes a render issue when reselecting the same option as currently selected
+    const val = option ? option.value : this.value;
+    this.emitInput(val);
+    this.$nextTick();
```
andycreeth: What is `$nextTick()` doing here?  `$nextTick` works by either passing it a callback function or waiting on the returned promise, neither of which are doing here.  I don't think this line of code does anything.

andycreeth: Also I recognize this line was here before your change, but we should probably just remove it.

PR #1512, opened 2019-03-18
fix dshow initialization at startup

andycreeth:


app/services/app/app.ts
```diff
@@ -152,7 +153,6 @@ export class AppService extends StatefulService<IAppState> {
 
     this.crashReporterService.endStartup();
 
-    this.FINISH_LOADING();
```
andycreeth: While debugging this issue, I discovered that we were removing the loading screen too early due to not consistently using the decorator.

PR #1517, opened 2019-03-19
test(streaming): optimized encoder redux

blackxored:
Awaits up to 10sec for Twitch responses for available games.

### DIFF
--- a/test/streaming/live-dialog.ts
+++ b/test/streaming/live-dialog.ts
@@ -1,0 +1,78 @@
+import { focusChild, focusMain, test, useSpectron } from '../helpers/spectron';
+import { getFormInput } from '../helpers/spectron/forms';
+import { sleep } from '../helpers/sleep';
+import { logIn } from '../helpers/spectron/user';
+
+useSpectron({ appArgs: '--nosync' });
+
+test('Shows optimized encoder for specific games', async t => {
+  const { app } = t.context;
+  await logIn(t);
+
+  await focusMain(t);
+  await app.client.waitForExist('button=Go Live');
+  await app.client.click('button=Go Live');
+  await focusChild(t);
+  await app.client.waitForExist('label=Title');
+
+  // This part also tests game search in some way
+  app.client.setValue('.multiselect__input', 'starcraft');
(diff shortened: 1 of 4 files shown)


test/streaming/live-dialog.ts
```diff
@@ -0,0 +1,78 @@
...
+   * "Use optimized encoder settings for Overwatch".
+   * Unfortunately, something seems to be messing up webdriver and it can't find
+   * 'label=Use optimized encoder settings for Overwatch'
+   * so we workaround by adding a CSS class for when it's using the default profile
+   */
+  const owSelector = '.multiselect__element=Overwatch';
+  await app.client.setValue('.multiselect__input', 'Overwa');
+  await app.client.waitForExist(owSelector, 10000);
+  await app.client.click(owSelector);
+  await sleep(1000);
```
holiber: that may be flaky, use waitForVisible/waitForExist

test/streaming/live-dialog.ts
```diff
@@ -0,0 +1,78 @@
...
+    await focusMain(t);
+    await app.client.waitForExist('button=End Stream');
+    await app.client.click('.top-nav .icon-settings');
+
+    await focusChild(t);
+    await app.client.click('li=Output');
+
+    t.is('Advanced', await app.client.getValue('[data-name=Mode] input'));
+    t.is('Software (x264)', await app.client.getValue('[data-name=Encoder] input'));
+    t.is('2500', await getFormInput(t, 'Bitrate'));
```
holiber: `FormMonkey` provides a unified way to fill out forms. Also, it prevents some issues with flaky tests

test/streaming/live-dialog.ts
blackxored: There doesn't seem to be a selector I can tap in (we're looking for input's value as our wait condition) so I did something similar to what FormMonkey seems to do with dropdowns.

test/streaming/live-dialog.ts
blackxored: IMHO using FormMonkey's `read` is more complex for this test and scenario, even if it works.

PR #1518, opened 2019-03-19
test: re-enable themes test

blackxored:
Re-enables "Installing a theme" test, after fixing by increasing timeout for theme install which seems to take noticeably-longer on CI.


holiber:
Are you sure the failure it's related to installing overlay time and not to themes searching?

blackxored:
Only theme install failed when I RDP'ed into it. You could see "Installing" progress bar at about one third done when the test failed.

PR #1522, opened 2019-03-19
get rid of confusing previewEnabled customization setting

andycreeth:
This setting name didn't make any sense the way we now use this setting.  Hopefully this new setup should be a lot clearer, and fixes an issue where chat was disabled in performance mode.

### DIFF
--- a/app/components/LiveDock.vue.ts
+++ b/app/components/LiveDock.vue.ts
@@ -188,11 +188,8 @@ export default class LiveDock extends Vue {
     this.chatService.refreshChat();
   }
 
-  get hideChat() {
-    // previewEnabled is poorly named.  It really just means stuff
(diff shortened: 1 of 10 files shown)


app/components/LiveDock.vue.ts
```diff
@@ -188,11 +188,8 @@ export default class LiveDock extends Vue {
     this.chatService.refreshChat();
   }
 
-  get hideChat() {
-    // previewEnabled is poorly named.  It really just means stuff
-    // we disable while resizing.  It should not be in the customization
-    // service.  This is lazy.
```
andycreeth: I thought of creating a new service, but decided we should keep it in the customization service for now.  It's not technically a user setting, but it's related to resizing, which is handled by the customization service.  So it's not the worst place for now.

PR #1525, opened 2019-03-20
use separate sessions for platform auth windows

andycreeth:
Have the user log in to the streaming platform in a separate partition, and then use that same partition for chat.  This has a couple benefits:
- It makes it very easy to ensure we aren't accidentally leaking cookies when the user logs out, as the id of the partition is deleted and can't be re-generated.
- It will hopefully work around an issue we've been seeing where the default session will get corrupted and stop updating cookies eventually.  This appears to be a chromium bug.

### DIFF
--- a/app/services/chat.ts
+++ b/app/services/chat.ts
@@ -59,8 +59,11 @@ export class ChatService extends Service {
   private initChat() {
     if (this.chatView) return;
 
+    const partition = this.userService.state.auth.parition;
+
     this.chatView = new electron.remote.BrowserView({
       webPreferences: {
+        partition,
         nodeIntegration: false,
       },
     });
--- a/app/services/chat.ts
+++ b/app/services/chat.ts
@@ -59,7 +59,7 @@ export class ChatService extends Service {
   private initChat() {
     if (this.chatView) return;
 
-    const partition = this.userService.state.auth.parition;
+    const partition = this.userService.state.auth.partition;
 
     this.chatView = new electron.remote.BrowserView({
       webPreferences: {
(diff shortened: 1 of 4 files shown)


app/services/chat.ts
```diff
@@ -59,8 +59,11 @@ export class ChatService extends Service {
   private initChat() {
     if (this.chatView) return;
 
+    const partition = this.userService.state.auth.parition;
```
blackxored: This seems like a typo.

app/services/platforms/index.ts
```diff
@@ -116,6 +116,12 @@ export interface IPlatformAuth {
     id: string;
     channelId?: string;
   };
+
+  /**
+   * Session partition used to separate cookies associated
+   * with this user login.
+   */
+  parition?: string;
```
blackxored: See previous comment.

app/services/user.ts
```diff
@@ -261,9 +270,17 @@ export class UserService extends PersistentStatefulService<IUserServiceState> {
     await this.chatbotApiService.Base.logOut();
     // Navigate away from disabled tabs on logout
     this.navigationService.navigate('Studio');
+
+    let session = electron.remote.session.defaultSession;
```
blackxored: Nitpick: would prefer session to be `const` and assigned as a ternary.

app/services/chat.ts
andycreeth: ty

andycreeth:
Thanks, good comments.  Addressed.

PR #1528, opened 2019-03-20
properly flush cookies

andycreeth:


app/services/user.ts
```diff
@@ -132,8 +132,15 @@ export class UserService extends PersistentStatefulService<IUserServiceState> {
    */
   flushUserSession() {
```
holiber: Suggest to mark this method as async now via adding a returning type
```flushUserSession(): Promise<void>```

PR #1530, opened 2019-03-20
fix: do not launch slobs after updater uac blocked (WIP)

summeroff:
Add promise to get stderr from updater spawner. 
If it return message "Access is denied." we can use it as a sing of canceled UAC request. 
In this case function return true which tells SLOBS to close.

### DIFF
--- a/updater/bootstrap.js
+++ b/updater/bootstrap.js
@@ -306,22 +306,33 @@ async function entry(info) {
 
     log.info('updater process ' + `pid ${update_spawned.pid}`);
 
-    //make promises for app exit , error , data and some timeout
-    const primiseExit = new Promise(resolve => {
+    //make promises for app exit , error , dataerror
+    const promiseDataError = new Promise(resolve => {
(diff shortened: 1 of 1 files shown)


updater/bootstrap.js
```diff
@@ -306,22 +306,35 @@ async function entry(info) {
...
     });
 
-    const primiseError = new Promise(resolve => {
+    const promiseError = new Promise(resolve => {
         update_spawned.on('error', resolve);
     });
 
     //wait for something to happen
-    const promise = await Promise.race([primiseError, primiseExit]);
+    const promise = await Promise.race([promiseError, promiseExit, promiseDataError]);
```
holiber: suggesting simplifing:
```
const returnCode = await new Promise(resolve => {
  update_spawned.stderr.on('data', resolve);
  update_spawned.on('exit', resolve);
  update_spawned.on('error', resolve);
})
```
And remove promises declarations above

PR #1531, opened 2019-03-20
Fix notification settings

holiber:
The original issue is when you click `Restore Defaults` in the Notification Settings the sliders don't change values.

The actual issue is that SliderInput and ObsSliderInput don't react on `value` property change

### DIFF
--- a/app/components/shared/inputs/BaseFormGroup.ts
+++ b/app/components/shared/inputs/BaseFormGroup.ts
@@ -1,4 +1,3 @@
-import { Prop } from 'vue-property-decorator';
 import { ErrorField } from 'vee-validate';
 import { EInputType, IInputMetadata } from './index';
 import { BaseInput } from './BaseInput';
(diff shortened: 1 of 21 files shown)


app/components/shared/inputs/BaseFormGroup.ts
```diff
@@ -7,17 +6,10 @@ import { BaseInput } from './BaseInput';
...
-
-  @Prop()
-  readonly metadata: IInputMetadata;
-
-  @Prop()
-  readonly title: string;
+  abstract readonly type: EInputType;
+  abstract readonly value: undefined;
+  abstract readonly metadata: IInputMetadata;
+  abstract readonly title: string;
```
holiber: I realized that Vue doesn't like when you use `watchers` for fields that use  `@Prop` decorator in the base class.

app/components/shared/inputs/CodeInput.vue.ts
```diff
@@ -14,6 +14,9 @@ export default class CodeInput extends BaseInput<string, IInputMetadata> {
   @Prop({ default: '' })
   readonly value: string;
 
+  @Prop()
+  readonly title: string;
```
holiber: I added the `title` prop in each Input component for backward capability with old code that added this field in the base class

app/components/shared/inputs/SliderInput.vue.ts
```diff
@@ -15,6 +15,12 @@ export default class SliderInput extends BaseInput<number, ISliderMetadata> {
   @Inject() customizationService: CustomizationService;
 
   @Prop() readonly value: number;
+  @Watch('value')
+  private syncLocalValue(newVal: number) {
+    this.localValue = newVal;
+  }
```
andycreeth: Is this a behavioral change?  I guess I don't fully understand why this is needed, or if it is needed, how it was working before.

holiber: It's a behavioral change. The 2-way binding didn't work at all after some of refactoring commits. You could render an initial value or manually change a slider position

PR #1537, opened 2019-03-21
mutation safety

andycreeth:
# Introduction

Execute all mutations with a restricted `this` context in the development environment.  The restricted context only allows accessing `this.state`.

# FAQ

## Why restrict what mutations can do?

Mutations shouldn't be treated like normal functions.  They are very specific functions that are designed to take a set of arguments and modify their local slice of the state object.

Philosophically, mutations should:
- Do as little as possible to modify state, because they are executed once per window
- Not have side effects, because those side effects will happen N times with N windows
- Not rely on context that is inconsistent across windows (class instance attributes)
- Definitely not make any OBS IPC calls
- Not access a slice of the store that they are not actively managing
- Not have heavy lifting of business logic

By restricting the `this` context, we are forcing mutations to only rely on their arguments and `state`.  While they are not pure functions like Redux reducers, this gets us closer to that ideal.  If a mutation needs access to something else, it is almost definitely something that shouldn't be happening inside a mutation to begin with.

## Is there an escape hatch?

Passing `{ unsafe: true }` will run the mutation with a full `this` context.  We have 1 mutation that uses this currently, because it was proving difficult to rewrite safely.  Please don't use this escape hatch without discussing it in slack first.

## Why only in the dev environment?

Proxies in JavaScript can be expensive.  Given the primary purpose here is to catch poor development practices, it seems silly to incur this performance hit in production.  It also makes this PR much safer to merge, since it has essentially zero effect in production.


andycreeth:
Ok judging by test failures, I definitely missed a bunch of unsafe mutations.  Will look into those.

andycreeth:
I just want to add that I think a better long term solution would be to make a pure functional wrapper over mutations that would require us to rewrite all mutations as a pure function.  Then we can enforce this stuff at compile time rather than a runtime proxy.

PR #1542, opened 2019-03-22
Fix FB stream info overwrite

gettinToasty:
We were caching the live video id (used to set stream info) and never actually clearing it


app/services/platforms/facebook.ts
```diff
@@ -201,6 +201,7 @@ export class FacebookService extends StatefulService<IFacebookServiceState>
           this.SET_LIVE_VIDEO_ID(info.id);
           this.SET_STREAM_URL(info.stream_url);
         }
+        this.SET_LIVE_VIDEO_ID(null);
```
blackxored: We're unconditionally setting it to `null` (despite whether the above is truthy), is this intended?

gettinToasty: whoops no it isnt!

PR #1543, opened 2019-03-22
feat(settings): display tabs for subcategories in Advanced output

blackxored:
Display subcategories as tabs in Output settings when set to Advanced mode.

Fully functional, but pending backend change related to `ECategoryTypes` so we can safely refactor to check for that as opposed to the somewhat hardcoded paths existing in this branch. 

I'm also working in general refactor and cleanup of what we have that won't be affected by the above.

CR'ers: I've split standalone properties (such as Output mode) as they don't belong inside the tabs content. Audio sends a subcategory for every track and we do not want to display 6 audio tabs, there's additional code to somewhat group them together which complicates rendering (or making this component as generic as it could be). Unless we do want to make backend changes/refactoring, IMHO both are valid workarounds.

Screenshot:

[Screenshot](https://gyazo.com/d9ac26fd25f94413067601bf7e98cdd1)

Output settings tests would need to be updated to include tab clicks.

EDIT: we only have Simple Mode Output tests, so this shouldn't fail builds, but we should write these after completion (it'd probably require CI debugging).


app/components/obs/inputs/GenericFormTabbed.vue.ts
```diff
@@ -0,0 +1,59 @@
...
+      .filter(val => !val.nameSubCategory.startsWith('Audio - Track'))
+      .map(toTab);
+
+    /*
+     * Insert "Audio" tab as the 3rd tab, since it looks more intuitive this way.
+     * TODO: I'd avoid mutation, but since we're not sold on that idea yet, we rather do it like
+     *  this instead of adding complexity or introducing deps.
+     */
+
+    const audioTab: ITab = { name: 'Audio', value: 'Audio' };
```
holiber: I have a concern about you name this component as `GeneriggFormTabbed` but with this `Audio` tab handler and `currentTab: string = 'Streaming';` it looks like a specific component for streaming settings.

I see 2 options here:
1) Rename component to `StreamingSettingsForm`
2) Leave it as `GenericFormTabbed` but and make it reusable. (Move all special handlers for audio and default tab in SettingsService)

blackxored: Went with #2.

PR #1549, opened 2019-03-25
Add transition wildcards

gettinToasty:
As per the survey result we'll prefer the scene being transitioned to to win in conflicting transition cases

### DIFF
--- a/app/components/ConnectionSettings.vue.ts
+++ b/app/components/ConnectionSettings.vue.ts
@@ -4,6 +4,7 @@ import { Inject } from 'util/injector';
 import { TransitionsService } from 'services/transitions';
 import { ScenesService } from 'services/scenes';
 import VFormGroup from 'components/shared/inputs/VFormGroup.vue';
+import { $t } from 'services/i18n';
 
 @Component({
   components: { VFormGroup },
@@ -43,7 +44,12 @@ export default class SceneTransitions extends Vue {
   }
 
   get sceneOptions() {
-    return this.scenesService.scenes.map(scene => ({ title: scene.name, value: scene.id }));
+    const sceneOptions = this.scenesService.scenes.map(scene => ({
+      title: scene.name,
+      value: scene.id,
+    }));
+    sceneOptions.unshift({ title: $t('All'), value: 'ALL' });
+    return sceneOptions;
   }
 
   get transitionOptions() {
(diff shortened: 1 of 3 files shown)


app/components/windows/SceneTransitions.vue.ts
```diff
@@ -144,6 +144,8 @@ export default class SceneTransitions extends Vue {
   }
 
   getSceneName(id: string) {
+    if (id === 'ALL') return $t('All');
```
holiber: `getSceneName(id: string |  'ALL')`
I'm also sugesting use `*` instead `ALL`

gettinToasty: hmm why do you suggest that?

gettinToasty: also not sure why we need to type differently here since id will never not be a string

holiber: Using `*` is just a common solution to specify a wildcard in many libs.  It also allows to add more flexible filters like a partial match in the future: `my-scene-*` <-- match all scenes names starting with `my-scene-`.

`getSomething(id: string | 'ALL') ` <-- will give you better intelliSence for understanding that method uses some reserved keywords
image

gettinToasty: right now we aren't actually using any matching for this though, nor do i think this is a use-case where we would have partial matching. the user isn't inputting IDs for us to match, we're programmatically generating them without any distinguishing features. if we do allow user-generated IDs in the future i still don't think we'll be able to usefully implement partial matching since we'll have no way to enforce what users submit as an ID while still allowing for useful flexibility

app/components/ConnectionSettings.vue.ts
```diff
@@ -43,7 +44,12 @@ export default class SceneTransitions extends Vue {
   }
 
   get sceneOptions() {
-    return this.scenesService.scenes.map(scene => ({ title: scene.name, value: scene.id }));
+    const sceneOptions = this.scenesService.scenes.map(scene => ({
+      title: scene.name,
+      value: scene.id,
+    }));
+    sceneOptions.unshift({ title: $t('All'), value: 'ALL' });
+    return sceneOptions;
```
andycreeth: Super nitpicky, but a cleaner way to write this might be:
```TypeScript
return [
  { title: $t('All'), value: 'ALL' },
  ...this.scenesService.scenes.map(scene => ({
    title: scene.name,
    value: scene.id,
  }))
];
```
The primary advantage is that you're composing the object you want to return, rather that using a temporary variable, mutating it, then returning it.

PR #1555, opened 2019-03-27
Fix mutation buffering

holiber:
After investigating Sentry errors from Projector window I found that we don't commit mutations coming from the child window because `mutationsBuffering` prevent this. Here is a PR with a fix

### DIFF
--- a/app/services/api/internal-api-client.ts
+++ b/app/services/api/internal-api-client.ts
@@ -1,6 +1,6 @@
 import electron from 'electron';
 import { Observable, Subject } from 'rxjs';
-import { IJsonRpcEvent, IJsonRpcResponse, JsonrpcService } from 'services/api/jsonrpc';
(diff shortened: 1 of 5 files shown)


app/services/api/internal-api-client.ts
```diff
@@ -73,7 +75,13 @@ export class InternalApiClient {
...
 
           const result = response.result;
-          response.mutations.forEach(mutation => commitMutation(mutation));
+          const mutations = response.mutations;
+
+          // commit all mutations caused by the api-request now
+          mutations.forEach(mutation => commitMutation(mutation));
+          // we'll still receive already committed mutations from async IPC event
+          // mark them as ignored
+          this.skippedMutationsCount = mutations.length;
```
andycreeth: Just thinking about this.  In an n-window asynchronous environment, blindly skipping the N mutations and assuming they are the mutations you already committed here seems very brittle and dangerous.  This seems like a recipe for weird hard to track down bugs IMO.  I'm not really comfortable with this.  Trying to think of a better more robust way to handle this.

PR #1562, opened 2019-03-27
Add better error reporting when we fail to deload the application state

andycreeth:


andycreeth:
Going into staging so we can reap the benefit of this additional reporting earlier.  Given it's simply changing a log statement, it's extremely low risk.

PR #1563, opened 2019-03-28
Use realpath for image sources

soatok:
I'm building an app that lets users swap out image sources from a simple point-and-click UI.

It's breaking on GIF images because of the special handling for them, [in a weird way](https://github.com/soatok/fursona-sticker-switcher#caveats-with-animated-gifs-and-streamlabs-obs). Using the realpath of the file to get the extension should enable the correct behavior.

> If you append `.gif` to the Image Source file path, you can use GIFs (but
only GIFs) and they will animate. If you leave the image source without an
extension, it works with all image types, but GIFs will not be animated.
> 
> In the future, I'll devise a workaround (even if it means sending a pull
request to Streamlabs OBS) to make this usability wart go away.

The cause for this behavior was that GIFs are handled by ffmpeg and other images are not.

### DIFF
--- a/app/services/sources/sources.ts
+++ b/app/services/sources/sources.ts
@@ -219,7 +219,7 @@ export class SourcesService extends StatefulService<ISourcesState> implements IS
   }
 
   addFile(path: string): Source {
-    let realpath = fs.realpathSync(path);
+    const realpath = fs.realpathSync(path);
     const SUPPORTED_EXT = {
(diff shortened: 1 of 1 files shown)


andycreeth:
Hey @soatok thanks for your submission.  I'm just trying to understand the issue here.  Does your application work by creating symbolic links?

soatok:
Generally, yes. On Windows it can't create symlinks unless you run it as administrator, so it will unlink/copy instead.

PR #1569, opened 2019-03-29
Fix dropdown menu night mode

gettinToasty:
There are bound to be more little regressions like this


app/components/shared/DropdownMenu.vue
```diff
@@ -86,6 +86,7 @@
   border: 0 !important;
   padding: 8px !important;
   color: inherit !important;
+  background-color: var(--background) !important;
```
holiber: I assume you can get rid of all `!important` here if you specify selector `.popper.dropdown-menu`

PR #1571, opened 2019-03-29
Advanced audio settings tests

holiber:
Here is also some little refactoring for `FormMonkey`


test/helpers/form-monkey.ts
```diff
@@ -19,19 +18,18 @@ interface IUIInput {
  * helper for simulating user input into SLOBS forms
  */
 export class FormMonkey {
-  private client: SpectronClient;
-
-  constructor(private t: TExecutionContext, private showLogs = false) {
-    this.client = t.context.app.client;
+  constructor(
+    private t: TExecutionContext,
+    private formSelector = 'body',
```
blackxored: Can we make `formSelector` the last argument? I agree that it could be more intuitive this way, but it is a breaking change for tests, which wouldn't be if the took that last.

test/settings/advanced-audio.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  // check settings are still updated after window close
+  await closeWindow(t);
+  await focusMain(t);
+  await clickAdvancedAudio(t);
+  await focusChild(t);
+  t.true(await desktopAudioForm.includes(updatedSettings));
+  t.true(await micAuxForm.includes(updatedSettings));
+
+  t.pass();
```
blackxored: Do we need this?

test/helpers/form-monkey.ts
holiber: This PR already introduced a breaking change for tests. But there are still no so many places affected by this breaking change so I fixed all of them.

PR #1575, opened 2019-04-01
Update translation strings

holiber:


app/i18n/en-US/widget-alertbox.json
```diff
@@ -0,0 +1,73 @@
...
+  "Font Size": "Font Size",
+  "Font": "Font",
+  "Min. Amount to Show": "Min. Amount to Show",
+  "Variation Amount": "Variation Amount",
+  "Min. Amount to Trigger Alert": "Min. Amount to Trigger Alert",
+  "Min. Amount to Show in Recent Events": "Min. Amount to Show in Recent Events",
+  "Min. Amount to Read": "Min. Amount to Read",
+  "Show Animation'), filter: 'in": "Show Animation'), filter: 'in",
+  "Hide Animation'), filter: 'out": "Hide Animation'), filter: 'out",
+  "Text Animation'), filter: 'text": "Text Animation'), filter: 'text",
```
gettinToasty: these three seem to have picked up some non-string portions

app/i18n/en-US/custom-loader.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Initiating doggo petting engine...": "Initiating doggo petting engine...",
+  "Compiling food pic library...": "Compiling food pic library...",
+  "Callibrating RNG blame threshold...": "Callibrating RNG blame threshold...",
+  "Dropping fire mixtape...": "Dropping fire mixtape...",
+  "Reticulating emoji splines...": "Reticulating emoji splines...",
+  "Waking up 24/7 support team...": "Waking up 24/7 support team...",
+  "Asking the stream gods for perfect settings...": "Asking the stream gods for perfect settings...",
+  "Powering up the clouds...": "Powering up the clouds...",
+  "Constructing additional widgets...": "Constructing additional widgets..."
```
gettinToasty: i don't think these jokes/references will translate perfectly to other languages, so i wonder if we can somehow add a note to these to add some localized references

vit1977: The jokes are definitely funny. But how about people who are stuck with them forever. I can't run Streamlabs OBS at all because it frozen on these funny comments. I even tried to uninstall the program and install again. Still doesn't work.

Untitled-1

PR #1579, opened 2019-04-02
Remove backtrace and setup Sentry for the main process

RodrigoHolztrattner:
Remove backtrace and add Sentry on its place for crash reporting on the main process.


main.js
```diff
@@ -119,19 +119,12 @@ function startApp() {
...
-    bt.initialize({
-      disableGlobalHandler: true,
-      endpoint: 'https://streamlabs.sp.backtrace.io:6098',
-      token: 'e3f92ff3be69381afe2718f94c56da4644567935cc52dec601cf82b3f52a06ce',
-      attributes: {
-        version: pjson.version,
-        processType: 'main'
-      }
-    });
+    Sentry.init({ dsn: 'https://6971fa187bb64f58ab29ac514aa0eb3d@sentry.io/251674' });
```
andycreeth: Which project on sentry does this DSN point to?

main.js
```diff
@@ -119,19 +119,12 @@ function startApp() {
   }
 
   function handleUnhandledException(err) {
-    bt.report(err, {}, handleFinishedReport);
+    Sentry.captureException(err);
```
andycreeth: We should at the very least send along the version number.  Does sentry's node library automatically pull that from the `package.json` or do we need to send it along?

main.js
```diff
@@ -119,19 +119,12 @@ function startApp() {
...
+    handleFinishedReport();
```
andycreeth: We also need to change the dialog that pops up.  The copy is really bad.  I can do that in a separate PR.

main.js
RodrigoHolztrattner: streamlabs-obs project

main.js
RodrigoHolztrattner: I believe it should be added manually, this is a sample report from these modifications https://sentry.io/organizations/streamlabs-obs/issues/966567397/events/51437036986/?project=251674&statsPeriod=14d&utc=false

RodrigoHolztrattner: Looking into this

RodrigoHolztrattner: I was unable to test but I believe my modifications should populate the version and release tags (Sentry is probably filtering me out)

RodrigoHolztrattner: Since I don't know what is accessible from the `main.js` file I wasn't sure what extra info to send, but it should be easy to add if necessary. Just ping me with what need to be sent.

main.js
andycreeth: This was done

andycreeth:
Just to make sure everything is good, can you induce an exception and post the sentry link?  I just want to make sure everything is reported consistently with our current javascript errors in that project.  Thanks!

RodrigoHolztrattner:
I'm trying to do it but for some reason it sent 4 reports yesterday but now I can't get new ones. Maybe it is conflicting with the other instances of Sentry?
I will try using `raven` tomorrow, see if that solves the issue.

RodrigoHolztrattner:
Sorry the delay, for some reason the default Sentry library wasn't working correctly but I switched to Raven and now it's ok.
A sample report: https://sentry.io/organizations/streamlabs-obs/issues/972215227/events/51670463988/?project=251674&statsPeriod=14d&utc=false
Also a list of the recent reports that I forced: https://sentry.io/organizations/streamlabs-obs/events/?project=251674&query=user.ip%3A216.71.212.26&statsPeriod=14d&utc=false

main.js
```diff
@@ -125,7 +125,7 @@ function startApp() {
     );
   }
 
-  const bt = require('backtrace-node');
+  var Raven = require('raven');
```
andycreeth: Use `const` instead of `var`.

main.js
```diff
@@ -139,36 +139,25 @@ function startApp() {
...
-    bt.initialize({
-      disableGlobalHandler: true,
-      endpoint: 'https://streamlabs.sp.backtrace.io:6098',
-      token: 'e3f92ff3be69381afe2718f94c56da4644567935cc52dec601cf82b3f52a06ce',
-      attributes: {
-        version: pjson.version,
-        processType: 'main'
-      }
-    });
+  Raven.config('https://6971fa187bb64f58ab29ac514aa0eb3d@sentry.io/251674', {
```
andycreeth: Unless I'm missing something, it looks like the production check was removed.  I'm assuming this was for testing, but we should add it back in before merging.l

main.js
```diff
@@ -139,36 +139,25 @@ function startApp() {
...
+    release: process.env.SLOBS_VERSION, 
+    sampleRate: 0.1
```
andycreeth: Let's actually kill the sample rate here.  I may regret this but... Main process exceptions should be extremely rare and low volume (they are fatal to the application) so let's just record everything.

main.js
RodrigoHolztrattner: Done!

main.js
RodrigoHolztrattner: Oh I totally forgot, thanks for pointing, fixed!

main.js
RodrigoHolztrattner: Removed it!

RodrigoHolztrattner:
Ops, now it should be ok.
I was trying to send the username but I believe that this info is only available on the `app.ts` file. It's ok to not send the username or there is a way to retrieve it from the `main.js`?

andycreeth:
There's not really a good way to get it in the `main.js` context.  I think it's fine not to send for now.

PR #1583, opened 2019-04-02
Add Edit Transform menu

gettinToasty:
awaiting design review from Morgan but the functionality should be all there


andycreeth:
A couple things before I dive into review:
- Negative values are not valid for crop or size.  You can't type in a negative value, but you can get a negative value by clicking the down arrow from `0`
- Negative values are valid for position, but right now they can't be typed.

PR #1588, opened 2019-04-03
De-jank folders UI

andycreeth:
Changes:
- Show a open/close "caret" even for empty folders.  We were displaying a `+` here before (default in sl-vue-tree maybe?) but this just makes things consistent.
- Fix the hover class so hovering over a single element in a folder doesn't highlight actions for the entire folder and all parent folders too
- Fix so that hiding a single item in a folder doesn't show the entire folder as hidden.  For a folder to show as hidden, every item in the folder needs to be hidden.  A future refactor would be to make folders have a third icon which indicates when they are in a "mixed" state - i.e. not all items are visible or hidden, but it's a mix of visibility.  Same for locked.  But for now this brings visibility in line with the way we handle locked state so at least we are consistent.

### DIFF
--- a/app/services/selection/selection.ts
+++ b/app/services/selection/selection.ts
@@ -457,15 +457,15 @@ export class Selection implements ISelection {
    * A selection is considered visible if at least 1 item
    * in the selection is visible.
    */
-  isVisible() {
-    return this.getItems().find(item => item.visible);
+  isVisible(): boolean {
+    return !!this.getItems().find(item => item.visible);
   }
 
   /**
    * A selection is considered locked if all items in the
    * selection are locked.
    */
-  isLocked() {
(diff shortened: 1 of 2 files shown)


app/services/selection/selection.ts
```diff
@@ -453,10 +453,18 @@ export class Selection implements ISelection {
     }
   }
 
+  /**
+   * A selection is considered visible if at least 1 item
+   * in the selection is visible.
+   */
   isVisible() {
-    return !this.getItems().find(item => !item.visible);
+    return this.getItems().find(item => item.visible);
```
holiber: The returning type now is `Item`. Must be `boolean`

andycreeth: Good call.

PR #1591, opened 2019-04-03
Refactor themes to be scalable

gettinToasty:
TODO next phase:
-- add in a few more themes
~-- interface with display api to update display gutters when theme is changed~

### DIFF
--- a/app/components/MixerVolmeter.vue.ts
+++ b/app/components/MixerVolmeter.vue.ts
@@ -53,7 +53,7 @@ export default class MixerVolmeter extends Vue {
   }
 
   get backgroundColor() {
-    return this.customizationService.getThemeBackground();
+    return this.customizationService.themeBackground;
   }
 
   destroyed() {
--- a/app/components/MixerVolmeter.vue.ts
+++ b/app/components/MixerVolmeter.vue.ts
@@ -20,8 +20,6 @@ const YELLOW = 'rgba(255,205,71,.6)';
 const YELLOW_BG = 'rgba(255,205,71,.16)';
 const RED = 'rgba(252,62,63,.6)';
 const RED_BG = 'rgba(252,62,63,.16)';
-const NIGHT_BG = '#09161d';
-const DAY_BG = '#f7f9f9';
 
 @Component({})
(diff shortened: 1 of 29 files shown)


app/components/MixerVolmeter.vue.ts
```diff
@@ -55,7 +53,7 @@ export default class MixerVolmeter extends Vue {
   }
 
   get backgroundColor() {
-    return this.customizationService.nightMode ? NIGHT_BG : DAY_BG;
+    return this.customizationService.getThemeBackground();
```
andycreeth: This will probably only ever be mounted in the main window, and it will be reactive there.  But probably best to make this a getter so it's reactive anywhere.

app/services/customization/customization.ts
```diff
@@ -108,11 +117,16 @@ export class CustomizationService extends PersistentStatefulService<ICustomizati
...
-        type: 'OBS_PROPERTY_BOOL',
+      <IObsListInput<string>>{
+        value: settings.theme,
+        name: 'theme',
+        description: $t('Theme'),
+        type: 'OBS_PROPERTY_LIST',
+        options: [
+          { value: 'night-theme', description: $t('Night (Classic)') },
+          { value: 'day-theme', description: $t('Day (Classic)') },
+          { value: 'spacegray-theme', description: $t('Space Gray') },
```
andycreeth: I wish I was joking but this is probably trademarked by Apple lol.

gettinToasty: whoops meant to take out that option was just used to test

gettinToasty:
@andycreeth aside from removing the toggle on the main UI these changes should be invisible-- i'd like to work with design on developing new themes though

app/services/customization/customization.ts
andycreeth: Haha was gonna say.  I couldn't see any styles for it anywhere.  Honestly I was looking forward to having a space gray theme in SLOBS.  Sounds amazing 😄

gettinToasty: it honestly looks really good but i didn't want to include anything without design input

PR #1603, opened 2019-04-08
Fix validation messages

holiber:


gettinToasty:
i'm really confused on our looseness with the terms `options` and `metadata` when it gets down to `BaseInput` label-- in my mind `options` should be a property of `metadata` used for `ListInputs` while i dont see why we can't keep using `metadata` for the rest of these attributes (or rename it attributes?)

holiber:
@gettinToasty `metadata` is a component property. Vue doesn't allow to mutate property inside component so we need some other field to setup default values for metadata. 

Do you suggest renaming `options` to `attributes`?

holiber:
Worth to do after we will have tests for input components

PR #1607, opened 2019-04-08
Tests config reloading for advanced audio settings

holiber:


holiber:
This dest doesn't pass because we have a bug in OSN.  Waiting for fix https://app.asana.com/0/734357654754972/1117754653664838

PR #1609, opened 2019-04-09
add background workers to apps

andycreeth:


app/services/platform-apps/index.ts
```diff
@@ -359,14 +371,20 @@ export class PlatformAppsService extends StatefulService<IPlatformAppServiceStat
...
-    // Only 1 page per slot top is allowed
+    // Only 1 page per slot is allowed
     const seenSlots: Dictionary<boolean> = {};
 
     for (let i = 0; i < manifest.pages.length; i++) {
       const page = manifest.pages[i];
 
       this.validateObject(page, `manifest.pages[${i}]`, ['slot', 'file']);
 
+      if (!Object.values(EAppPageSlot).includes(page.slot)) {
```
blackxored: AFAIK this would fail at runtime because EAppPageSlot is not a const enum?

andycreeth: All enums in TypeScript are objects at runtime these days.

blackxored: Resolving per discussion.

PR #1612, opened 2019-04-09
Fix re-fetching streaming key

holiber:
`@requiresToken()` doesn't work correctly
`@requiresToken` decorator expects a `http response` , not a `http response body`
when it checks `error.status === 401` the field status can exist in `http response` but there are no guarantees that `status` field will be in the `http response body`


holiber:
@andycreeth https://app.asana.com/0/959277823451735/1117962668774686

PR #1613, opened 2019-04-09
hotkeys settings performance

andycreeth:


app/components/Hotkeys.vue.ts
```diff
@@ -10,21 +10,19 @@ import HotkeyGroup from './HotkeyGroup';
...
-  hotkeySet = this.hotkeysService.getHotkeysSet();
-
-  created() {
+  mounted() {
     // We don't want hotkeys registering while trying to bind.
     // We may change our minds on this in the future.
     this.hotkeysService.unregisterAll();
+
+    // Render a blank page before doing synchronous IPC
+    setTimeout(() => (this.hotkeySet = this.hotkeysService.getHotkeysSet()), 100);
```
holiber: Our ChildWindow component doesn't handle this?

andycreeth: Unfortunately not, because the top level component (`Settings.vue`) is already mounted and usable.  This is for when you switch to the `Hotkeys` tab, and things hang for a little bit.

PR #1618, opened 2019-04-10
Split prod and dev webpack bundles

gettinToasty:
This does a couple of things:
* Lets us optimize better for both use-cases (speed in dev, size in prod)
* Updates Beaker to a new version that includes an esm bundle which allows it to be tree-shaken properly (as well as various other optimizations)

The configs:
* Dev
  * Refocus entry to be solely on the `renderer.js` bundle, there are huge wins to build time here but with a gotcha that any work done on the updater JS portions/external-api will need to `compile:production` instead of `yarn watch`
  * Only resolve rules in appropriate directories-- by whitelisting our loader rules we can significantly decrease build-time by not checking each and every file
  * Move to `cheap-module-source-map`, a speedier version of source-map that cuts our rebuild time about 3x
  * The numbers: Initial build went from ~85s to ~49s on my machine, re-build from ~3s to ~1s

* Prod
  * Mostly the same aside from loader whitelisting
  * Run TerserPlugin in parallel mode for improved performance
  * Minified bundle is down to about ~6.3mb with beaker treeshaking

One other thing this does is include a commented out path for our biggest bottleneck: ts-loader. Running it in `transpileOnly` mode with `ForkTsCheckerWebpackPlugin` would do wonders for build time but needs a lot of debugging and further research


andycreeth:
Hmm, I think we either want the production one to extend the dev one, or we need a base config that they both extend.  Otherwise we have far too much duplication between the two and they will get out of sync.

gettinToasty:
@andycreeth that should be totally do-able, lemme see if i can get to a good solution there 👍

PR #1627, opened 2019-04-12
Increase tests timeout

holiber:


andycreeth:
Do we have any theories as to why this would be needed.  Our longest test should only take ~1min AFAIK.

holiber:
@andycreeth I saw tests with ~1:30 in CI. So it's not a big gap to the 2m limit. I have a theory that tests just started hanging on appclose. If this PR won't help this will be a proof  for this theory

PR #1629, opened 2019-04-12
Use CSS and transitions to fix transisiton test failure

gettinToasty:


gettinToasty:
@andycreeth wouldnt tests in infinite loading states just time out or run into other functionality blockers? i feel like having a single UI element being the only gateway for app functionality is already not really robust if that's the case

gettinToasty:
We could wrap the main view's actual content in a v-if that only renders if the app is loaded, kind of like we do with the child window

andycreeth:
> wouldnt tests in infinite loading states just time out or run into other functionality blockers?
Not necessarily.  It really depends what the test is testing for.  We've had issues in the past where everything loading properly but for whatever reason the loader didn't disable properly.

> i feel like having a single UI element being the only gateway for app functionality is already not really robust if that's the case
I don't understand what you mean by this.

holiber:
I debugged some failed tests and looks like our code that disables animation doesn't work sometimes. Please hold on this PR while I'm digging more info

holiber:
This PR fixes the issue with transitions tests https://github.com/stream-labs/streamlabs-obs/pull/1640

PR #1640, opened 2019-04-16
Tests fixes

holiber:
- fixed a bug when animations are not correctly disabled in the Child Window that makes `transition` tests failed (race-condition)
- fixed a shutdown error when child window sends `showErrorAlert` message to already closed main window (race-condition)
- fixed a bug when tests unable to skip onboarding steps because of the loader screen
- fixed `recording` test (race-condition)
- fixed flaky `goal-create` tests
- added catching uncaught errors and unhandled promises rejection via electron log
- fixed console errors in `Stream Boss` test
- fixed console errors in `Onboarding` test
- fixed uncaught promise rejection from `stream-info.ts`
- fixed flaky `goal` tests


app/services/stream-info.ts
```diff
@@ -44,7 +44,7 @@ export class StreamInfoService extends StatefulService<IStreamInfoServiceState>
   streamInfoChanged = new BehaviorSubject<IStreamInfo>(StreamInfoService.initialState);
 
   init() {
-    this.refreshStreamInfo();
+    this.refreshStreamInfo().catch(e => null);
```
gettinToasty: why are we swallowing the error here but you changed the promise rejection from null to a text string?

test/helpers/spectron/index.ts
```diff
@@ -47,6 +47,25 @@ export async function closeWindow(t: any) {
...
+      return;
+    } catch (e) {
+      err = e;
+      await sleep(500);
+    }
+  }
+  console.error(`unable to click ${selector}`);
+  throw err;
+}
+
```
gettinToasty: i'll be honest i really don't like this solution

app/services/stream-info.ts
holiber: Because `refreshStreamInfo()` is calling not only from this place. The rejection message helps to find unhandled promises rejections

test/helpers/spectron/index.ts
holiber: Replaced this

app/services/widgets/settings/widget-settings.ts
```diff
@@ -17,6 +17,7 @@ import { Subject } from 'rxjs';
 import { IInputMetadata } from 'components/shared/inputs/index';
 import { mutation, StatefulService } from 'services/stateful-service';
 import { WebsocketService } from 'services/websocket';
+import electronLog from 'electron-log';
```
gettinToasty: unused import

test/helpers/spectron/index.ts
```diff
@@ -173,11 +177,12 @@ export function useSpectron(options: ITestRunnerOptions = {}) {
     }
   }
 
-  async function stopApp(t: TExecutionContext) {
+  async function stopApp() {
     try {
-      await context.app.stop();
+      await app.stop();
```
gettinToasty: is there a reason we remove the execution context here?

test/helpers/widget-helpers.ts
```diff
@@ -2,19 +2,37 @@ import { getClient } from './api-client';
...
+  return new Promise(async (resolve, reject) => {
+    // start listen widgets SettingsUpdate events
+    const listener = await getListener();
+
+    // execute code
+    await fn();
+
+    // catching events for 5s
+    listener.pipe(first()).subscribe(ev => resolve(ev));
+    await sleep(5000);
```
gettinToasty: is looks like it'll greatly increase our test run-time, is that not the case?

test/helpers/spectron/index.ts
```diff
@@ -147,7 +145,8 @@ export function useSpectron(options: ITestRunnerOptions = {}) {
 
     // Pretty much all tests except for onboarding-specific
     // tests will want to skip this flow, so we do it automatically.
-    await t.context.app.client.waitForVisible('a=Setup later'); // wait for the loader disappearing
+
+    await t.context.app.client.waitForExist('.main-loading', 5000, true);
```
gettinToasty: does this basically mean we're waiting 5s after the main loader appears each time? or is it a timeout of 5s? if it's the former i feel like with how many times we start the app in our test suite this is going to take forever

PR #1641, opened 2019-04-16
[WIP] init OBS before mounting vue

andycreeth:


app/components/windows/Main.vue
```diff
@@ -19,7 +19,7 @@
       />
     </div>
 
-    <div class="main-middle" :class="mainResponsiveClasses" v-if="!showLoadingSpinner" ref="mainMiddle">
+    <div class="main-middle" :class="mainResponsiveClasses" ref="mainMiddle">
```
holiber: Is that supposed to be in this PR?

PR #1647, opened 2019-04-18
Proposal for Migration system

gettinToasty:
If we're going to be handling migrations (which I think we should in order to remain extensible and future-proof) there should be a system for it. Here's an implementation that I think could work well. I chose to have migrations be passed in explicitly rather than using something like inheriting `this.migrations` to bring attention to the fact that they exist. Not sure if setting state directly is the right choice here, and would appreciate any other feedback as well


app/services/core/persistent-stateful-service.ts
```diff
@@ -38,4 +45,15 @@ export abstract class PersistentStatefulService<TState extends object> extends S
       },
     );
   }
+
+  runMigrations(persistedState: any, migrations: IMigration[]) {
+    const migratedState = cloneDeep(persistedState);
+    migrations.forEach(migration => {
+      if (!persistedState[migration.oldKey]) return;
```
andycreeth: `if (persistedState[migration.oldKey] == null)`
This way we properly migrate falsey values like `0`

app/services/core/persistent-stateful-service.ts
```diff
@@ -38,4 +45,15 @@ export abstract class PersistentStatefulService<TState extends object> extends S
...
+      migratedState[migration.newKey] = migration.transform(persistedState[migration.oldKey]);
+      migratedState[migration.oldKey] = null;
```
andycreeth: `delete migratedState[migration.oldKey]`
This will actually remove the old key from the object instead of keeping and nulling it.

PR #1649, opened 2019-04-19
Rewrite Volmeters in WebGL

andycreeth:
We need to put this through a battery of benchmarks, but it looks like this could save us a bunch of CPU.


main.js
```diff
@@ -34,7 +34,7 @@ if (process.argv.includes('--clearCacheDir')) {
   rimraf.sync(app.getPath('userData'));
 }
 
-app.disableHardwareAcceleration();
+// app.disableHardwareAcceleration();
```
holiber: is it required?

holiber:
Sometimes I'm just thinking about the idea of having all pre-rendered voltmeters states into a single image and changing the visible area of this image.

main.js
andycreeth: Yes.  WebGL does not work with hardware acceleration off.  I will remove this rather than commenting out before this is merged.

andycreeth:
@holiber I'm not sure that would be as efficient.  Our original volmeter implementation was based on using CSS transforms, which ended up being less efficient than our most recent Canvas 2D implementation, and significantly less efficient than this WebGL implementation.  But if you do have any other ideas you want to try I am all ears and certainly not married to this implementation.

app/services/performance.ts
```diff
@@ -50,15 +52,34 @@ export class PerformanceService extends StatefulService<IPerformanceState> {
...
+
+        stats.CPU += electronCpu;
+
         this.SET_PERFORMANCE_STATS(stats);
       },
     );
+
+    window['startMeasure'] = () => {
+      this.measurements = [];
+    };
```
andycreeth: I'll be backing these changes out of this PR.  Was just using this to sanity check the performance increases I was seeing.

holiber:
I mean if you have a set of pre-rendered frames in one image you can just change the position of the visible part of the image. So the computer renders the image only once. However, I can't say this way will be more efficient without testing. Just letting know that we didn't try it. 
image

andycreeth:
Yeah essentially a sprite sheet.  My gut tells me using sprites to animation the volmeters with be more expensive.  But yes we haven't tried it.

andycreeth:
Looks like our CI machines don't support hardware acceleration, so this will always raise an error there.  I am either going to keep the old volmeters as a fallback or just disable it on our CI.  We need to do more research on how many users are going to be on machines that won't support WebGL.

andycreeth:
On second thought I'm definitely going to keep the old code as a fallback.  It's easy enough to do and will guarantee the best support for the most people.

andycreeth:
Ok should properly fall back to Canvas 2D now on unsupported machines.  The 2 renderers produce visually identical results.

app/components/MixerVolmeter.vue.ts
```diff
@@ -149,7 +282,12 @@ export default class MixerVolmeter extends Vue {
   subscribeVolmeter() {
     this.volmeterSubscription = this.audioSource.subscribeVolmeter(volmeter => {
       this.setChannelCount(volmeter.peak.length);
-      this.drawVolmeter(volmeter.peak);
+
+      if (this.gl) {
+        this.drawVolmeterWebgl(volmeter.peak);
+      } else {
+        this.drawVolmeterC2d(volmeter.peak);
+      }
```
holiber: If we pay attention on performance in this PR, we should draw volmeters on `requestAnimationFrame` event. It allows skipping some frames if CPU is overloaded

andycreeth: I don't think that helps us here.  `requestAnimationFrame` executes the callback before the next paint.  In an animation loop using `requestAnimationFrame`, you would queue up your next `requestAnimationFrame` from within your callback.  This means that effectively your your callback interval matches your paint cycle.  As that slows, so does your interval.

However, simply adding a call in this callback to `requestAnimationFrame` does nothing for us, because the interval is a set 40ms by OBS.  Our callback will be called every 40ms by OBS regardless of the paint speed.  `requestAnimationFrame` would not save us any work because we would be using it incorrectly.

A potential solution could be to separate the OBS callback and the render callback.  So we simply store the latest data on the frontend, and then asynchronously run a traditional `requestAnimationFrame` loop that picks up that data and draws it with WebGL.  However I'm skeptical that would actual cause any tangible benefit in CPU usage.

PR #1651, opened 2019-04-20
Add mouse hotkey support.

boarder2:
https://ideas.streamlabs.com/ideas/SL-I-3412

Implemented by using a select box for the mouse button you want to bind to since hooking to mouse button presses in electron isn't great.

Also fixed the CSS to be more responsive.


andycreeth:
Hey @boarder2 thanks for taking the time to implement this.  Mouse hotkey support is something that has been on our backlog for a long time, and we love getting contributions from the community on stuff like this.  I especially appreciate you added a test case as well.

We're currently focused on getting a new version shipped, so we might be a little slow to give this PR the full attention it deserves.  I just wanted to let you know so you don't think we're ignoring this PR.  Thanks again.

andycreeth:
Hey, thanks for your patience.  I took some time to play around with your branch and discuss with the rest of the team.  We appreciate the effort you put into this PR, but we need the UX to be a bit different before we can merge it in.  We would like to match OBS's behavior for mouse hotkeys.  This means that rather than selecting from a dropdown, you just press the mouse button to bind it.

I don't believe there are javascript events for getting Mouse4 and above.  I assume that's what you meant by mouse hooking in electron not being great.  However, I think this can be worked around by using our `key-listener` service to temporarily listen for those mouse buttons while the settings window is open/focused.  Additionally we would like to prevent binding Left/Right (again, same as OBS Studio) because these are critical for navigating SLOBS and other applications and binding them as a global hotkey is likely a mistake.

andycreeth:
Let me know if you are interested in taking another crack at this given the above feedback.  Otherwise I can also have someone at Streamlabs take what you've started here and make the necessary modifications.  Thanks again for your help.

boarder2:
@andycreeth I'm interested in giving it a shot. I played around with it a little bit this afternoon and I'm running into some trouble trying to get the binding callback to work at the Hotkeys.vue.ts level. I figure that's where it should go and then maybe use EventBus to inform each of the Hotkey's that an event happened and let them decide if they have focus and whether to respond to it.

My experience with Vue is limited to only the tiny amount I've done here with this project so I may not be the best person to get this done in a timely fashion.  If you guys were looking to prioritize this next then by all means have at it!

andycreeth:
Ah yeah so what you're running into is that our application runs multiple electron windows which means multiple processes.  When you inject a service into a component in the child window (where settings lives), you are only getting a proxy that relays your function call and response to main window.  Your callback is getting lost because it's not serializable.  Our RPC mechanism does however have support for subscribing to RxJs observables across the child-main barrier, so maybe something could be set up there.

boarder2:
This may be as close as I can get without help.  I'm directly using libuiohook in the Hotkeys vue which I think violates your standard for requiring service wrappers for that kind of stuff.  Someone more versed in this framework ought to be able to fix that up pretty quick.

Everything does work, sort of. Some hotkey groups don't get the observable that's being passed down. Given the log on https://github.com/boarder2/streamlabs-obs/blob/b2aa423892e5984fc1cc35408d658adf6753ea4d/app/components/HotkeyGroup.tsx#L29

It has the following output

```
HotkeyGroup.tsx:29 Rendering hot key group:  undefined undefined
HotkeyGroup.tsx:29 Rendering hot key group:  Starting Soon undefined
HotkeyGroup.tsx:29 Rendering hot key group:  Live Scene undefined
HotkeyGroup.tsx:29 Rendering hot key group:  Be Right Back undefined
HotkeyGroup.tsx:29 Rendering hot key group:  Panels undefined
HotkeyGroup.tsx:29 Rendering hot key group:  Webcam Subject {…}
HotkeyGroup.tsx:29 Rendering hot key group:  Game Capture Subject {…}
HotkeyGroup.tsx:29 Rendering hot key group:  Mic/Aux Subject {…}
HotkeyGroup.tsx:29 Rendering hot key group:  Desktop Audio Subject {…}
HotkeyGroup.tsx:29 Rendering hot key group:  Game Audio Subject {…}
HotkeyGroup.tsx:29 Rendering hot key group:  Chat (Discord) Subject {…}
```

I'm not sure why some groups would get the observable and some wouldn't.

If anyone's able to give some guidance I'll keep plugging away at this, otherwise I think this is a pretty good base to start from.

I'm also not sure if it's possible to write a test for this now so I've commented that out.

andycreeth:
Thanks again for your help with this - this was a great base to work off of.  We recently upgraded to electron 5, and the mouse events are handled a bit differently.  I was able to implement the mouse hotkey bindings using just vanilla JS events.  The PR here supersedes this one: https://github.com/stream-labs/streamlabs-obs/pull/1885

PR #1652, opened 2019-04-22
De-jankify window resizing

gettinToasty:
This PR does a couple things--
1. Makes out Electron windows transparent which removes the white flash we see on resizes
2. Add more disabling of style blocking elements, specifically to all window resizes and other places we use animations

Slight side effect that when resizing a child window the preview on the main window will also be blocked from displaying which is a little weird but overall a much smoother experience. Might be worth isolating the style-blocking elements on a per-window basis in a future iteration


andycreeth:
Ty for taking the initiative on this.  After playing around with the branch I think this is a definite improvement in jank reduction.  I'm wondering however if we should try to tackle the 1 remaining piece of jank that you mentioned.  Theoretically we should be able to make `hideStyleBlockingElements` to a different service (it doesn't really belong in customization anymore anyway) and then add some mechanism in to prevent this mutation from getting synced to other windows (I can help with this).  I think that should result in the behavior you want.

andycreeth:
I think we should have QA play with this PR.  This touches a lot of high risk areas.  Especially the `ChildWindow` component which is really really brittle and does some very complex things.

app/components/windows/Main.vue.ts
```diff
@@ -158,6 +158,8 @@ export default class Main extends Vue {
 
   created() {
     window.addEventListener('resize', this.windowSizeHandler);
+
+    document.getElementsByTagName('body')[0].classList.add(this.theme);
```
andycreeth: Why does this need to be on `body` now?  I'm not a huge fan of this or the replacement we're doing in the customization service.

app/components/windows/Main.vue.ts
```diff
@@ -168,27 +170,34 @@ export default class Main extends Vue {
 
   hasLiveDock = true;
 
+  windowResizeTimeout: number;
+
   windowSizeHandler() {
+    this.onResizeStartHandler();
```
andycreeth: As far as I can tell, this is going to cause us to commit a new mutation in the windows service every time a resize event happens, which is many times per second while resizing.  This mutation will be synced across all windows through the main process, which is pretty slow.  If that is happening (I think it likely is) then we should make sure we only commit 1 mutation at the start of resizing and one at the end.  This will be much better for smooth resizing on slow machines I think.

It's also going to be doing a synchronous IPC to the windows service on every resize event (`updateStyleBlockers`) which is also going to slow things down considerably.

app/components/windows/ChildWindow.vue.ts
```diff
@@ -85,8 +94,33 @@ export default class ChildWindow extends Vue {
...
     this.refreshingTimeout = window.setTimeout(() => {
-      this.components.push({ name: options.componentName, isShown: true, title: options.title });
+      this.components.push({
+        name: options.componentName,
+        isShown: true,
+        title: options.title,
+        hideStyleBlockers: options.hideStyleBlockers,
+      });
       this.setWindowTitle();
+      window.addEventListener('resize', this.windowSizeHandler);
```
andycreeth: It seems a little complex to be adding and removing the resize listener every time we mount and unmount a component to the window.  Is the a reason we can't just have the the resize handler mount when the `ChildWindow` component mounts and just keep it there?

app/components/windows/ChildWindow.vue.ts
```diff
@@ -57,24 +57,33 @@ export default class ChildWindow extends Vue {
     electron.remote.getCurrentWindow().setTitle(this.currentComponent.title);
   }
 
+  windowResizeTimeout: number;
+
+  windowSizeHandler() {
+    this.windowsService.updateStyleBlockers('child', true);
```
andycreeth: Same comment as `Main.vue.ts` - I think we're going to be using a lot of redundant IPC doing this.

app/components/windows/OneOffWindow.vue.ts
```diff
@@ -27,4 +35,16 @@ export default class OneOffWindow extends Vue {
   get theme() {
     return this.customizationService.currentTheme;
   }
+
+  windowResizeTimeout: number;
+
+  windowSizeHandler() {
+    this.windowsService.updateStyleBlockers(this.windowId, true);
```
andycreeth: Same comment about potentially being too IPC heavy.

app/components/windows/Main.vue.ts
gettinToasty: this isn't actually necessary, was an artifact from when i was trying out some things; will remove it

app/components/windows/ChildWindow.vue.ts
gettinToasty: using the vue lifecycle methods was giving me some extremely weird bugs when i was testing that out, so this was the closest compromise i could find

andycreeth: Hmm.  Could be related to the fact that we are hiding the window without destroying it.

PR #1655, opened 2019-04-23
Multiplatform streaming tests

holiber:
Mixer tests are disabled due issue with a blocked account


app/components/shared/inputs/ListInput.vue.ts
```diff
@@ -29,6 +29,8 @@ export default class ListInput extends BaseInput<string, IListMetadata<string>>
     const options = super.getOptions();
     return {
       ...options,
+      // internalSearch is `true` by default in vue-multiselect
+      internalSearch: options.internalSearch === void 0 ? true : options.internalSearch,
```
andycreeth: `== null` is more robust as it also checks for `null` and is also more readable.  We should use `== null` over `=== void 0` in pretty much every scenario.

PR #1660, opened 2019-04-24
 fix webgl warnings when audio source has 0 channels

andycreeth:


gettinToasty:
should this be `<=` or does it not error out with width/height of 0

andycreeth:
AFAIK `0` is a valid but fairly useless value to set.  Main objective was removing the errors but maybe it should be `<= 0`

PR #1664, opened 2019-04-27
Browser frame

AnkhHeart:
- BrowserView implementation
- Replaced Chatbot.Vue & ts with Browser Frame instead
- Replaced Webview in Dashboard with BrowserFrame
- Added support for hidenav & l (locale) query params to dashboardUrl instead of cookie injection
- Added Affinity to browserframes `streamlabs` improves tab swapping between Cloudbot, Dashboard & Live (Faster Load)
- Replaced recent events with browser frame


andycreeth:
Going to revisit this at a later date.  For now we are going to stick with webviews for these types of pages.  The webview tag is still not officially deprecated as of electron 5, it's just not recommended.  When we have more time we will revisit.

PR #1671, opened 2019-05-01
Finalize ServiceManager refactoring

holiber:
All base classes for services now are in the `services/core` folder
image

Also, `handleServiceRequest` method has been refactored in `tpc-api.ts`


andycreeth:
Still reviewing this, but just wanted to comment that you are using several different import styles even within this PR:
`import { Inject } from '../../services/core/injector'`
`import { Inject } from 'services/core/injector'`
`import { Inject } from 'services'`

Not a big deal, but seems like there is some value in being consistent.  Option 3 seems the cleanest to me.

holiber:
My IDE automatically refactored import paths. I tried to replace all of them to `import { Inject } from 'services'` but after that the application doesn't start (feels like some issue with dependency resolving). So I rolled back to my IDE's solution

andycreeth:
I feel like there was a mistake somewhere.  This is webpack syntactic sugar.  It shouldn't change anything else.

PR #1675, opened 2019-05-02
Add Patch Notes Notification

gettinToasty:


app/components/NotificationsArea.vue.ts
```diff
@@ -50,6 +50,10 @@ export default class NotificationsArea extends Vue {
 
       this.showExtendedNotifications = this.$refs.notificationsContainer.offsetWidth >= 150;
     }, 1000);
+
+    if (this.notificationsService.state.notifications.length) {
+      this.notificationsService.state.notifications.forEach(this.onNotificationHandler);
+    }
```
andycreeth: Is this solving for a race condition where there are notifications already in the queue when the component mounts?

gettinToasty: exactly

PR #1676, opened 2019-05-03
Implement Push to Show/Hide Source feature.

commagray:
This one implements a missing feature what I need: an ability to hide and show things while key is hold, not just pressed.

I'm not a JS developer, so it's a silly hack and I don't know if I'm doing it in the right way.


andycreeth:
Hey, we talked about this as a team and think this is a great useful feature.  Thank you for taking the time to submit a pull request so everyone can benefit!

PR #1690, opened 2019-05-09
Use awesome-typescript-loader

gettinToasty:
Also uses hard-source-plugin, which caches webpack builds and drastically reduces build time after an initial build


base.config.js
```diff
@@ -57,20 +59,24 @@ module.exports = {
...
           },
+          loaders: { ts: 'awesome-typescript-loader' }
         }
       },
       {
         test: /\.ts$/,
-        loader: 'ts-loader',
-        options: { experimentalWatchApi: true },
+        loader: 'awesome-typescript-loader',
+        options: { useCache: true, forceIsolatedModules: true, reportedFiles: ['app/**/*.ts'] },
```
holiber: Why we need `isolatedModules` here?

gettinToasty: it's just mentioned in the `awesome-typescript-loader` docs as something that improves build speed, but if there's a reason to not have it would definitely be ok without it

holiber: As I see from docs:
> Transpile each file as a separate module (similar to 'ts.transpileModule'). 
That means we're gonna lose cross-file type checking with this option.

gettinToasty: afaik this only effects the transpilation process and not the typechecking process, which runs in a separate thread

PR #1699, opened 2019-05-11
Fix : typos in french translation

r00tKiller:
There was two typos in the french translation.


holiber:
Thanks for your PR. We maintain translation via Crowding platform. I'll add your fixes there.

PR #1701, opened 2019-05-13
Adds an error message if something goes wrong while trying to go live…

gettinToasty:
… on FB


app/components/windows/EditStreamInfo.vue.ts
```diff
@@ -298,9 +298,19 @@ export default class EditStreamInfo extends Vue {
...
-    this.windowsService.closeChildWindow();
+  async goLive() {
+    try {
+      await this.streamingService.toggleStreaming();
+      this.windowsService.closeChildWindow();
+    } catch (e) {
+      this.$toasted.show(e, {
+        position: 'bottom-center',
+        className: 'toast-alert',
+        duration: 1000,
```
andycreeth: Is this 1 second?  That doesn't seem long enough to read anything.

andycreeth: Also will this just show the contents of the exception to the user?  I'm not sure that's the best thing to show, as we don't really know what it will contain.

gettinToasty: can lengthen it. the error message is abstracted in the `facebook` service:
```
    return fetch(request)
      .then(handlePlatformResponse)
      .then(json => {
        const streamKey = json.stream_url.substr(json.stream_url.lastIndexOf('/') + 1);
        this.SET_LIVE_VIDEO_ID(json.id);
        this.setSettingsWithKey(streamKey);
      })
      .catch(resp => Promise.reject(resp.error.message));
```

andycreeth: Ok so this is the error message that comes back in the facebook response.  From your experience are these generally messages that are meant to be shown to a user, or are they more developer centric error messages?

gettinToasty: in the case of the 500 we were getting the message was just something like`An unknown error has occured`

andycreeth: Just just show a generic message along the lines of "Something went wrong while going live, please try again.".  Every time we've shown an API/library error message directly to the user before, I've regretted it because they are rarely meant for end users and look unprofessional.

PR #1710, opened 2019-05-15
Implement Undo/Redo - V1

andycreeth:
This PR implements multi-level undo/redo in SLOBS.  This is implemented using the command pattern.  I still have a handful of commands I want to implement, but this PR should have extremely broad coverage.


app/components/windows/EditTransform.tsx
```diff
@@ -21,52 +24,71 @@ const dirMap = (dir: string) =>
...
     validForm: ValidatedForm;
   };
 
   get transform() {
-    return this.selectionService.getTransform();
+    return this.selection.getItems()[0].transform;
   }
 
-  setTransform(key: string, subkey: string) {
+  setCrop(cropEdge: string) {
```
holiber: It's a limited amount of Edges, right?  cropEdge: 'top' | 'right' | etc..

app/components/windows/SourceFilters.vue.ts
```diff
@@ -44,6 +47,39 @@ export default class SourceFilters extends Vue {
...
+      this.refreshFilters(),
+    );
+  }
+
+  destroyed() {
+    this.addFilterSub.unsubscribe();
+    this.removeFilterSub.unsubscribe();
+    this.updateFilterSub.unsubscribe();
+    this.reorderFilterSub.unsubscribe();
+  }
```
holiber: A little hint on how to reduce some lines of code:
```
const componentSubscriptions = [
  filterUpdated.subscribe( ... ), 
  filtersReordered.subscribe( ... ),
  ...
]

componentSubscriptions.forEach(s => s.unsubscribe);
```

app/services/editor-commands/commands/center-items.ts
```diff
@@ -0,0 +1,32 @@
+import { ModifyTransformCommand } from './modify-transform';
+import { Selection } from 'services/selection';
+
+export enum ECenteringType {
+  Screen = 'screen',
+  Horizontal = 'horizontal',
+  Vertical = 'vertical',
+}
+
+export class CenterItemsCommand extends ModifyTransformCommand {
```
holiber: Why we need to define a special subset of commands like `CenterItemsCommand` if we already have one general Transform command. Looks like this adds more boilerplate code

app/components/windows/EditTransform.tsx
andycreeth: Yes, would be more correct to have this be `keyof ICrop` I think.

app/services/editor-commands/commands/center-items.ts
andycreeth: It's definitely more boilerplate.  One reason is that this command has a different description that is shown in the UI for undo/redo operations.  I tried to encapsulate as much shared logic as possible in the `ModifyTransformCommand` base command.  This also allowed me to leverage all of our existing transform helper functions that operate on selections/scene-items directly, instead of having to generate a new transform and pass it the `SetTransform` command.

app/components/SourceSelector.vue.ts
```diff
@@ -167,11 +174,26 @@ export default class SourceSelector extends Vue {
     const destNode = this.scene.getNode(position.node.data.id);
 
     if (position.placement === 'before') {
-      nodesToMove.placeBefore(destNode.id);
+      this.editorCommandsService.executeCommand(
+        'ReorderItemsCommand',
```
andycreeth: Realizing now that to be correct this command should be renamed to `ReorderNodesCommand` because it operates on Nodes and not just Items.

app/components/windows/EditTransform.tsx
```diff
@@ -21,52 +24,71 @@ const dirMap = (dir: string) =>
...
+        AnchorPositions[AnchorPoint.NorthWest],
+      );
+
       this.rect[dir] = Number(value);
     };
   }
 
   rotate(deg: number) {
-    return () => this.selectionService.rotate(deg);
+    return () => this.selection.rotate(deg);
```
andycreeth: Looks like I missed a command here

app/components/windows/EditTransform.tsx
```diff
@@ -21,52 +24,71 @@ const dirMap = (dir: string) =>
...
   }
 
   reset() {
-    this.selectionService.resetTransform();
+    this.selection.resetTransform();
```
andycreeth: And here

app/services/editor-commands/editor-commands.ts
```diff
@@ -0,0 +1,273 @@
...
+  static initialState: IEditorCommandsServiceState = {
+    undoMetadata: [],
+    redoMetadata: [],
+    operationInProgress: false,
+  };
+
+  // IMPORTANT: Must be kept in sync with metadata stored in
+  // this service's vuex state.
+  undoHistory: Command[] = [];
+  redoHistory: Command[] = [];
```
gettinToasty: not sure why this isnt kept in state in the first place

app/services/editor-commands/editor-commands.ts
```diff
@@ -0,0 +1,273 @@
...
+      this.combineTimeout = null;
+    }, COMBINE_TIMEOUT);
+  }
+
+  executeCommand<TCommand extends keyof typeof COMMANDS>(
+    commandType: TCommand,
+    ...commandArgs: ConstructorParameters<(typeof COMMANDS)[TCommand]>
+  ) {
+    // Executing any command clears out the redo history, since we are
+    // creating a new branch in the timeline.
```
gettinToasty: but what if we replace the infinity stones at the exact moment they were taken?

app/services/editor-commands/editor-commands.ts
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    if (instance instanceof CombinableCommand) {
+      if (this.combineActive) {
+        const previousCommand = this.undoHistory[this.undoHistory.length - 1];
+        if (previousCommand.constructor === instance.constructor) {
+          // This check is actually redundant since we just checked the constructor
+          // is the same above, but TypeScript isn't that smart.
+          if (previousCommand instanceof CombinableCommand) {
+            if (previousCommand.shouldCombine(instance)) {
```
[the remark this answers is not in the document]
gettinToasty: i feel like we could combine all these nests into a single conditional

app/services/editor-commands/editor-commands.ts
```diff
@@ -0,0 +1,273 @@
...
+          });
+      }
+
+      this.undoHistory.push(command);
+      this.PUSH_UNDO_METADATA({ description: command.description });
+    }
+  }
+
+  private handleUndoRedoError(undo: boolean, e: any) {
+    console.error('Error performing undo operation', e);
```
gettinToasty: `undo or redo`

app/services/editor-commands/editor-commands.ts
```diff
@@ -0,0 +1,273 @@
...
+
+  get nextRedo() {
+    return this.state.redoMetadata[this.state.redoMetadata.length - 1];
+  }
+
+  get nextRedoDescription() {
+    return this.nextRedo ? this.nextRedo.description : '';
+  }
+
+  // Shortcuts for undo-able editor commands go here:
```
gettinToasty: why are these here now?

app/services/scenes/scenes.ts
```diff
@@ -227,13 +232,16 @@ export class ScenesService extends StatefulService<IScenesState> {
...
     this.scenes.forEach(scene => {
-      const items = scene.getItems().filter(sceneItem => sceneItem.sourceId === sourceId);
-      if (items.length > 0) resultScenes.push(scene);
+      scene.getItems().forEach(sceneItem => {
+        if (sceneItem.sourceId === sourceId) count += 1;
+      });
     });
-    return resultScenes;
+
+    return count;
```
gettinToasty: i feel like a double filter would be more effective here

app/services/selection.ts
```diff
@@ -648,19 +626,19 @@ export class Selection {
     this.getNodes().forEach(node => node.remove());
   }
 
-  nudgeActiveItemsLeft() {
+  nudgeLeft() {
```
gettinToasty: not sure if these are missed commands or not

app/store/index.ts
```diff
@@ -104,7 +104,7 @@ export function createStore(): Promise<Store<any>> {
     modules: {
       ...statefulServiceModules,
     },
-    strict: debug,
+    strict: false,
```
gettinToasty: i'm guessing this should be changed back in the final PR

app/util/menus/GroupMenu.ts
```diff
@@ -52,13 +59,18 @@ export class GroupMenu extends Menu {
...
-          .copyTo(this.scenesService.activeSceneId);
-        selectedItem.remove();
-        scene.remove();
+        // const scene = this.scenesService.getScene(selectedItem.getSource().sourceId);
+        // scene
+        //   .getSelection()
+        //   .selectAll()
+        //   .copyTo(this.scenesService.activeSceneId);
+        // selectedItem.remove();
+        // scene.remove();
```
gettinToasty: extraneous comments?

app/components/ConnectionSettings.vue.ts
```diff
@@ -20,27 +22,33 @@ export default class SceneTransitions extends Vue {
...
   set transitionModel(value: string) {
-    this.transitionsService.updateConnection(this.connectionId, { transitionId: value });
+    this.editorCommandsService.executeCommand('EditConnectionCommand', this.connectionId, {
+      transitionId: value,
+    });
   }
 
   get connection() {
-    return this.transitionsService.getConnection(this.connectionId);
+    return this.transitionsService.state.connections.find(conn => conn.id === this.connectionId);
```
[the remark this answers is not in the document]
andycreeth: Yup it was.  I can check if `getConnection` is still in use elsewhere.

app/components/TransitionSettings.vue.ts
```diff
@@ -16,17 +20,29 @@ import HFormGroup from 'components/shared/inputs/HFormGroup.vue';
...
 
+  propertiesChanged: Subscription;
+
+  mounted() {
+    this.propertiesChanged = this.transitionsService.transitionPropertiesChanged.subscribe(id => {
+      if (id === this.transitionId) {
+        this.properties = this.transitionsService.getPropertiesFormData(this.transitionId);
+      }
+    });
+  }
```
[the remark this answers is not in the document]
andycreeth: This is for reactivity.  Properties are pretty complex so we don't keep them in the store.  This allows us to update the form when the properties change due to an external operation (such as undoing a command).  You'll see something similar in source properties as well as filter properties.  Maybe in the future we should just keep properties in the store.  Not sure what the best option is.

app/components/windows/SourceFilters.vue.ts
```diff
@@ -87,19 +123,21 @@ export default class SourceFilters extends Vue {
...
+      this.sourceId,
+      this.selectedFilterName,
+    );
   }
 
   toggleVisibility(filterName: string) {
     const sourceFilter = this.filters.find(filter => filter.name === filterName);
-    this.sourceFiltersService.setVisibility(
+    this.editorCommandsService.executeCommand(
+      'EnableFilterCommand',
```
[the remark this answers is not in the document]
andycreeth: Yeah.  Well, it's not a toggle since it doesn't switch to the opposite of the state it's currently in (although it could be changed to do that).  A long form name would be something like `SetFilterEnabledStateCommand`.  I felt that was a little verbose but it's probably better than using something called `EnableFilterCommand` to disable a filter.

app/services/editor-commands/editor-commands.ts
andycreeth: They are complex classes that can't be serialized.

app/services/editor-commands/editor-commands.ts
andycreeth: I was originally trying not to have the selection service call into the editor commands service, but that broke down when I realized `remove` would be too difficult to port over.  I can probably move them back.

app/services/scenes/scenes.ts
andycreeth: Unless I'm missing something, I think the outer iteration would need to be a reduce operation.  This seemed quick and obvious, but I suppose definitely less functional.

app/services/selection.ts
andycreeth: Nope.  These are lower level functions that are called by the command.

In a typical web app, you can think of our command layer sitting at the endpoint/controller level.  It's as close as possible to where actual requests are coming in from the user.  These functions are more at the model / business logic layer, if that makes sense.  I decided on this separation because I want us to be able to re-use logic as much as possible internally without accidentally pushing undo operations onto the stack.  This is why commands are almost exclusively (with a few exceptions) triggered from Vue components rather than services.

app/store/index.ts
andycreeth: Yup, good catch!  `debug` mode is so insanely slow, I turn it off when I'm working on code that does a lot of mutations.

app/util/menus/GroupMenu.ts
andycreeth: good catch

app/services/editor-commands/editor-commands.ts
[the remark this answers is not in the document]
andycreeth: yup

app/services/scenes/scenes.ts
gettinToasty: ah wait yea i misread this functionality nevermind

app/components/windows/SourceFilters.vue.ts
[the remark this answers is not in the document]
gettinToasty: isn't it setting it to an opposite state via `!sourceFilter.visible` unless im misreading something here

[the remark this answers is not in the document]
andycreeth: Yeah I somewhat stupidly made the rollback command a toggle, even though the execute command isn't.  I should fix that.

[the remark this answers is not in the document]
gettinToasty: how does one turn visibility off in this case? the parent function still appears to assume this functions as a toggle

app/services/editor-commands/editor-commands.ts
```diff
@@ -0,0 +1,273 @@
...
+            this.handleUndoRedoError(true, e);
+          });
+      }
+
+      this.redoHistory.push(command);
+      this.PUSH_REDO_METADATA({ description: command.description });
+    }
+  }
+
+  @shortcut('Ctrl+Y')
```
holiber: We should add the `Ctrl+Y` hint somewhere in interface. It's not always obvious that `Ctrl+Y` is redo

app/services/editor-commands/commands/copy-nodes.ts
```diff
@@ -0,0 +1,106 @@
...
+  constructor(
+    private selection: Selection,
+    private destSceneId: string,
+    private duplicateSources = false,
+  ) {
+    super();
+  }
+
+  get description() {
+    return `Paste ${this.selection.getNodes()[0].name}`;
```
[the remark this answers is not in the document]
andycreeth: Yeah good call.  I should move this over to the constructor so we generate the the string once and save it.  That's how most of the other commands work.

app/services/editor-commands/editor-commands.ts
andycreeth: Hmm yeah.  I could make them accelerators in the right click menu instead, so they would show up there.  I could also add them to the icon tooltips.

app/components/windows/SourceFilters.vue.ts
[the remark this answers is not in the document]
andycreeth: The parent function was translating the toggle into a set, since the service expects a set, but the UI has it implemented as a toggle.  I just refactored this command to be a toggle instead to better match the UI rather than the service, since commands are meant to reflect user actions.

PR #1720, opened 2019-05-17
Reinstate beaker slider

gettinToasty:


holiber:
Looks like here is the issue with tests retrying. Try to merge this branch with staging

PR #1729, opened 2019-05-21
Add specific workload for Visual Studio build tools

pdamasceno:
Changed the install of Visual Studio Build Tools to install a specific workload and all the recommended packages

### DIFF
--- a/scripts/install-ci.ps1
+++ b/scripts/install-ci.ps1
@@ -34,6 +34,9 @@ choco install yarn
 echo "Install Git for Windows"
 choco install git.install
 
+echo "Install Visual Studio 2017 Build Tools"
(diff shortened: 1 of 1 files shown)


andycreeth:
Will the full install be performed manually?

pdamasceno:
I figured out that in order to use Visual Studio Build tools to build obs-studio-node, I needed to use a specific workload for C++ and install all the recommended packages which include CMake compatibility. Was able to build successfully on Azure. Just need to set fix somethings on our end in the azure-pipelines.yml file.

holiber:
The full install must be performed only with a one-line command `install.ps1 your_azure_token host_user host_password`

PR #1731, opened 2019-05-21
update slobs icon font

andycreeth:
Sorry for the annoying diff.  I copied in the styles from our CDN and ran the stylelint fixer and this was the result.


gettinToasty:
is there a reason we don't just have this hosted externally with a style tag or something btw?

andycreeth:
In the past we've tried to have SLOBS work in a somewhat degraded state for recording when there isn't internet.  This may not be worth it anymore.  But yeah we need a better system as this was a bit painful to update.

PR #1744, opened 2019-05-23
Overlay v1

gettinToasty:
Non-interactive version of overlays featuring recent events and chat


app/components/GameOverlaySettings.tsx
```diff
@@ -0,0 +1,105 @@
...
+              <VFormGroup
+                value={this.enableGameOverlay}
+                onInput={this.setEnableGameOverlay}
+                metadata={metadata.toggle({ title: $t('Enable in-game overlay') })}
+              />
+              {/* <VFormGroup
+              value={this.enablePreview}
+              onInput={this.setEnablePreview}
+              metadata={metadata.toggle({ title: $t('Enable stream preview in overlay') })}
+            /> */}
```
andycreeth: Is this a feature that is going to be added in a future release?

app/components/windows/OverlayWindow.vue
```diff
@@ -0,0 +1,98 @@
+<template>
```
andycreeth: Just curious - Why didn't you use TSX for this component when all others are in TSX?  Was this code that Adrian started?

app/services/game-overlay/index.ts
```diff
@@ -0,0 +1,396 @@
...
+import overlay, { OverlayThreadStatus } from '@streamlabs/game-overlay';
+import { Inject, InitAfter } from 'services/core';
+import { LoginLifecycle, UserService } from 'services/user';
+import { CustomizationService } from 'services/customization';
+import { getPlatformService } from '../platforms';
+import { WindowsService } from '../windows';
+import { PersistentStatefulService } from 'services/core/persistent-stateful-service';
+import { mutation } from 'services/core/stateful-service';
+import { $t } from 'services/i18n';
+import { Polly } from 'aws-sdk';
```
andycreeth: This looks like an auto import mistake.

app/services/game-overlay/index.ts
```diff
@@ -0,0 +1,396 @@
...
+
+const { BrowserWindow } = electron.remote;
+
+interface ICoordinates {
+  x: number;
+  y: number;
+}
```
andycreeth: We have a ambient application type for this (you don't need to import it) called `IVec2` since it's such a common thing.

app/components/GameOverlaySettings.tsx
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  togglePreviewMode() {
+    this.gameOverlayService.setPreviewMode(!this.previewMode);
+  }
+
+  setOverlayTransparency(value: number) {
+    this.overlayTransparency = value;
+    this.gameOverlayService.setOverlayOpacity(value * 100);
+  }
```
andycreeth: What's the rationale for having the local state variable `overlayTransparency`.  It seems like you could just have a getter/setter with no local variable.

app/components/windows/OverlayWindow.vue
andycreeth: Oh - looks like this component isn't used currently?  Assuming this is a relic from when Adrian was working on interactivity?

app/services/game-overlay/index.ts
```diff
@@ -0,0 +1,396 @@
...
+      webPreferences: { nodeIntegration: false, contextIsolation: true, offscreen: true },
+    };
+  }
+
+  createBrowserWindows() {
+    this.overlayWindow = new BrowserWindow({
+      ...this.commonWindowOptions,
+      height: 600,
+      width: 600,
+    });
```
andycreeth: Is this window used by anything?  This seems like dead code unless I'm missing something.

app/services/game-overlay/index.ts
```diff
@@ -0,0 +1,396 @@
...
+    //     x: containerX - 600,
+    //     y: containerY + 300,
+    //     width: 600,
+    //     height: 300,
+    //     // OneOffWindow options
+    //     isFullScreen: true,
+    //     componentName: 'OverlayWindow',
+    //   },
+    //   'overlay',
+    // );
```
andycreeth: IMO let's get rid of all the interactivity stuff that isn't needed in this PR.  Let's use git to track changes we want to add in the future, not comments.

app/services/game-overlay/index.ts
```diff
@@ -0,0 +1,396 @@
...
+      if (!overlayId) return;
+      const pos = this.defaultPosition(key);
+      const size = key === 'chat' ? { width: 300, height: 600 } : { width: 600, height: 300 };
+      this.windows[key].setBounds({ ...pos, ...size });
+      this.previewWindows[key].setBounds({ ...pos, ...size });
+      overlay.setPosition(overlayId, pos.x, pos.y, size.width, size.height);
+    });
+  }
+
+  private defaultPosition(key: string) {
```
andycreeth: This function is a little confusing to me.  Its name and the fact that it returns a value makes me thinking this is simply a function for fetching the default position.  However, it has a side effect of also modifying the state.  IMO mutating the size of the actual window and changing the state should be done in the same place to keep things consistent and easy to maintain.  IMO this seems like a non-ideal separation of concerns.

app/services/game-overlay/index.ts
```diff
@@ -0,0 +1,396 @@
...
+    this.SET_WINDOW_POSITION(key, null);
+    const [containerX, containerY] = this.getWindowContainerStartingPosition();
+    const x = key === 'recentEvents' ? containerX - 600 : containerX;
+    return { x, y: containerY };
+  }
```
andycreeth: It feels like the above 3 functions: `determineStartPosition`, `resetPosition` and `defaultPosition` could be a little clearer if they used newlines to break things up a bit.

app/services/user.ts
```diff
@@ -449,6 +460,55 @@ export class UserService extends PersistentStatefulService<IUserServiceState> {
...
+    }
+
+    return {
+      destroy: async () => {
+        userLoginSubscription.unsubscribe();
+        userLogoutSubscription.unsubscribe();
+        await doDestroy();
+      },
+    } as LoginLifecycle;
+  }
```
andycreeth: This seems pretty nifty.

app/components/GameOverlaySettings.tsx
gettinToasty: yea thats a part of the interactive component

app/components/windows/OverlayWindow.vue
gettinToasty: correct on both accounts

app/services/game-overlay/index.ts
gettinToasty: it's the parent of the other windows

PR #1746, opened 2019-05-23
Update FM version to 0.9.0

glaba13:
Highlighted Release Notes:
- Optimize initialisation synchronous time
- Feature/depth improvement for masks
- Add more internal automated tests for verification
- Fix emitters crashing slobs
- Move/Clean unnecessary data files from release
- Add Landmark stabilization support


andycreeth:
Are you looking to include this in 0.15.0 (next release)?  If so, this PR needs to be against the `preview` branch since we cut that off `staging` this morning.

glaba13:
we can move for 0.16.0 as it doesn't include any critical bug fix or hotfix, but it has lots of changes, so it's better to be tested and verified

PR #1760, opened 2019-05-25
Validate ListInputs

holiber:
When we're switching a video device in the video capture source it's possible to get an unsupported resolution because we don't clear the resolution ListInput from the previous device

### DIFF
--- a/app/components/obs/inputs/ObsInput.ts
+++ b/app/components/obs/inputs/ObsInput.ts
@@ -452,11 +452,12 @@ export function setPropertiesFormData(obsSource: obs.ISource, form: TObsFormData
   updatedFormData.forEach(prop => {
     if (prop.type !== 'OBS_PROPERTY_LIST') return;
     const listProp = prop as IObsListInput<TObsValue>;
+    if (!listProp.options.length) return;
     const optionExists = !!listProp.options.find(option => option.value === listProp.value);
     if (optionExists) return;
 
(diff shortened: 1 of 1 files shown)


app/components/obs/inputs/ObsInput.ts
```diff
@@ -445,6 +445,21 @@ export function setPropertiesFormData(obsSource: obs.ISource, form: TObsFormData
...
+    if (prop.type !== 'OBS_PROPERTY_LIST') return;
+    const listProp = prop as IObsListInput<TObsValue>;
+    if (!listProp.options.length) return;
+    const optionExists = !!listProp.options.find(option => option.value === listProp.value);
+    if (optionExists) return;
+
+    needUpdatePropsAgain = true;
+    listProp.value = listProp.options[0].value;
+  });
+  if (needUpdatePropsAgain) setPropertiesFormData(obsSource, updatedFormData);
```
gettinToasty: is there a reason we do this on base ObsInput and not ObsListInput?

holiber: This is method is not a part of `ObsInput` it's just a utility function in the same file

gettinToasty: i see, in that case i think we should break apart this function it's kinda getting super big

holiber:
The only way to avoid extra IPC call is to implement this fix on the backend.

PR #1769, opened 2019-05-28
Fix slider interval in some cases

gettinToasty:
I think there was an issue with metadata not always being initialized at component load, this should address that


app/components/shared/inputs/SliderInput.vue.ts
```diff
@@ -40,6 +37,7 @@ export default class SliderInput extends BaseInput<number, ISliderMetadata> {
    * @param value The value that will be displayed on the interface.
    */
   updateLocalValue(value: number) {
+    console.log(this.interval, this.options.interval);
```
andycreeth: remove?

PR #1776, opened 2019-05-28
Fix slider constraints

gettinToasty:


app/components/shared/inputs/SliderInput.vue.ts
```diff
@@ -43,8 +43,14 @@ export default class SliderInput extends BaseInput<number, ISliderMetadata> {
     if ((isNaN(parsedValue) && isString(value)) || (isString(value) && value === '')) {
       // preview only, when there is no input or just a negative symbol.
       this.localValue = value.trim() !== '-' ? '' : value;
+    } else if (parsedValue < this.min) {
+      this.localValue = this.min;
+      this.updateValue(this.min);
```
andycreeth: Will this introduce jank while trying to backspace an input with a nonzero minimum value?

gettinToasty: hmm that's potentially possible here yea, can add a debounce like we originally did for NumberInput

PR #1788, opened 2019-05-30
Fix ScenItems reordering

holiber:


app/services/scenes/scene.ts
```diff
@@ -301,22 +301,23 @@ export class Scene {
...
-        .forEach((obsItem, idx) => {
-          if (obsItem.id === item.obsSceneItemId) currentIdx = idx;
-        });
-      this.getObsScene().moveItem(currentIdx, item.getItemIndex());
+  private reconcileNodeOrderWithObs() {
+    const obsScene = this.getObsScene();
+    const destOrder = this.getItems().map(item => item.obsSceneItemId);
+    const currentOrder = this.getObsScene()
+      .getItems()
+      .reverse()
```
andycreeth: This is an in-place reversal.  I think that's ok here, just double checking.

PR #1793, opened 2019-05-30
Expand containing folders when a source is selected

gettinToasty:


gettinToasty:
CC issues here are not new; just moved from one file to another

app/services/selection/selection.ts
```diff
@@ -285,6 +107,15 @@ export class Selection {
     return scene.getNodes().filter(node => this.state.selectedIds.includes(node.id));
   }
 
+  /**
+   * return node specified from a given ID
+   */
+  getNode(id: string): TSceneNode {
+    const scene = this.getScene();
+    if (!this.getSize()) return null;
+    return scene.getNode(id);
```
holiber: This method returns the node even if the selection doesn't have this node. That seems incorrect.
The correct method:
```
getNode(id: string): TSceneNode {
  return this.isSelected(id) ? this.getScene().getNode(id) : null;
}
```

gettinToasty: the reason i did that was i originally thought i would need the nodes of the parents, which it turned out i didn't, but i could see a use-case in the future where we might need a node that isn't currently selected so i left this implementation in. it's also just following the pattern of other similar methods, but if you think there's a reason it shouldn't behave this way let me know and i can change it.

holiber: If we need a node without any condition like it's selected or not we should use `Scene.getNode()`

gettinToasty: but i am using that? unless there is some sort of class method you're referring to that i don't see

gettinToasty:
@holiber correct it doesn't have scrolling, the way we do things currently it'll be a bit tricky to get scrolling working so i thought that would be better for a v2 in a different PR, but seeing as even VSCode doesn't have scrolling i think it's not a blocking omission at all

app/services/selection/selection.ts
holiber: In your `expandSelectedNodes` if you need to fetch any node from the current active scene use:
```
const node = this.selectionService.getScene().getNode(this.lastSelectedId);
// or
const node = this.scenesService.activeScene.getNode(this.lastSelectedId);
```
if you need to fetch a node from selected items use the method I posted above.

The current implementation of `Selection.getNode()` is like "fetch any node from the current activeScene, but only if the current selection has at list one node". Are you sure you need this case? Even if you need this case it seems to be so specific and not reusable that it should be a part of your component, but not a part of the Selection API.

PR #1794, opened 2019-05-30
fix object is destroyed error

andycreeth:
Fixes: https://sentry.io/organizations/streamlabs-obs/issues/980263462

We hit this error when you try to close Streamlabs OBS while it is still initializing.  We have ~3-4k occurrences daily.

### DIFF
--- a/main.js
+++ b/main.js
@@ -254,7 +254,9 @@ function startApp() {
   }
 
   ipcMain.on('services-ready', () => {
-    childWindow.loadURL(`${global.indexUrl}?windowId=child`);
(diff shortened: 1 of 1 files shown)


main.js
```diff
@@ -254,7 +254,9 @@ function startApp() {
   }
 
   ipcMain.on('services-ready', () => {
-    childWindow.loadURL(`${global.indexUrl}?windowId=child`);
+    if (!childWindow.isDestroyed()) {
```
holiber: I would prefer to add a comment here

PR #1797, opened 2019-05-31
More TSX Ports

gettinToasty:


app/components/ConnectionSettings.tsx
```diff
@@ -67,4 +66,51 @@ export default class SceneTransitions extends Vue {
...
+      from: this.fromSceneModel,
+      transition: this.transitionModel,
+      to: this.toSceneModel,
+    }[key];
+  }
+
+  render(h: Function) {
+    return (
+      <div>
+        {['from', 'transition', 'to'].map(key => (
```
andycreeth: I'll be honest, I think this refactor is far more complex and less readable than the original component.

gettinToasty: ¯\_(ツ)_/¯ maybe increase CC threshhold to 35 lines? TBF i think this is the "JSX Way" of doing things more, they really like mapping and such in renders

andycreeth: Yeah could be.  It's been a while since I've written JSX.  IMO this would be far clearer if you just kept the 3 `VFormGroup` tags and removed all the misdirection of functions like `title` and `value` that look up the options.  It could be that I'm just not used to it and will find it clearer in the future, but my gut reaction is that this is over-factoring.  Happy to be overruled though.  Maybe we should see what Alex thinks.

holiber: When you have a loop but you need to literally define the logic for each iteration of this loop than probably you don't need a loop. I prefer to have several `VFormGroup` in the template here.

PR #1802, opened 2019-06-01
Save replay buffer settings when using optimized encoder

holiber:


app/services/settings/settings.ts
```diff
@@ -186,6 +186,7 @@ export class SettingsService extends StatefulService<ISettingsState>
       }
     }
 
+    console.log('get settings form data', settings);
```
andycreeth: remove?

PR #1812, opened 2019-06-05
Fix scene collection events

holiber:


app/services/api/tcp-server/tcp-server.ts
```diff
@@ -382,9 +381,22 @@ export class TcpServerService extends PersistentStatefulService<ITcpServersSetti
     // send event to subscribed clients
     Object.keys(this.clients).forEach(clientId => {
       const client = this.clients[clientId];
+      const eventName = event.result.resourceId.split('.')[1];
+
+      // these events will be sent to the client even if isRequestsHandlingStopped = true
+      // this allows to send this event even if the app is in the loading state
```
andycreeth: Is there any reason not to send events while the app is in a loading state?  We shouldn't handle any requests, but I don't see why we can't send events.

holiber: We delete all sources between switching screen collections. So it causes multiple `removed` events.

PR #1813, opened 2019-06-05
Youtube wording update

trevorrl:
Youtube now refers to "Sponsors" as "Members". Updated verbiage to reflect that. 

Youtube API still refers to `sponsors` so variables were left unchanged, I only updated user-facing text.


andycreeth:
Re-running, but I think this might have broken a test.

PR #1816, opened 2019-06-05
Investigating zombie process in tests

pdamasceno:
This will not be merged


holiber:
I suppose this work is done

PR #1819, opened 2019-06-06
Re-enable flaky tests

holiber:
I found that it's pretty difficult to synchronize the widget settings via listening to websocket events. I just added a little loading indicator and decide that settings are synchronized when you don't see the loading spinner.

goal

For streaming tests, just increasing waiting time for `waitForVisible` resolves flakiness

Test for the video optimization step needs to be rewritten.


test/widgets/goals.ts
```diff
@@ -58,7 +57,6 @@ function testGoal(goalType: string) {
       font: 'Roboto',
     };
 
-    await formMonkey.fill(testSet1);
```
gettinToasty: if we remove this doesn't it kinda defeat the purpose of the test?

holiber: I fill the form in the line below. This line is not needed

PR #1832, opened 2019-06-07
Disallow logged out users from enabling overlay

gettinToasty:


app/services/game-overlay/index.ts
```diff
@@ -228,6 +228,8 @@ export class GameOverlayService extends PersistentStatefulService<GameOverlaySta
   }
 
   async setEnabled(shouldEnable: boolean = true) {
+    if (!this.userService.isLoggedIn()) return;
```
holiber: I prefer to also have a message in the Settings window, something like `Sing in to use Game Overlays`. Otherwise, it looks broken.

app/components/GameOverlaySettings.tsx
```diff
@@ -19,7 +19,16 @@ export default class GameOverlaySettings extends TsxComponent<{}> {
 
   async setEnableGameOverlay(val: boolean) {
     this.enabling = true;
-    await this.gameOverlayService.setEnabled(val);
+    try {
+      await this.gameOverlayService.setEnabled(val);
+    } catch (e) {
```
andycreeth: You should check the type of `e` here.  If `setEnabled` raises an exception this will be an exception.  I'm not sure how that will serialize or what toasted will do with it, but my guess is it will be jank.

PR #1842, opened 2019-06-11
add a few tests for undo and fix the bugs they uncovered

andycreeth:
I'm not going to write exhaustive tests for every single undo command, so I wanted to start with a few tests that that will have as broad a reach as possible.  The commands tested here are extremely complex and utilize lots of other subcommands to get their job done.  So in reality these 3 tests end up testing 10+ commands, so reach is fairly good.

The test to remove a scene ended up surfacing a few bugs, including the sentry error that I was having trouble locating the source of.  Those bugs are fixed as part of this PR.

More tests are coming later.

### DIFF
--- a/app/services/editor-commands/commands/remove-folder.ts
+++ b/app/services/editor-commands/commands/remove-folder.ts
@@ -8,6 +8,7 @@ export class RemoveFolderCommand extends Command {
 
   private name: string;
   private childrenIds: string[];
+  private parentId: string;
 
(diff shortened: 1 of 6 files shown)


holiber:
I'm starting worrying about we add tests with too big execution time. This test adds additional 21 seconds to the entire tests suite. I suggest using slow `await addSource` only when we're testing the `add source` feature. In other cases, we need to use fast API methods

andycreeth:
Fair point to consider.

PR #1853, opened 2019-06-13
Add partition to the chat browser window

gettinToasty:


app/services/windows.ts
```diff
@@ -104,7 +104,28 @@ export function getComponents() {
   };
 }
 
-export interface IWindowOptions {
+export interface IElectronWindowOptions {
```
andycreeth: You should be able to use `Electron.BrowserWindowConstructorOptions` instead of defining this yourself.

PR #1859, opened 2019-06-17
Add login check to outage notifications

gettinToasty:


holiber:
Hmm, I thought it affected only new tests and already added a fix here, lol https://github.com/stream-labs/streamlabs-obs/pull/1858/files#diff-af690875e4e36d80fbd07fa20bab4d63R82

PR #1866, opened 2019-06-18
Redesign Add Source flow

gettinToasty:
Place existing sources higher in hierarchy so we discourage people from adding too many new sources
addsource

### DIFF
--- a/test/filters.js
+++ b/test/filters.js
@@ -36,7 +36,7 @@ test('Adding and removing a Color Correction filter', async t => {
 
 test('Adding and removing a Image Mask filter', async t => {
   const app = t.context.app;
-  const sourceName = 'Color Source';
+  const sourceName = 'Color Source 2';
   const filterName = 'Image Mask/Blend';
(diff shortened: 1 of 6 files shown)


holiber:
This toggle button is not correctly centered. The text baseline should be on the button's text level

test/filters.js
```diff
@@ -60,7 +60,7 @@ test('Adding and removing a Image Mask filter', async t => {
 
 test('Adding and removing a Crop Pad filter', async t => {
   const app = t.context.app;
-  const sourceName = 'Color Source';
+  const sourceName = 'Color Source 3';
```
holiber: Why this change is needed?

gettinToasty:
@holiber i fixed the toggle centering in the actual branch, the preview gif was just made before i did that

test/filters.js
gettinToasty: just easy logic to force the creation of a new source

PR #1903, opened 2019-07-04
add autohide option for overlays

summeroff:


app/components/GameOverlaySettings.tsx
```diff
@@ -61,6 +67,16 @@ export default class GameOverlaySettings extends TsxComponent<{}> {
     });
   }
 
+  get sliderAutohideMetadata() {
+    return metadata.slider({
+      title: $t('Overlay autohide timeout'),
```
gettinToasty: maybe something like `Fade Time (s)` and `description: $t('Set to 0 to always appear on screen')`

PR #1906, opened 2019-07-08
Add performance service to external api

SidneyBovet:
This simply wraps services/performance and gives external apps access to live CPU usage, framerate, dropped frames, etc.

This was inspired from the implementation of `services/api/external-api/scenes/scenes.ts`.

**I am fairly new to typescript and Node, so please do review carefully what I propose here.** I tested the API using https://stream-labs.github.io/streamlabs-obs-api-docs/ and it works. I also ran a private YouTube stream without issue.

### DIFF
--- a/app/services/api/external-api/performance/index.ts
+++ b/app/services/api/external-api/performance/index.ts
@@ -1,0 +1,1 @@
+export * from './performance';
--- a/app/services/api/external-api/performance/performance.ts
+++ b/app/services/api/external-api/performance/performance.ts
@@ -1,0 +1,31 @@
+import { Singleton, Fallback } from 'services/api/external-api';
+import { Inject } from 'services/core/injector';
+import { PerformanceService as InternalPerformanceService } from 'services/performance';
+
+interface IPerformanceState {
+  CPU: number;
+  numberDroppedFrames: number;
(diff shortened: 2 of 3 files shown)


andycreeth:
Hey there, thanks for this submission!  Our developer who is our expert on the external API is on vacation this week.  He'll be back on Monday and I'll get him to look this over.  Thanks!

SidneyBovet:
Very good point, these values will be used in batch anyway, so this save quite a bit of traffic.

I also made `IPerformanceState ` in app/services/performance.ts export so we have access to the same data structure from the external api (easier to maintain), but I don't know if this causes any other issue.

app/services/performance.ts
```diff
@@ -5,7 +5,7 @@ import { StatefulService, mutation } from './core/stateful-service';
 import electron from 'electron';
 import * as obs from '../../obs-api';
 
-interface IPerformanceState {
+export interface IPerformanceState {
```
holiber: Please create a copy of this interface in the external-api/performance/performance.ts

The interface in the internal API may be changed in the future and it should not affect the external API. So external API should not expose interfaces from the internal API.

SidneyBovet: Makes sense, thanks!

PR #1909, opened 2019-07-08
Prevent announcement banner for new users under 1 week

kuhasuki:
The timestamp stored for firstActivationTimestamp won't be set until announcements service updateBanner method is run. This means that the value may not exactly match the install day. In my testing, it seems like announcement service is only called if SLOBS is opened while already logged in. I could look at setting this value in another service which would guarantee it is written on first activation after installing.

This is meant to be the most simple solution, happy to discuss ways of improving it.

### DIFF
--- a/app/services/announcements.ts
+++ b/app/services/announcements.ts
@@ -60,7 +60,7 @@ export class AnnouncementsService extends StatefulService<IAnnouncementsInfo> {
   }
 
   underOneWeek() {
-    return (Date.now() - this.activationTimestamp) / (1000 * 60 * 60 * 24 * 7) < 1;
+    return Date.now() - this.activationTimestamp < 1000 * 60 * 60 * 24 * 7;
   }
 
   private async fetchBanner() {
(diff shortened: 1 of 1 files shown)


app/services/announcements.ts
```diff
@@ -45,8 +47,24 @@ export class AnnouncementsService extends StatefulService<IAnnouncementsInfo> {
...
+    this.activationTimestamp = Date.now();
+    return this.activationTimestamp;
+  }
+
+  set activationTimestamp(timestamp: number) {
+    localStorage.setItem(this.localStorageKey, timestamp.toString());
+  }
+
+  underOneWeek() {
+    return (Date.now() - this.activationTimestamp) / (1000 * 60 * 60 * 24 * 7) < 1;
```
andycreeth: Why not just
```TypeScript
return (Date.now() - this.activationTimestamp) < (1000 * 60 * 60 * 24 * 7);
````
Doesn't really matter, just curious why the extra division before the comparison.

kuhasuki: Hmm good point. Will update.

kuhasuki:
Refactored this to check the creation date for log.log file in slobs-client directory to prevent issues with the prior implementation, namely disabling the announcements banner for all users for 1 week.

PR #1919, opened 2019-07-10
Update Goals for Facebook

kuhasuki:
1. Adds 2 new widgets for FB users: supporter goal and stars goal
2. Updates endpoints used for managing goals on SL API
3. Removes `goalResetUrl`, issuing DELETE request to `goalUrl` will suffice now

@karljakober


app/services/widgets/settings/donation-goal.ts
```diff
@@ -12,10 +12,9 @@ export class DonationGoalService extends GenericGoalService {
       type: WidgetType.DonationGoal,
       url: `https://${this.getHost()}/widgets/donation-goal?token=${this.getWidgetToken()}`,
       previewUrl: `https://${this.getHost()}/widgets/donation-goal?token=${this.getWidgetToken()}`,
-      dataFetchUrl: `https://${this.getHost()}/api/v5/slobs/widget/donationgoal`,
-      settingsSaveUrl: `https://${this.getHost()}/api/v5/slobs/widget/donationgoal`,
-      goalUrl: `https://${this.getHost()}/api/v5/slobs/donation/goal`,
-      goalResetUrl: `https://${this.getHost()}/api/v5/slobs/donation/goal/end`,
+      dataFetchUrl: `https://${this.getHost()}/api/v5/slobs/widget/donationgoal/settings/new`,
+      settingsSaveUrl: `https://${this.getHost()}/api/v5/slobs/widget/donationgoal/settings/new`,
```
gettinToasty: is there hope of resolving this on the site side or is this our lives now

kuhasuki: The problem is that we had it backwards for donation goal endpoint and were already calling an API that did exactly the opposite of what each other goal does. Adding new was chosen as the least bad option which ensures consistency. Depending on how many people continue using <0.16 we could eventually remove all the misformatted legacy endpoints.

PR #1925, opened 2019-07-11
Onboarding 2: The Squeakquel

gettinToasty:


app/components/pages/Onboarding.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+  currentStep = 1;
+  importedFromObs = false;
+  processing = false;
+  fbSetupEnabled: boolean = null;
+
+  stepsState = [{ complete: false }, { complete: false }, { complete: false }];
+
+  checkFbPageEnabled() {
+    if (this.fbSetupEnabled === null || !this.userService.isLoggedIn()) return;
```
andycreeth: Maybe I'm reading this wrong, but I don't see how this can ever become not-null if this function returns here before fetching the value.

app/components/pages/Onboarding.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+  proceed() {
+    if (this.processing) return;
+    if (this.currentStep === this.stepsState.length) return this.complete();
+
+    this.stepsState[this.currentStep - 1].complete = true;
+    this.currentStep = this.currentStep + 1;
+
+    if (this.currentStep === this.stepsState.length) {
+      this.stepsState[this.currentStep - 1].complete = true;
```
andycreeth: Lol - Might have been simpler to make `currentStep` 0-indexed.

app/components/pages/onboarding-steps/Connect.tsx
```diff
@@ -1,42 +1,41 @@
...
 import { Inject } from 'services/core/injector';
 import { OnboardingService } from 'services/onboarding';
-import electron from 'electron';
+import TsxComponent from 'components/tsx-component';
 import { $t } from 'services/i18n';
+import styles from './Connect.m.less';
 
 @Component({})
-export default class Connect extends Vue {
+export default class Connect extends TsxComponent<{ continue: Function }> {
```
andycreeth: It's best to be explicit about the function signature
```TypeScript
{ continue: () => void }
```

app/components/pages/onboarding-steps/Connect.tsx
```diff
@@ -1,42 +1,41 @@
...
   @Inject() userService: UserService;
   @Inject() onboardingService: OnboardingService;
 
+  @Prop() continue: Function;
```
andycreeth: And here

app/components/pages/onboarding-steps/StreamlabsFeatures.tsx
```diff
@@ -0,0 +1,101 @@
...
+  @Inject() scenesService: ScenesService;
+
+  get hasExternalMonitor() {
+    return electron.remote.screen
+      .getAllDisplays()
+      .find(display => display.bounds.x !== 0 || display.bounds.y !== 0);
+  }
+
+  get isIrlStreamer() {
+    return this.scenesService.scenes.find(scene => /irl/i.test(scene.name));
```
andycreeth: nice

app/components/pages/Onboarding.tsx
gettinToasty: yea thats supposed to be !== null good catch

app/components/pages/Onboarding.tsx
gettinToasty: the beaker component uses a 1-indexed system so i just went with that :P

PR #1931, opened 2019-07-16
Refactor all webviews to browser views

andycreeth:
Added a helper component that makes it behave a bit more like a webview in the DOM.


app/components/pages/BrowseOverlays.vue.ts
```diff
@@ -28,34 +26,31 @@ export default class BrowseOverlays extends Vue {
...
+      const protocol = urlLib.parse(url).protocol;
 
       if (protocol === 'http:' || protocol === 'https:') {
-        electron.remote.shell.openExternal(e.url);
+        electron.remote.shell.openExternal(url);
       }
     });
 
-    I18nService.setWebviewLocale(this.$refs.overlaysWebview);
+    I18nService.setBrowserViewLocale(view);
```
gettinToasty: is there a reason you didn't just use the `setLocale` option here?

andycreeth: Good catch - I added the `setLocale` prop the second time I needed it, and forgot to refactor the initial place.

PR #1938, opened 2019-07-19
proper typing of return values for editor commands

andycreeth:


app/services/editor-commands/editor-commands.ts
```diff
@@ -63,7 +63,7 @@ export class EditorCommandsService extends StatefulService<IEditorCommandsServic
   executeCommand<TCommand extends keyof typeof COMMANDS>(
     commandType: TCommand,
     ...commandArgs: ConstructorParameters<(typeof COMMANDS)[TCommand]>
-  ) {
+  ): ReturnType<InstanceType<(typeof COMMANDS)[TCommand]>['execute']> {
```
holiber: Wow, TS magic

andycreeth: I know right haha.  Took a while to find the secret TS formula.

PR #1951, opened 2019-07-29
Fix source releasing on shutdown

holiber:
Destroying the child on the shutdown window didn't release the projector's source in the OSN. Closing the child window before shutdown resolves this issue.

### DIFF
--- a/main.js
+++ b/main.js
@@ -199,7 +199,6 @@ if (!gotTheLock) {
     mainWindow.on('close', e => {
       if (!shutdownStarted) {
         shutdownStarted = true;
-        childWindow.destroy();
         mainWindow.send('shutdown');
 
         // We give the main window 10 seconds to acknowledge a request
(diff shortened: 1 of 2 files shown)


main.js
```diff
@@ -199,7 +199,6 @@ if (!gotTheLock) {
     mainWindow.on('close', e => {
       if (!shutdownStarted) {
         shutdownStarted = true;
-        childWindow.destroy();
```
andycreeth: I'm confused how the `childWindow` ever gets destroyed now.  I see you added `this.windowsService.closeChildWindow()` to the shutdown handler above, but that doesn't actually properly close/destroy the window, it just hides it.  Unless I'm missing something, I don't think the application will cleanly exit now.

holiber: Do we really need to destroy it? Didn't have any issues with closing the app

andycreeth: Electron will not exit until all windows are closed.  Just because all windows are gone that doesn't mean it's not still running in the background.

holiber: Hmm. The child window is destroying even without this line. But don't see where we can do this.

andycreeth: Interesting.  I'm not sure either.

andycreeth: Yeah I tested it a bit and it seems fine.  Think we're probably ok here.

PR #1962, opened 2019-07-31
Test the EditStream window for network errors

holiber:


test/helpers/spectron/network.ts
```diff
@@ -0,0 +1,29 @@
+import { focusMain, TExecutionContext } from './index';
+
+/**
+ * Replaces the `fetch` method to simulate network issues
+ */
+export async function setResponseCode(t: TExecutionContext, code: number) {
```
andycreeth: This is cool, that said - I'm a little worried this might make tests brittle.  We have lots of web requests that happen on interval timers.  I think this network request mocking would be more robust if you were required to pass in a string or regex, and it will only mock the response if the URL matches what was passed in.  If it doesn't get a match, it would just delegate to the original fetch method.

andycreeth:
Seems like tests are failing

PR #1965, opened 2019-08-01
Move TopNav to SideNav

gettinToasty:


test/themes.ts
```diff
@@ -15,8 +15,8 @@ test('Installing a theme', async (t: any) => {
 
   await logIn(t);
 
-  await app.client.waitForExist('.top-nav.loading', 5000, true);
-  await app.client.click('button=Themes');
+  // await app.client.waitForExist('button=themes', 5000, true);
```
holiber: Seems like it's gonna be flaky

test/undo.ts
```diff
@@ -14,11 +14,15 @@ import { sleep } from './helpers/sleep';
 useSpectron();
 
 async function undo(t: TExecutionContext) {
-  await t.context.app.client.click('.fa-undo');
+  t.context.app.client.keys(['Control', 'z']);
+  t.context.app.client.keys('Control');
+  await sleep(500);
```
holiber: add await for sending keys instead of sleep
```
await t.context.app.client.keys(['Control', 'z']);
```

test/undo.ts
```diff
@@ -14,11 +14,15 @@ import { sleep } from './helpers/sleep';
...
 }
 
 async function redo(t: TExecutionContext) {
-  await t.context.app.client.click('.fa-redo');
+  t.context.app.client.keys(['Control', 'y']);
+  t.context.app.client.keys('Control');
+  await sleep(500);
```
holiber: same as above

test/themes.ts
gettinToasty: how so?

test/undo.ts
gettinToasty: keys doesn't return a promise it cant be awaited

test/themes.ts
holiber: Just reviewed the code. We have updated `logIn(t)` method now. So this is not gonna be an issue.

test/undo.ts
holiber: Just looked into the source code. It returns promise but seems like d.ts files are incorrect.
Try to use a typecasting here
```
await ((this.client.keys(['Control', 'z']) as any) as Promise<any>); 
```

PR #1986, opened 2019-08-07
Tweet when going live

kuhasuki:
Allows logged in users to compose a tweet and have it automatically posted when going live.

Functionality added:
1. Ability to link a Twitter account
2. Ability to unlink a Twitter account
3. Tweet when going live
4. Crudely automated tweet composer
5. Deleted trailing period in go live update error section


app/services/integrations/twitter.ts
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  setTweetPreference(preference: boolean) {
+    this.SET_TWEET_PREFERENCE(preference);
+  }
+
+  private linkTwitterUrl() {
+    const host = Util.isPreview() ? this.hostsService.beta3 : this.hostsService.streamlabs;
+    const token = this.apiToken;
+    const i18nService = I18nService.instance as I18nService; // TODO: replace with getResource('I18nService')
```
holiber: Just inject I18nService service into the current service

app/services/integrations/twitter.ts
```diff
@@ -0,0 +1,156 @@
...
+    const locale = i18nService.state.locale;
+
+    return `https://${host}/slobs/twitter/link?oauth_token=${token}&l=${locale}`;
```
holiber: Why we need to use `authorizedHeaders(this.apiToken)` in the `unlinkTwitter` method, but use the GET parameter here?

holiber:
I think we should reduce this space 
image

app/services/integrations/twitter.ts
kuhasuki: Look at the implementation for dashboard, app store, themes. We are opening auth window on streamlabs.com which is redirected to twitter.com and back. Current implementation of linking twitter gets user info from login session not from a request header.

app/services/integrations/twitter.ts
kuhasuki: Done

kuhasuki:
> I think we should reduce this space
> image

The section for composing a tweet is a bit taller than the unlinked view. Some room at the bottom is necessary for the updateError prompt

app/services/integrations/twitter.ts
kuhasuki: But I did discuss this approach for reducing number of logins required with Salman, Andy, Lohith and we agree that there should be a better way to handle creating authenticated webviews. I will explore the options in the future

app/components/Twitter.tsx
```diff
@@ -0,0 +1,210 @@
...
+  midStreamMode: boolean;
+  updatingInfo: boolean;
+  value: string;
+}> {
+  @Inject() twitterService: TwitterService;
+  @Inject() userService: UserService;
+
+  @Prop() streamTitle: string;
+  @Prop() midStreamMode: boolean;
+  @Prop() updatingInfo: boolean;
```
gettinToasty: looks like you're missing the `value` prop here, unless that's unused in which case you can remove it from the type declaration above

app/components/Twitter.tsx
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  @Watch('siteUrl')
+  onSiteUrlChange() {
+    this.setInitialTweetBody();
+  }
+
+  @Watch('streamTitle')
+  onTitleUpdate(item: string) {
+    if (this.tweetModel.indexOf(this.priorTitle) !== -1 && this.tweetModel.length < 280) {
```
gettinToasty: i'm not sure why you're checking character length before replacement but not after here?

app/components/windows/EditStreamInfo.vue
```diff
@@ -86,6 +86,12 @@
             :metadata="{ title: $t('Do not show this message when going live') }"
           />
         </h-form-group>
+        <Twitter
+          :streamTitle="channelInfo.title"
+          :midStreamMode="midStreamMode"
+          :updatingInfo="updatingInfo"
+          v-model="tweetModel"
```
gettinToasty: looks like this v-model should no longer be here?

app/components/Twitter.tsx
```diff
@@ -0,0 +1,210 @@
...
+    this.tweetModel = tweet;
+  }
+
+  updateShouldTweetModel(shouldTweet: boolean) {
+    this.shouldTweetModel = shouldTweet;
+  }
+
+  @Watch('tweetModel')
+  onTweetChange(tweet: string) {
+    this.$emit('input', tweet);
```
gettinToasty: wondering why you're having an internal tweet model with a watcher instead of just emitting the changes

app/components/windows/EditStreamInfo.vue.ts
```diff
@@ -341,6 +380,10 @@ export default class EditStreamInfo extends Vue {
     return this.streamInfoService.state.error;
   }
 
+  tweet() {
+    this.twitterService.postTweet(this.tweetModel);
+  }
+
```
gettinToasty: this looks unused

app/services/streaming/streaming.ts
```diff
@@ -262,7 +262,7 @@ export class StreamingService extends StatefulService<IStreamingServiceState>
       queryParams: {},
       size: {
         width: 600,
-        height: 550,
+        height: 620,
```
gettinToasty: shouldn't the height of this depend on if the twitter component is visible?

kuhasuki: The twitter component is always visible if you're logged in. If you're not logged in the edit stream info window doesn't show.

gettinToasty: kk wasn't sure about that, thanks for clarifying

app/components/windows/EditStreamInfo.vue.ts
kuhasuki: yup deleted

app/components/Twitter.tsx
kuhasuki: Changed to passing strings around

app/components/Twitter.tsx
kuhasuki: Changed it to check character count after modification is made

app/components/Twitter.tsx
kuhasuki: Got rid of tweetModel and used `value` prop where I need it

PR #1987, opened 2019-08-08
Refactor stats to appear on hover

gettinToasty:


holiber:
Why you are going to hide performance stats event if SLOBS window has enough space to show it?

gettinToasty:
@holiber there are a couple reasons for this, the design discussion may have happened while you were out of town. basic idea is simplifying, removing clutter, and taking away focus on perceived performance losses when it actually doesn't affect their machine.

PR #1988, opened 2019-08-09
Refactor requiresToken code

holiber:
Replace a fragile `@requiresToken` decorator  with a more robust `platformRequest` method.

### DIFF
--- a/app/services/platforms/facebook.ts
+++ b/app/services/platforms/facebook.ts
@@ -110,7 +110,7 @@ export class FacebookService extends StatefulService<IFacebookServiceState>
   }
 
   getHeaders(req: IPlatformRequest, useToken: boolean | string) {
-    const token = typeof useToken === 'string' ? useToken : this.oauthToken;
+    const token = typeof useToken === 'string' ? useToken : useToken && this.oauthToken;
     return {
       'Content-Type': 'application/json',
       ...(token ? { Authorization: `Bearer ${token}` } : {}),
--- a/app/services/platforms/facebook.ts
+++ b/app/services/platforms/facebook.ts
@@ -6,13 +6,14 @@ import {
   TPlatformCapability,
   TPlatformCapabilityMap,
   EPlatformCallResult,
+  IPlatformRequest,
 } from '.';
 import { HostsService } from '../hosts';
 import { SettingsService } from '../settings';
 import { Inject } from '../core/injector';
 import { authorizedHeaders, handleResponse } from '../../util/requests';
 import { UserService } from '../user';
-import { handlePlatformResponse, requiresToken } from './utils';
+import { handlePlatformResponse, platformAuthorizedRequest, platformRequest } from './utils';
 import { IListOption } from '../../components/shared/inputs';
 import { $t } from 'services/i18n';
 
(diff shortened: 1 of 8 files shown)


app/services/platforms/facebook.ts
```diff
@@ -108,15 +109,16 @@ export class FacebookService extends StatefulService<IFacebookServiceState>
     return this.state.activePage.access_token;
   }
 
-  getHeaders(token = this.oauthToken): Headers {
-    const headers = new Headers();
-    headers.append('Content-Type', 'application/json');
-    headers.append('Authorization', `Bearer ${token}`);
-    return headers;
+  getHeaders(req: IPlatformRequest, useToken: boolean | string) {
+    const token = typeof useToken === 'string' ? useToken : this.oauthToken;
```
gettinToasty: i'm not sure why we have `useToken` as a boolean or string here since it seems like we never actually check the boolean value of it, why can't it just be an optional param as a string or a default argument like before?

app/services/platforms/facebook.ts
```diff
@@ -108,15 +109,16 @@ export class FacebookService extends StatefulService<IFacebookServiceState>
     return this.state.activePage.access_token;
   }
 
-  getHeaders(token = this.oauthToken): Headers {
-    const headers = new Headers();
-    headers.append('Content-Type', 'application/json');
-    headers.append('Authorization', `Bearer ${token}`);
-    return headers;
+  getHeaders(req: IPlatformRequest, useToken: boolean | string) {
```
gettinToasty: not sure why request is bring sent in here-- we dont seem to use it at all?

app/services/platforms/facebook.ts
```diff
@@ -261,29 +263,23 @@ export class FacebookService extends StatefulService<IFacebookServiceState>
...
-        .then(handlePlatformResponse)
-        .then(() => true);
+      return platformRequest(
+        {
+          url: `${this.apiBase}/${this.state.liveVideoId}`,
+          method: 'POST',
+          body: JSON.stringify({ title, description, game_specs: { name: game } }),
+        },
+        this.state.activePage.access_token,
+      ).then(() => true);
```
gettinToasty: we don't seem to be attaching headers to this request

app/services/platforms/facebook.ts
```diff
@@ -261,29 +263,23 @@ export class FacebookService extends StatefulService<IFacebookServiceState>
...
     if (searchString.length < 2) return;
-    const url = `${this.apiBase}/v3.2/search?type=game&q=${searchString}`;
-    const headers = this.getHeaders();
-    const request = new Request(url, { headers, method: 'GET' });
-    return fetch(request)
-      .then(handlePlatformResponse)
-      .then((json: any) => json.data);
+    return platformAuthorizedRequest(
+      `${this.apiBase}/v3.2/search?type=game&q=${searchString}`,
+    ).then(json => json.data);
```
gettinToasty: it looks like we're using both old and new request formations here? i feel like it'd be better for them to be uniform across the service

app/services/platforms/facebook.ts
holiber: It's added for compatibility with `IPlatformService`. For twitch API, depending on url, we use different authorization headers.

app/services/platforms/facebook.ts
holiber: If you call `platformAuthorizedRequest` method you will get `useToken=true` as an argument. So the default value will be not applied.

app/services/platforms/facebook.ts
holiber: Headers will be attached in the `platformRequest` method

app/services/platforms/facebook.ts
gettinToasty: it will be though since you never check the truthiness of useToken, just if it's a string

given that `useToken === true` then `typeof useToken === 'string'` will be falsey and `token` will be the default value of `this.oauthToken`

app/services/platforms/facebook.ts
holiber: I refactored only methods with @requiredToken decorator. But yeah, it's better to refactor all methods

PR #1990, opened 2019-08-09
Fix store sync

holiber:
Looking on inconsistent goals screentests I found that it's possible to get in the situation where the Main and Child window has a different state. I found a flaw in the mutation skipping mechanism. 

image

### DIFF
--- a/app/store/index.ts
+++ b/app/store/index.ts
@@ -28,6 +28,7 @@ const actions = {};
 
 const plugins: any[] = [];
 
+let mutationId = 1;
 let makeStoreReady: Function;
 let storeCanReceiveMutations = Util.isMainWindow();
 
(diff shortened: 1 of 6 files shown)


app/store/index.ts
```diff
@@ -28,6 +28,7 @@ const actions = {};
 
 const plugins: any[] = [];
 
+let mutationId = 1;
```
gettinToasty: is there a reason we want to store IDs like this instead of something like generating a uuid?

holiber: Is there a reason we want to use uuid instead of much faster integer incrementation? This id is needed only per application life, we don't store it in configs.

gettinToasty: no reason specifically was just wondering since it was the first time i saw this pattern

test/screentest/tests/widgets/goals.ts
```diff
@@ -3,6 +3,7 @@ import { logIn, logOut } from '../../../helpers/spectron/user';
 import { makeScreenshots, useScreentest } from '../../screenshoter';
 import { FormMonkey } from '../../../helpers/form-monkey';
 import { addWidget, EWidgetType, waitForWidgetSettingsSync } from '../../../helpers/widget-helpers';
+import moment = require('moment');
```
andycreeth: What's with the weird import?

PR #1997, opened 2019-08-13
Various nav fixes

gettinToasty:
- Make popout metric window not resizable
- Hide app nav arrow if not enough apps

### DIFF
--- a/app/components/AppsNav.tsx
+++ b/app/components/AppsNav.tsx
@@ -23,7 +23,11 @@ export default class AppsNav extends Vue {
   };
 
   upArrowVisible = false;
-  downArrowVisible = true;
+  downArrowVisible = false;
+
+  mounted() {
+    this.handleScroll();
+  }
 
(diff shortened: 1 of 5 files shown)


app/components/AppsNav.tsx
```diff
@@ -86,7 +90,7 @@ export default class AppsNav extends Vue {
     }
     if (el.scrollHeight - el.scrollTop === el.clientHeight) {
       this.downArrowVisible = false;
-    } else {
+    } else if (el.scrollHeight > el.clientHeight) {
```
andycreeth: I think this means the arrow won't appear if you resize the window smaller.

gettinToasty: this is true, but if you start scrolling it'll appear immediately. a resize handler is a little tricky here since it's like the 3rd or 4th child component down from the main window

andycreeth: Seems acceptable then.

PR #2003, opened 2019-08-14
Change cache delete warning string

kuhasuki:


andycreeth:
Hmm this wasn't actually the string I was referring to.  I was referring to this one.  https://github.com/stream-labs/streamlabs-obs/blob/staging/app/components/ExtraSettings.vue#L8

However the one you found also needs to be updated.

`WARNING! You will lose all scenes, sources, and settings. This cannot be undone!` => `WARNING! You will lose all encoder settings.  If you are logged in, your scenes and sources will be restored from the cloud.  This cannot be undone.`

`If you are experiencing weird behavior, you can try deleting your cache directory. This will result in you losing your scene configuration and settings, but can fix some stability issues.` => `Deleting your cache directory will cause you to lose some settings. Do not delete your cache directory unless instructed to do so by a Streamlabs staff member.`

PR #2010, opened 2019-08-16
Refactor youtube service

holiber:
Streaming to youtube was always very flaky to me if I authorized SLOBS via youtube. However, everything was fine if I just manually insert the stream key in the settings window.  I investigated youtube API and realized we do some things wrong. As a result:
- we can populate and change channel info only if the stream is in the active state
- sometimes we cache incorrect streamId and fetch an incorrect steamKey. As result, nobody can see the stream on youtube
- we do a redundant amount of API calls. In some tests I tried to write for youtube I saw a message that requests limit is exceeded
- the stream scheduling feature created multiple events in the Youtube channel and didn't connect these events with a real stream when it began.

### DIFF
--- a/app/services/platforms/youtube.ts
+++ b/app/services/platforms/youtube.ts
@@ -10,16 +10,82 @@ import {
 import { HostsService } from '../hosts';
 import { SettingsService } from '../settings';
 import { Inject } from '../core/injector';
-import { authorizedHeaders } from '../../util/requests';
+import { authorizedHeaders, handleResponse } from '../../util/requests';
 import { UserService } from '../user';
 import { $t } from 'services/i18n';
-import { handlePlatformResponse, platformAuthorizedRequest, platformRequest } from './utils';
+import { platformAuthorizedRequest, platformRequest } from './utils';
 
 interface IYoutubeServiceState {
   liveStreamingEnabled: boolean;
   liveStreamId: string;
   scheduledStartTime: string;
-  channelInfo: IChannelInfo;
+  liveBroadcast: IYoutubeLiveBroadcast;
+}
+
+/**
+ * Represents an API response with a paginated collection
+ */
+interface IYoutubeCollection<T> {
+  items: T[];
+  pageInfo: { totalResults: number; resultsPerPage: number };
+}
+
+/**
+ * A liveBroadcast resource represents an event that will be streamed, via live video, on YouTube.
+ * For the full set of available fields:
+ * @see https://google-developers.appspot.com/youtube/v3/live/docs/liveBroadcasts
+ */
+interface IYoutubeLiveBroadcast {
+  id: string;
+  contentDetails: {
+    boundStreamId: string;
+    enableAutoStart: boolean;
+  };
+  snippet: {
+    title: string;
+    description: string;
+    scheduledStartTime: string;
+    isDefaultBroadcast: boolean;
+    liveChatId: string;
+  };
+  status: {
+    lifeCycleStatus:
+      | 'complete'
+      | 'created'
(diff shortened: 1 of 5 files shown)


app/services/platforms/youtube.ts
```diff
@@ -10,16 +10,82 @@ import {
...
+      /**
+       * streamName is actually a secret stream key
+       */
+      streamName: string;
+      ingestionAddress: string;
+      backupIngestionAddress: string;
+    };
+    resolution: string;
+    frameRate: string;
+  };
```
gettinToasty: thanks for the documentation, but i'm wondering if we should be including every field here or just the ones that are relevant to us

app/services/platforms/youtube.ts
```diff
@@ -117,109 +183,62 @@ export class YoutubeService extends StatefulService<IYoutubeServiceState>
...
-    return platformRequest(`${this.apiBase}/${endpoint}&access_token=${this.oauthToken}`).then(
-      json => {
-        if (json.items.length) {
-          this.SET_STREAM_ID(json.items[0].id);
-        }
-      },
-    );
+    )
+      .then(streams => streams.items[0].cdn.ingestionInfo.streamName)
+      .catch(e => $t('Please enable account for streaming to recieve a stream key'));
```
gettinToasty: what if they error out for other reasons?

app/services/platforms/youtube.ts
```diff
@@ -117,109 +183,62 @@ export class YoutubeService extends StatefulService<IYoutubeServiceState>
...
+        resp.json().then((json: any) => {
+          if (
+            json.error &&
+            json.error.errors &&
+            json.error.errors[0].reason === 'liveStreamingNotEnabled'
+          ) {
+            this.SET_ENABLED_STATUS(false);
+          }
+        });
+        return null;
```
gettinToasty: not sure why we're not throwing the enable error here but we are above

app/services/platforms/youtube.ts
```diff
@@ -229,39 +248,37 @@ export class YoutubeService extends StatefulService<IYoutubeServiceState>
...
+      const data: Dictionary<any> = {
+        id: this.state.liveBroadcast.id,
+        snippet: {
+          title,
+          description: fullDescription,
+        },
+        status: { privacyStatus: 'public' },
+      };
+      if (scheduledStartTime) {
+        data.snippet.scheduledStartTime = scheduledStartTime;
```
gettinToasty: not sure why this is in a separate part of the code instead of assigning it in the object above

holiber: `scheduledStartTime`  is an optional parameter. If you provide this param than Youtube considers this stream as scheduled. We don't need to provide `new Date().toISOString()` as we did before for the regular start streaming flow

app/services/platforms/youtube.ts
holiber: I tried to list only properties that we in some way use in our frontend or backend code.

app/services/platforms/youtube.ts
gettinToasty: right but if the param is option when it's not passed in it'll be `undefined` and having a key be undefined in JS is the same as it not existing

holiber: I see what you mean. That that will be omitted in JSON.stringify. That makes sense so

app/services/platforms/youtube.ts
holiber: The next request after `fetchStreamKey` will be `repopulateStreamInfo`. We handle errors better there

PR #2012, opened 2019-08-16
Show recording time

holiber:
image

### DIFF
--- a/app/components/StudioFooter.vue.ts
+++ b/app/components/StudioFooter.vue.ts
@@ -37,15 +37,20 @@ export default class StudioFooterComponent extends Vue {
 
   metricsShown = false;
   recordingTime = '';
+  private recordingTimeIntervalId: number;
 
   mounted() {
     this.confirmYoutubeEnabled();
 
     // update recording time
-    setInterval(() => {
+    this.recordingTimeIntervalId = window.setInterval(() => {
       if (!this.streamingService.isRecording) return;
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -131,6 +131,7 @@
     "lodash-decorators": "4.3.1",
     "mime": "2.4.3",
     "moment": "2.22.2",
+    "moment-duration-format": "^2.3.2",
```
andycreeth: I don't think we really need to add a new library to the codebase.  The streaming service already has a function that does this called `formattedDurationSince`.  Look at how we do this for streaming timer.

holiber: It's actually just a plugin for moment.js

Will use `formattedDurationSince`

app/components/StudioFooter.vue.ts
```diff
@@ -35,9 +36,16 @@ export default class StudioFooterComponent extends Vue {
...
+  recordingTime = '';
 
   mounted() {
     this.confirmYoutubeEnabled();
+
+    // update recording time
+    setInterval(() => {
+      if (!this.streamingService.isRecording) return;
+      this.recordingTime = this.streamingService.formattedDurationInCurrentRecordingState;
+    }, 1000);
```
andycreeth: You should clear this interval in the `destroyed` hook.  I know the footer is unlikely to ever unmount, but it's good practice to ensure we never leak interval handlers.

andycreeth:
Please fix so we can merge this in.

PR #2015, opened 2019-08-19
Trim leading slash from app assets

andycreeth:


andycreeth:
Should be using a purpose built URI joining library for this to make this more robust.

PR #2018, opened 2019-08-19
Fix translations

holiber:


holiber:
Was too rushing when preparing the project for translations. Seems like the regexp expression that fetches strings wrapped in `$t()`  is not perfect

PR #2019, opened 2019-08-20
fix facebook errors for new/blank pages

andycreeth:


andycreeth:
@holiber Also for people who have never streamed or scheduled an event on their pages before.

gettinToasty:
i thought i had fixed this at some point but maybe it regressed in the refactor

andycreeth:
This regressed in the refactor.

PR #2020, opened 2019-08-20
Fix focus bug when deleting a scene collection

kuhasuki:
Using `confirm` method caused strange focus behavior in the manage scene collections window. Using an electron dialog seems to solve this issue.

### DIFF
--- a/app/components/EditableSceneCollection.vue.ts
+++ b/app/components/EditableSceneCollection.vue.ts
@@ -4,6 +4,7 @@ import { SceneCollectionsService } from 'services/scene-collections';
 import { Inject } from 'services/core/injector';
 import moment from 'moment';
 import { $t } from 'services/i18n';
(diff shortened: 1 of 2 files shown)


app/components/EditableSceneCollection.vue.ts
```diff
@@ -85,15 +86,17 @@ export default class EditableSceneCollection extends Vue {
...
-    ) {
-      return;
+    const ok = electron.remote.dialog.showMessageBox(electron.remote.getCurrentWindow(), {
+      type: 'warning',
+      message: $t('Are you sure you want to remove %{collectionName}?', {
+        collectionName: this.collection.name,
+      }),
+      buttons: [$t('Cancel'), $t('OK')],
+      noLink: true,
+    });
```
andycreeth: Let's switch to using the asynchronous version of this API.

PR #2027, opened 2019-08-21
Global search in the settings window

holiber:
search


holiber:
Yeah. All not vue-style code is wrapped into one method here, BTW

PR #2028, opened 2019-08-21
Electron 6

andycreeth:


app/components/pages/Studio.vue
```diff
@@ -40,7 +40,6 @@
 @import "../../styles/index";
 
 .studio-page {
-  height: 100%;
```
gettinToasty: wondering what this was for

andycreeth: When I booted up electron 6, the footer was gone.  It was pushed into overflow.  Removing this fixed the issue.  It's a flex column layout so the `height: 100%` was a bit strange there anyway.  I'm assuming new chrome just changed how they handle that case.

PR #2033, opened 2019-08-23
Add native events and merge live and studio

gettinToasty:


app/services/recent-events.ts
```diff
@@ -0,0 +1,273 @@
...
+  @mutation()
+  SET_RECENT_EVENTS(eventArray: IRecentEvent[]) {
+    this.state.recentEvents = eventArray;
+  }
+
+  @mutation()
+  SET_MUTED(muted: boolean) {
+    this.state.muted = muted;
+  }
+}
```
holiber: Please mark as `private` the all methods that are not calling outside this file. It provides better IntelliSense and refactoring capabilities

app/components/RecentEvents.tsx
```diff
@@ -0,0 +1,204 @@
...
+  popoutMediaShare: Function;
+  popoutRecentEvents: Function;
+  muteEvents: Function;
+  skipAlert: Function;
+  toggleQueue: Function;
+  queuePaused: boolean;
+  muted: boolean;
+}
+
+// TODO: Refactor into stateless functional component
```
holiber: Honestly, I don't think it gives us any profit to split a single component into multiple components if you are not going to reuse this sub-components. Especially if you have only a single instance of sub-component like `Toolbar`. As a result, you need extra boilerplate code to transfer shared props between parent and child.

I'm not going to say you should refactor this now. But I don't think it should be refactored into `stateless functional component`. Our stateless components are in the `components/shared` folder, and they are really generic

holiber:
Tried to play with this PR. Here are some questions/concerns that I found
image

andycreeth:
Going to merge-as is to start getting feedback from preview testers.  There are a few more things I would like to address, but will do that in a later PR.

PR #2035, opened 2019-08-23
Re-request token in case of 403 http code

holiber:


andycreeth:
On second investigation I don't think this will help.

PR #2043, opened 2019-08-28
Facebook account merging

andycreeth:
New flow for facebook merging.


app/components/pages/FacebookMerge.tsx
```diff
@@ -0,0 +1,85 @@
...
+    return (
+      <div style={{ display: 'flex', alignItems: 'center', justifyContent: 'center' }}>
+        <div style={{ width: '400px' }}>
+          <h1>Connect Facebook</h1>
+          {this.showLogin ? this.loginStep(h) : this.createPageStep(h)}
+        </div>
+      </div>
+    );
+  }
+}
```
holiber: Strings should be wrapped into `$t()`

PR #2056, opened 2019-09-03
Embed recent events

andycreeth:
Add an option for using legacy recent events, as well as a tip explaining this change.


app/components/RecentEvents.tsx
```diff
@@ -67,6 +83,68 @@ export default class RecentEvents extends TsxComponent<{}> {
...
+      </div>
+    );
+  }
+
+  renderEmbeddedEvents(h: Function) {
+    return (
+      <BrowserView
+        class={styles.eventContainer}
+        src={this.userService.recentEventsUrl()}
+        setLocale={true}
```
holiber: Don't see `setLocale` in BrowserView docs. How does it work? Does it take a locale from cookies?

app/components/shared/HelpTip.vue
```diff
@@ -22,12 +22,10 @@
   position: absolute;
   background: var(--teal);
   color: var(--white);
-  top: -8px;
-  right: 102px;
   width: 260px;
   padding: 10px;
   font-size: 14px;
-  animation: bounce 18s ease-in-out 2s 3 both;
+  z-index: 9999;
```
holiber: I wish we'll define a constants for layers soon

app/services/app/app.ts
```diff
@@ -84,6 +84,8 @@ export class AppService extends StatefulService<IAppState> {
 
   readonly pid = require('process').pid;
 
+  onboarded: boolean;
```
holiber: All public properties should be a part of the IAppState to be accessible from all windows

app/components/RecentEvents.tsx
andycreeth: This is not the electron `BrowserView` directly, since `BrowserView` is not a DOM-based API like `webview`.  This is our custom Vue component that makes it behave more like a `webview`.  The `setLocale` code was ported from our `webview` implementation.

PR #2057, opened 2019-09-03
Integrate hotkeys search with global search

holiber:


andycreeth:
Having a few issues with this.  Going to hold off merging until next week and we can work them out.

andycreeth:
Sorry been meaning to return to this.  Can't remember what my original issue was when I tested.

PR #2068, opened 2019-09-04
recent events timestamp fixes

andycreeth:


app/components/RecentEvents.tsx
```diff
@@ -178,10 +179,39 @@ class EventCell extends TsxComponent<{
...
+  get createdAt(): moment.Moment {
+    if (this.event.iso8601Created) {
+      return moment(this.event.iso8601Created);
+    }
+
+    return moment.utc(this.event.created_at);
+  }
+
   render(h: Function) {
+    console.log(this.event);
```
kuhasuki: Remove console log

app/components/RecentEvents.tsx
```diff
@@ -83,6 +83,7 @@ export default class RecentEvents extends TsxComponent<{}> {
           {!!this.recentEvents.length &&
             this.recentEvents.map(event => (
               <EventCell
+                key={event.id.toString()}
```
kuhasuki: Most events don't have an id. Most do have an _id which is somewhat but not entirely reliable. Event hash is the current state of the art for unique event identification

andycreeth:
Will wait on Peter's hash function to use as the key before merging this.

PR #2073, opened 2019-09-05
fix some weirdness with forms

andycreeth:


andycreeth:
It really should have never been a `v-model` because `sourceProperties` only had a getter defined and not a setter.  So I would have expected that to cause issues.  What confused me is why this only recently started happening.  I don't recall seeing the error before and I couldn't find it in sentry.  This code hasn't changed in a year.

PR #2074, opened 2019-09-05
Native event feed feature: read events

kuhasuki:
Adds the ability to click on events to mark them as read which will persist read state across all instances of recent events.


app/services/recent-events.ts
```diff
@@ -123,14 +185,19 @@ export class RecentEventsService extends StatefulService<IRecentEventsState> {
 
   private async formEventsArray() {
     const events = await this.fetchRecentEvents();
+    // const readReciepts = await this.fetchReadReceipts();
     let eventArray: IRecentEvent[] = [];
     if (!events.data) return;
     Object.keys(events.data).forEach(key => {
+      const fortifiedEvents: IRecentEvent[] = events.data[key].map(event => {
```
andycreeth: I think the type annotation isn't needed here or on `culledEvents` below.  It was left over from some older code but I think we can try removing and see if the type inference figures this out.

app/services/recent-events.ts
```diff
@@ -52,6 +57,59 @@ const subscriptionMap = (subPlan: string) => {
   }[subPlan];
 };
 
+function getHashForRecentEvent(event: IRecentEvent) {
```
andycreeth: Should we leave some sort of comment that this is duplicated from the website, and that this will be refactored eventually?

app/services/recent-events.ts
```diff
@@ -34,6 +33,12 @@ export interface IRecentEvent {
   displayString?: string;
   comment?: string;
   title?: string;
+  createdAt?: string;
```
andycreeth: This isn't coming from the server is it?  I see it's used in some hashes, but I didn't see it yesterday when I was working on the timestamps code (not merged yet, blocked by this PR).

app/services/recent-events.ts
```diff
@@ -123,14 +185,19 @@ export class RecentEventsService extends StatefulService<IRecentEventsState> {
 
   private async formEventsArray() {
     const events = await this.fetchRecentEvents();
+    // const readReciepts = await this.fetchReadReceipts();
```
andycreeth: Delete?

app/services/recent-events.ts
```diff
@@ -156,19 +223,43 @@ export class RecentEventsService extends StatefulService<IRecentEventsState> {
...
+          'read',
+          'hash',
         ]);
       });
 
       eventArray = eventArray.concat(culledEvents);
     });
 
+    // Format string of keys to look for in server event cache
+    const hashValues = eventArray.map(event => event.hash).join('|##|');
```
andycreeth: This is bizarre

app/services/recent-events.ts
```diff
@@ -183,6 +274,21 @@ export class RecentEventsService extends StatefulService<IRecentEventsState> {
     return await fetch(new Request(url, { headers, body, method: 'POST' })).then(handleResponse);
   }
 
+  async readAlert(event: IRecentEvent) {
```
andycreeth: This is a strangely named function but I guess it matches the API route so it's ok.

app/services/recent-events.ts
```diff
@@ -298,6 +404,15 @@ export class RecentEventsService extends StatefulService<IRecentEventsState> {
     this.state.recentEvents = events.concat(this.state.recentEvents);
   }
 
+  @mutation()
+  private TOGGLE_RECENT_EVENT_READ(eventHash: string) {
+    const matches = this.state.recentEvents.forEach((event, index) => {
+      if (event.hash === eventHash) {
+        this.state.recentEvents[index].read = !this.state.recentEvents[index].read;
+      }
+    });
```
andycreeth: `matches` is not used.

Also, no need to modify with index like this.  You can mutate the object from within the `forEach`.

```TypeScript
    this.state.recentEvents.forEach(event => {
      if (event.hash === eventHash) {
        event.read = !event.read
      }
    });
```

app/services/recent-events.ts
kuhasuki: Good idea, will add a comment

app/services/recent-events.ts
kuhasuki: It's hard for me to say definitively where each property comes from. Each event can originate from io server or streamlabs.com but client side will also fill in missing values. `created_at`, `createdAt` and `CreatedAtTimestamp` properties are present on some but not all events. I only added `createdAt` because it is used in the generation of the cache key used to fetch the read status of `treat` events.

app/services/recent-events.ts
andycreeth: To elaborate, this works because objects are always passed by reference in javascript.

app/services/recent-events.ts
andycreeth: In my PR I added another field called something like `iso8601Timestamp`.  I guess the pattern continues.

app/services/recent-events.ts
kuhasuki: Good point. Was using find earlier.

app/services/recent-events.ts
kuhasuki: Yep, this API is particularly unwieldy for no discernible reason

app/services/recent-events.ts
kuhasuki: I also used it in the cache key I constructed for merch events, since merch events actually don't have a hash or read property defined.

app/services/recent-events.ts
kuhasuki: Yup wasn't needed

PR #2076, opened 2019-09-06
Various fixes for recent events

kuhasuki:
Fix bug where wrong value was being sent as read
Fix subscription related text
Remove facebook prefixed events
Add uuid property to events and use where client side unique identifier is needed


app/services/recent-events.ts
```diff
@@ -42,6 +43,7 @@ export interface IRecentEvent {
   hash: string;
   isTest?: boolean;
   repeat?: boolean;
+  uuid: string;
```
andycreeth: Add a comment that this is a local id and is NOT persistent across app restarts/fetches.

PR #2077, opened 2019-09-06
fix duplicate keys in events

andycreeth:


andycreeth:
Closing due to alternate solution

PR #2083, opened 2019-09-09
Hide media share button when media share is not enabled

kuhasuki:


andycreeth:
Forgot to merge this in and looks like there is a conflict now

PR #2086, opened 2019-09-11
Prevent double subscription to websocket service

kuhasuki:


app/services/recent-events.ts
```diff
@@ -164,10 +166,20 @@ export class RecentEventsService extends StatefulService<IRecentEventsState> {
...
     return this.fetchMutedState();
   }
 
+  makeSocketConnection() {
+    if (this.socketConnection) {
+      this.socketConnection.unsubscribe();
+    }
+
+    this.socketConnection = this.websocketService.socketEvent.subscribe(
+      this.onSocketEvent.bind(this),
```
holiber: Don't use `.bind` since it loses typechecking. Use arrow functions instead. 
```
this.socketConnection = this.websocketService.socketEvent.subscribe(ev => this.onSocketEvent(ev))
```

PR #2091, opened 2019-09-13
Import widgets from OBS

holiber:


app/services/obs-importer.ts
```diff
@@ -182,8 +186,16 @@ export class ObsImporterService extends StatefulService<{ progress: number; tota
 
         if (isSourceAvailable) {
           if (sourceJSON.id !== 'scene') {
+            let propertiesManager: TPropertiesManager = 'default';
+            let propertiesManagerSettings: Dictionary<any> = {};
+
             if (sourceJSON.id === 'browser_source') {
               sourceJSON.settings.shutdown = true;
+              const widgetType = this.widgetsService.getWidgetTypeByUrl(sourceJSON.settings.url);
```
andycreeth: For the purposes of this migration, we should only convert it to a widget if there are no query parameters in the URL.  It's a bit of an edge case, but out widget URLs accept query parameters for doing stuff like using the non-default alert profile.  To avoid breaking people's setups when we import, it's probably best to just ignore those for now.

andycreeth:
Can we get my suggestion addressed so we can merge this in?

app/services/widgets/widgets-api.ts
```diff
@@ -3,6 +3,7 @@ import { AnchorPoint } from 'util/ScalableRectangle';
 import { WidgetType } from './widgets-data';
 import { ISourceApi } from 'services/sources';
 import { IAlertBoxVariation } from './settings/alert-box/alert-box-api';
+import { key } from 'aws-sdk/clients/signer';
```
andycreeth: Looks like an auto-import gone wrong

PR #2092, opened 2019-09-13
Event feed filtering

kuhasuki:
Adds a child window which allows users to filter most event types.
Doesn't support filtering redemptions.


app/components/windows/EventFilterMenu.tsx
```diff
@@ -0,0 +1,163 @@
...
+            </div>
+          </div>
+        </div>
+        <div slot="controls">
+          <button class="button button--action" onClick={this.cancel}>
+            {$t('Done')}
+          </button>
+        </div>
+      </ModalLayout>
+    );
```
andycreeth: Low priority but should probably split this up into some nicely named helper functions.

app/services/recent-events.ts
```diff
@@ -334,7 +391,102 @@ export class RecentEventsService extends StatefulService<IRecentEventsState> {
...
+        name: filterName(filter),
+      };
+    });
+
+    return {
+      main,
+      sub,
+      resub,
+    };
+  }
```
andycreeth: This function throws away way too much typing information and should be refactored.  Recognize that this API is very dynamic but I think we can do better.  Let's revisit this together after we get this in preview.

anthonyalayo:
Hey There! Sorry for the wrong place to put this, but there is no issues page for this repository.

I'm testing event feed filtering with 2 alert boxes, 1 alert box with Facebook likes turned off and 1 alert box with likes turned on. This is on SLOBS with a multi-stream setup of both Twitch and Facebook (more on that later).

Screenshot 2020-10-27 193331

What I'm seeing is both alert boxes display the Facebook like event even though only one of them has it toggled. I'm testing by hitting the below endpoint. I had to do it this way and reverse engineer it because the "Test Widgets" button doesn't provide all event types.

```
GET https://streamlabs.com/api/v5/slobs/test/facebook_account/like HTTP/1.1
Host:streamlabs.com
Connection:keep-alive
Sec-Fetch-Mode:cors
authorization:Bearer *******************************
User-Agent:Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) StreamlabsOBS/0.25.0 Chrome/76.0.3809.146 Electron/6.1.11 Safari/537.36
Accept:*/*
Sec-Fetch-Site:cross-site
Accept-Encoding:gzip, deflate, br
Accept-Language:en-US
Cookie:__**********************
```

On top of that, when I attempt to use the advanced option on the streamlabs.com dashboard to build a URL that has filtering built into it and put that into a browser source, I see nothing on that browser source. For example with the url:

```
https://streamlabs.com/alert-box/v3/***********?like=1
```

It works when I remove the filter query parameters, but with that I can't filter events which is the whole point of the advanced url version.

Screenshot 2020-10-27 193421

Somewhat related, I noticed that when opening up the alert box settings within SLOBS after configuring multi-streaming, the only events that are editable (and hence save to the server) are the settings corresponding to the streaming platform selecting on the streamlabs.com website toggle below the profile photo on the top right.

Screenshot 2020-10-27 193346

For example If I have "Twitch" specified in the dropdown on the streamlabs profile photo, then only the Twitch events can get modified in SLOBS. Sometimes SLOBS will show all of the events, and sometimes it will only show the Twitch events listed. I'm not sure why exactly it is set up like this at all, but it is almost impossible to figure out for the average person.

The reason I had to go down this rabbit hole is this:

> Doesn't support filtering redemptions.

I wanted to have loyalty point redemptions in a different, much larger alert box. Without that ability, the feature was more or less a throwaway for me. Digging into things I noticed that not only was it not there, but the ability to filter events just isn't working at all, even with the documented way on the dashboard.

Could we have someone look into these issues? It seems to land into use cases that are quite important.

I've attached some photos in case it helps.

Thanks,
Anthony

PR #2094, opened 2019-09-16
enable crash handler logging

summeroff:


andycreeth:
We don't have linting on the main process since it's not TypeScript so I pushed a commit to this branch bringing it in line with our code style.

PR #2095, opened 2019-09-16
Fix bug with socket event filtering

kuhasuki:
Type property was missing, adding necessary event type info for proper filtering.


andycreeth:
I think there's a cleaner way to do this.  Let's chat.

PR #2110, opened 2019-09-19
Use Class as a definition for component's props (proof of concept)

holiber:


andycreeth:
I like this a lot actually.

gettinToasty:
my only wonder here is if we can access `TProps` like a class in `TsxComponent` and remove the need for a helper method, that might not be possible though

andycreeth:
I kind of like how you have to be explicit with `this.props`.  It makes it clear you're accessing a prop and not a computed property.  Maybe that's just me though.

gettinToasty:
that's how props are typically accessed in react, i'm also a fan of that pattern

holiber:
Accessing props via `this.props` seems a bit clearer than just mixing it into the `this` instance. But it seems like not a big deal at all

holiber:
I actually can't figure out an easy way to incorporate props into this without additional boilerplate code

PR #2112, opened 2019-09-19
Selective recording

eddyStreamlabs:


app/components/shared/Display.vue.ts
```diff
@@ -21,6 +21,8 @@ export default class Display extends TsxComponent<DisplayProps> {
   @Prop() sourceId: string;
   @Prop({ default: 0 }) paddingSize: number;
   @Prop({ default: false }) drawUI: boolean;
+  @Prop() clickHandler: boolean;
```
andycreeth: Is this meant to be here?

kuhasuki: Nope got added on accident

PR #2118, opened 2019-09-23
Remove disabled class from selective recording icon 

kuhasuki:
Users should be able to cycle through the options on each source even after they've gone live or started recording.


kuhasuki:
Added the disabled class when the source is locked

PR #2133, opened 2019-09-25
Fix the expanded-folder icon

holiber:
Fix the regression from #2112


gettinToasty:
should this be based against preview?

PR #2153, opened 2019-10-03
Clear stream key on logout

gettinToasty:


app/services/user.ts
```diff
@@ -306,6 +307,18 @@ export class UserService extends PersistentStatefulService<IUserServiceState> {
...
+    const settings = this.settingsService.getSettingsFormData('Stream');
+
+    settings.forEach(subCategory => {
+      subCategory.parameters.forEach(parameter => {
+        if (parameter.name === 'key') parameter.value = '';
+      });
+    });
+
+    this.settingsService.setSettings('Stream', settings);
+  }
```
andycreeth: There's a function in the settings service called `setSettingValue` that makes this much easier now.

PR #2157, opened 2019-10-04
translations: fail tests if translations are not found

holiber:


gettinToasty:
i think this branch can be closed now right?

holiber:
> i think this branch can be closed now right?

I should remove HTML and CSS changes from this branch and continue working on translatioins

PR #2161, opened 2019-10-04
Add icons to footer buttons

gettinToasty:


andycreeth:
Replay buffer icon is cut off on the themes tab.

andycreeth:
I wasn't logged into a platform that supports scheduling, but should check if scheduled has the same issue since it also looks like a top tooltip.

PR #2162, opened 2019-10-04
Default props for TsxComponent

holiber:


gettinToasty:
tbh i didn't know you could default types like this but seems like a nice QOL change :+1:

PR #2167, opened 2019-10-07
fix undo for removing scene items

andycreeth:


app/services/editor-commands/commands/remove-item.ts
```diff
@@ -3,9 +3,9 @@ import { ScenesService } from 'services/scenes';
 import { Inject } from 'services/core/injector';
 import { SourcesNode } from 'services/scene-collections/nodes/sources';
 import { Source } from 'services/sources';
-import { ResetTransformCommand } from './reset-transform';
 import { ReorderNodesCommand, EPlaceType } from './reorder-nodes';
 import { $t } from 'services/i18n';
+import { ISceneItemSettings } from 'services/api/external-api/scenes';
```
holiber: We should not use interfaces for extenral-api here

PR #2171, opened 2019-10-08
Advanced stats redesign

kuhasuki:
Update:
-Merged performance monitor and performance services
-Adds advanced statistics window
-Color coordinate icon to stream status
-Status is one of three states, good, ok and bad. currently the hardcoded limits for dropped frames, skipped frames and lagged frames are < 30% = good, >30% = ok, > 50% = bad, stream reconnecting = bad
-Allow users to selectively add performance metrics into their footer


app/components/windows/AdvancedStatistics.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+  get status(): { type: string; description: string } {
+    if (this.streamingStatus === 'reconnecting' || this.percentDropped > 50) {
+      return {
+        type: 'error',
+        description: $t('Streamlabs OBS is experiencing difficulties broadcasting'),
+      };
+    }
+
+    if (this.percentDropped > 30) {
```
holiber: AFAIK we have  `droppedThreshold=0.1` value in the TroubleshooterService. Are we going to use another value here?

kuhasuki: Yeah the idea is that we don't want everyone to think their stream is going bad, and 10% dropped frames isn't necessarily going to significantly reduce the quality of the stream. Discussed with Andy and we haven't decided on final numbers yet.

app/components/StudioFooter.vue
```diff
@@ -125,6 +125,18 @@
   }
 }
 
+.warning {
+  color: var(--warning);
+}
+
+.info {
+  color: #e3973e;
```
gettinToasty: is there a reason to use a hex value here instead of one of our existing variable colors?

app/components/windows/AdvancedStatistics.m.less
```diff
@@ -0,0 +1,135 @@
...
+  margin-bottom: 20px;
+  .radius(1);
+
+  span {
+    margin-left: 5px;
+  }
+}
+
+.error {
+  background: rgba(251, 72, 76, 0.28);
```
gettinToasty: we should define this color in colors.less and assign it a variable if it doesn't already exist

app/components/windows/AdvancedStatistics.m.less
```diff
@@ -0,0 +1,135 @@
...
+  color: var(--warning);
+}
+
+.warning {
+  background-color: rgba(227, 151, 62, 0.08);
```
gettinToasty: same as above, althought i'm pretty sure we already have a transparent yellow

app/components/windows/AdvancedStatistics.m.less
```diff
@@ -0,0 +1,135 @@
...
+  color: #e3973e;
```
gettinToasty: same as first comment, i feel like `var(--info)` should work here

app/components/windows/AdvancedStatistics.m.less
```diff
@@ -0,0 +1,135 @@
...
+  color: var(--warning);
+}
+
+.warning-text {
+  color: #e3973e;
+}
+
+.success {
+  background-color: rgba(49, 195, 162, 0.08);
+  color: #31c3a2;
```
gettinToasty: not sure why `success` uses `var(--teal)` above but a hex color here

app/components/windows/AdvancedStatistics.tsx
```diff
@@ -0,0 +1,194 @@
...
+    this.streamingStatusChange.unsubscribe();
+    clearInterval(this.updateInterval);
+  }
+
+  get streamQuality() {
+    return this.performanceService.streamQuality;
+  }
+
+  get status(): { type: string; description: string } {
+    if (this.streamingStatus === 'reconnecting' || this.streamQuality === EStreamQuality.POOR) {
```
gettinToasty: should be `this.streamingStatus === EStreamingState.Reconnecting` right?

app/components/windows/AdvancedStatistics.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  moment(time: number): string {
+    return moment(time).fromNow();
+  }
+
+  onNotificationHandler(notification: INotification) {
+    if (notification.subType === ENotificationSubType.DEFAULT) {
+      return;
+    }
+    this.notifications.push(notification);
```
gettinToasty: do we want to unshift here so new notifs are at the top?

app/components/windows/AdvancedStatistics.tsx
```diff
@@ -0,0 +1,194 @@
...
+          )}
+        >
+          {this.statusIcon}
+          <span>{this.status.description}</span>
+        </div>
+      </div>
+    );
+  }
+
+  notificationsArea() {
```
gettinToasty: should be a getter if there are no parameters accepted

app/services/performance.ts
```diff
@@ -1,43 +1,99 @@
...
+  @Inject() private troubleshooterService: TroubleshooterService;
+
   static initialState: IPerformanceState = {
     CPU: 0,
     numberDroppedFrames: 0,
     percentageDroppedFrames: 0,
+    numberSkippedFrames: 0,
+    percentageSkippedFrames: 0,
+    numberLaggedFrames: 0,
+    percentageLaggedFrames: 0,
```
gettinToasty: do we need percentages stored in state? i feel like those could all be calculated values

app/components/windows/AdvancedStatistics.tsx
kuhasuki: If I make it a getter, the setInterval forceUpdate that makes the `2 minutes ago` labels update doesn't work

app/components/windows/AdvancedStatistics.tsx
kuhasuki: Yep, will fix

app/components/windows/AdvancedStatistics.tsx
kuhasuki: We probably do. Will check and update.

app/components/windows/AdvancedStatistics.tsx
kuhasuki: Making it a getter breaks the moment driven time label updates

app/services/performance.ts
kuhasuki: Can refactor this I think

kuhasuki: So percentage dropped frames comes directly from OBS pre-calculated.
I can make the other two percentages getters with calculated values but I think it's better to keep things consistent, unless there's a good reason to avoid storing things in state as much as possible.

app/services/performance.ts
```diff
@@ -1,43 +1,99 @@
...
     bandwidth: 0,
     frameRate: 0,
   };
 
-  droppedFramesDetected = new Subject<number>();
-  private intervalId: number;
+  private historicalDroppedFrames: number[] = [];
+  private historicalSkippedFrames: number[] = [];
+  private historicalLaggedFrames: number[] = [];
+  private numberOfSamples: number = SAMPLING_DURATION / STATS_UPDATE_INTERVAL;
```
andycreeth: Seems like it would make more sense for this to be defined up with the other constants.

app/services/customization/customization.ts
```diff
@@ -62,6 +63,12 @@ export class CustomizationService extends PersistentStatefulService<ICustomizati
     folderSelection: false,
     navigateToLiveOnStreamStart: true,
     legacyEvents: false,
+    pinnedStatistics: {
+      cpu: true,
```
andycreeth: I think we should default everything to off

kuhasuki: Oh whoops I put that in for testing and didn't change it back. Good catch

app/services/performance.ts
kuhasuki: Done

app/components/StudioFooter.vue
kuhasuki: fixed

app/components/windows/AdvancedStatistics.m.less
kuhasuki: fixed

app/components/windows/AdvancedStatistics.m.less
kuhasuki: fixed

PR #2172, opened 2019-10-08
Make searchable pages reactive again

gettinToasty:


app/components/shared/SearchablePages.tsx
```diff
@@ -25,9 +25,9 @@ class SearchablePagesProps {
  */
 @Component({ props: createProps(SearchablePagesProps) })
 export default class SearchablePages extends TsxComponent<SearchablePagesProps> {
-  currentPage: string;
+  currentPage: string = '';
   pagesInfo: Dictionary<IPageInfo> = null;
-  searchResultPages: string[];
+  searchResultPages: string[] = null;
```
andycreeth: Should this be empty array instead?  Generally empty array makes more sense and avoids accidental null errors.

PR #2174, opened 2019-10-09
Add basic test for selective recording

kuhasuki:
Would love some feedback on this test.

creates a temporary recording path
adds a single browser source
enables selective recording
cycles selective recording to record only
records 2s of video
checks that file exists 

I'm concerned that the entire test hinges on css class names for specific icons and treats presence of a given icon as an indicator of proper functionality. It might be better to inspect the customization service to ensure the selective recording mode matches expectations.

### DIFF
--- a/test/selective-recording.ts
+++ b/test/selective-recording.ts
@@ -1,6 +1,5 @@
 import { readdir } from 'fs-extra';
-import { focusChild, focusMain, test, useSpectron } from './helpers/spectron';
-import { setFormDropdown } from './helpers/spectron/forms';
+import { focusMain, test, useSpectron } from './helpers/spectron';
 import { sleep } from './helpers/sleep';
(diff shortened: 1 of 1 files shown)


andycreeth:
Just to add to the above, wanting to record 2 seconds of video is a perfect example of why you might want to use `sleep` in your test.  That one can stay.  All the others should go though.

test/selective-recording.ts
```diff
@@ -0,0 +1,59 @@
...
+  // Add a browser source
+  await addSource(t, sourceType, sourceName);
+
+  // Toggle selective recording
+  await focusMain(t);
+  await client.click('.studio-controls-top .icon-smart-record');
+
+  // Check that selective recording icon is active
+  await sleep(2000);
+  t.true(await client.isExisting('.icon-smart-record.icon--active'));
```
holiber: You can use `client.waitForVisible` to get rid of `sleep`

PR #2175, opened 2019-10-09
Fix bug in recent event filter menu after h autoinject

kuhasuki:
Looks like h is autoinjected but render functions need get. Wondering if this is happening anywhere else.


gettinToasty:
yea that was my bad, i missed another spot as well

PR #2177, opened 2019-10-09
Reorganize colors for day mode

gettinToasty:


holiber:
BTW this list inputs look like disabled elements for me
image

PR #2182, opened 2019-10-10
Correcting french translations

dreais:


dreais:
84ad3e0
I'm not sure about this one - was the file intended to stay in English? If yes, I may fallback!

holiber:
Thanks for your corrections

dreais:
will do another one when i figure out the correct meanings of some files! but thank you for your feedback, much appreciated :)

PR #2199, opened 2019-10-15
Prevent moving locked selections with arrow keys

kuhasuki:
This is the naive fix for this issue. With this change, if you have multiple sources selected and any one of them isn't locked, all sources in the selection will be nudged by pressing arrow keys.

I don't think many users will try to nudge a selection containing both locked and unlocked sources.


gettinToasty:
does `isLocked` check if all sources are locked or any? i thought it was the latter

kuhasuki:
> does `isLocked` check if all sources are locked or any? i thought it was the latter

It's the former, isLocked is only true when all sources in the selection are locked

PR #2218, opened 2019-10-21
Youtube rework

holiber:
Youtube now requires to select/create an event to broadcast.

This PR includes some changes in streaming services.

# Highlights 
- 'any' is forbidden for platform's API requests
- Added more strict typechecking
- Each platform service setup stream key and update channel info in `beforeGoLive` hook
- Each platform service emit `canneInfoChanged` with updated info like `streamUrl`, `chatUrl`, `title`, `description`
- Vue components should use `StreamInfoService` for obtaining the streamInfo or listening to the `streamInfoChanged` event. `StreamInfoService` takes care of listening `canneInfoChanged` from platform services. It subscribes/unsubscribes to this event on logIn/logOut.
- `UserService` emits the 'userLogin' event even when the app starts with saved auth credentials

# Test plan
- check streaming for all platforms
- stream to new youtube event
- schedule stream to youtube and stream to the scheduled event
- check that users with manually changed streamKey are automatically migrated to the "Custom injest mode"
- check streaming works with disabled GoLive window for TW/MX
- check that disabled GoLive window setting does not apply for FB/YT
- check schedule stream for FB/YT
- check that in case of networks issues you have an error message in GoLive window and able to retry
- check that chatUrl and streamUrl are always up to date with each new YT/FB stream
- check that chatUrl and streamUrl are always up to date in Game Overlay
- check that "Use recommended settings" button works in Settings -> stream
- check streaming in the "Custom ingest mode"
- check "recording while streaming" feature

I'm going to provide  automated tests for this PR here https://github.com/stream-labs/streamlabs-obs/pull/2227
However, it's worth to go through this list before the release


app/services/user.ts
```diff
@@ -141,6 +145,9 @@ export class UserService extends PersistentStatefulService<IUserServiceState> {
 
     fetch(request)
       .then(res => {
+        const service = getPlatformService(this.state.auth.platform.type);
+        this.login(service, this.state.auth);
+        this.refreshUserInfo();
         return res.text();
       })
       .then(valid => {
```
holiber: Don't think it works correctly. this.LOGOUT() will be never called because we didn't use 'handleResponse' method here. It ofter returns '401' for me

app/services/scene-collections/scene-collections.ts
```diff
@@ -232,7 +231,7 @@ export class SceneCollectionsService extends Service implements ISceneCollection
     const removingActiveCollection = id === this.activeCollection.id;
 
     if (removingActiveCollection) {
-      this.appService.runInLoadingMode(async () => {
+      await this.appService.runInLoadingMode(async () => {
```
holiber: Got race-condition issues when tried to remove several scene-collections in a row

app/services/user.ts
```diff
@@ -144,7 +148,13 @@ export class UserService extends PersistentStatefulService<IUserServiceState> {
...
-        if (valid.match(/false/)) this.LOGOUT();
+        if (valid.match(/false/)) {
+          this.LOGOUT();
```
andycreeth: So this API always returns true now but we may change that.  Can be a separate PR, but I think we should display a message that says something like "You have been logged out.  Please log back in to continue using Streamlabs OBS" and then take them to the login screen.

app/components/platforms/youtube/BroadcastInput.tsx
```diff
@@ -0,0 +1,104 @@
...
+  render() {
+    // define custom slot for the list-item
+    const scopedSlots = {
+      item: (props: { option: IListOption<string> }) => {
+        const broadcast = this.options.broadcasts.find(
+          broadcast => broadcast.id === props.option.value,
+        );
+
+        // "Create New" option
+        if (!broadcast) {
```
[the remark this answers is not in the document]
holiber: We allow creating a new broadcast in all scenarios. Just 'Create new' is an list-item option with `value=''`.

app/services/chat.ts
```diff
@@ -47,6 +76,7 @@ export class ChatService extends Service {
   }
 
   unmountChat(electronWindowId: number) {
+    this.electronWindowId = 0;
```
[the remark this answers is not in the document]
holiber: It is always kind of weird to mix Number and Object types (like null) in one type. But here it can be a good case

app/services/user.ts
holiber: Added a message

app/services/chat.ts
```diff
@@ -8,26 +8,55 @@ import { YoutubeService } from 'services/platforms/youtube';
...
+      if (!this.chatUrl) return;
+
+      // chat url changed to a new valid url, init or reload chat
+      if (oldChatUrl) {
+        this.deinitChat();
+        this.initChat();
+      } else {
+        this.initChat();
+      }
+    });
```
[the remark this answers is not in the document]
holiber: Just check it didn't deinit on logout. Thanks for catching this. I send an event with updated `channelInfo` on logout now. It contains empty url

PR #2219, opened 2019-10-21
Remove new themes badge and media share icon

gettinToasty:


gettinToasty:
not sure how to fix the media share test without relying on the success state icon

PR #2224, opened 2019-10-23
Fix janky login buttons

gettinToasty:
The color would default to Foundation blue when loading on hover

### DIFF
--- a/app/styles/buttons.less
+++ b/app/styles/buttons.less
@@ -179,7 +179,7 @@
   background: #6441a4;
 
   &:hover {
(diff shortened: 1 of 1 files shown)


holiber:
using '!important' in the codebase is not a good practice, can we avoid this?

gettinToasty:
@holiber there are a few use-cases where using !important is acceptable and i felt this followed enough constraints for it to be alright, instead of following down a likely convoluted specificity chain:
- overwriting 3rd party styles (from foundation.css)
- styles that are only used in one place
- already high amounts of specificity for the important tags to even apply

PR #2227, opened 2019-10-24
Platform tests

holiber:
Tests for integration with platforms


holiber:
close in favor of https://github.com/stream-labs/streamlabs-obs/pull/2254

PR #2231, opened 2019-10-26
Properly filter FB events in minifeed

kuhasuki:
Facebook events were being erroneously filtered due to inconsistency between the keys used to identify FB alerts (event type) and the setting to filter that event.


kuhasuki:
> i think we probably want to handle this transform without actually mutating state, can use lodash's `cloneDeep` to avoid this

Ah yeah can do.

PR #2232, opened 2019-10-29
Hide display by dragging

kuhasuki:
Allow users who want to hide their display to hide it by dragging controls or recent events to the top of the window.

### DIFF
--- a/app/components/pages/Studio.vue.ts
+++ b/app/components/pages/Studio.vue.ts
@@ -74,18 +74,20 @@ export default class Studio extends Vue {
    */
   @Watch('performanceMode')
   reconcileHeightsWithinContraints(isControlsResize = false) {
-    const containerHeight = this.$root.$el.getBoundingClientRect().height;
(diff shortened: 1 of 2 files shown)


app/components/pages/Studio.vue.ts
```diff
@@ -76,14 +77,22 @@ export default class Studio extends Vue {
     const containerHeight = this.$root.$el.getBoundingClientRect().height;
 
     // This is the maximum height we can use
-    this.maxHeight = containerHeight - (this.performanceMode ? 200 : 400);
+    this.maxHeight = containerHeight - (this.performanceMode ? 200 : 1);
```
gettinToasty: not sure why we're leaving a 200px maxHeight for performance mode here?

kuhasuki: If you do enable performance mode, you probably want to see the button to disable performance mode?

gettinToasty: but could also just drag down the mini feed to do that too right?

PR #2233, opened 2019-10-29
Mark all historical events older than a month as read

kuhasuki:


app/services/recent-events.ts
```diff
@@ -379,6 +379,13 @@ export class RecentEventsService extends StatefulService<IRecentEventsState> {
       event.read = readReceipts[event.hash] ? readReceipts[event.hash] : false;
     });
 
+    // Events older than 1 month are treated as read
+    eventArray.forEach(event => {
+      if (new Date(event.created_at).getTime() < new Date().getTime() - 1000 * 60 * 60 * 24 * 30) {
+        event.read = true;
+      }
+    });
+
```
gettinToasty: i feel like this check can be added to the loop directly above where it checks for readReceipts

PR #2245, opened 2019-10-31
fix scene collection sync retry

andycreeth:


andycreeth:
Retry hasn't worked for over a year.

PR #2259, opened 2019-11-06
Refactor startup

holiber:
Since we introduced a change where we always emit the 'userLogin' event at the startup, we need to review and apply fixes to all services that listen to this event.

This PR introduces a new rule for the initialization of services.
All services that require to be initialized in a specific order one by one should be initialized in the `AppService.load` method. Other services that can be initialized in parallel mode should use `@InitAfter()` decorator.

# Test Plan
The order of initialization for some services has been changed and require additional testing
- test scene collection properly syncronized
- test stream labels works
- test recent events works
- test game overlay works and changes the chat URL for YT streams
- test apps work


gettinToasty:
this looks ok but i'm not 100% on what the implications are here so i'd like to have andy look at it as well

app/services/platform-apps/index.ts
```diff
@@ -178,11 +178,6 @@ export class PlatformAppsService extends StatefulService<IPlatformAppServiceStat
 
     if (!this.userService.isLoggedIn()) return;
 
-    this.SET_DEV_MODE(await this.getIsDevMode());
-
-    this.loadProductionApps();
-    this.SET_APP_STORE_VISIBILITY(await this.fetchAppStoreVisibility());
-
```
gettinToasty: not really sure why this got removed?

app/services/streamlabels/index.ts
```diff
@@ -122,8 +124,6 @@ export class StreamlabelsService extends Service {
 
   init() {
     this.ensureDirectory();
-    this.fetchInitialData();
-    this.fetchSettings();
```
gettinToasty: also not sure why these are getting removed

holiber: Both methods just exit if the user is not in logged-in  state. The user will never be in a logged-in state here. So these calls are useless now

app/services/platform-apps/index.ts
holiber: Having this code here gives a chance to run `loadProductionApps` twice. First time when the service has been itinitalized and the second time when we get the `loggedIn` event. BTW just reviewed and realized that I need to add more changes here

holiber: fixed

PR #2262, opened 2019-11-06
Update game overlay event feed

kuhasuki:
Switch game overlay to use native event feed allowing for synchronization between overlay read state and main application read state. Also makes better use of the available space to display more events.


app/components/windows/GameOverlayEventFeed.m.less
```diff
@@ -0,0 +1,7 @@
+@import "../../styles/index.less";
```
gettinToasty: this compiles to nothing so not a big deal but just FYI you dont need to import anything to use css variables

app/services/game-overlay/index.ts
```diff
@@ -117,7 +117,15 @@ export class GameOverlayService extends PersistentStatefulService<GameOverlaySta
...
     const chatWebPrefences = { ...this.commonWindowOptions.webPreferences, partition };
-    this.windows.recentEvents = new BrowserWindow({ ...this.commonWindowOptions, width: 600 });
+    this.windows.recentEvents = this.windowsService.createOneOffWindowForOverlay({
+      ...this.commonWindowOptions,
+      width: 600,
+      componentName: 'GameOverlayEventFeed',
+      queryParams: { gameOverlay: true },
+      webPreferences: { offscreen: true, nodeIntegration: true },
+      isFullScreen: true,
+      alwaysOnTop: true,
```
gettinToasty: do alwaysOnTop and isFullScreen need to be here? would have assumed the overlay plugin took care of that

app/services/recent-events.ts
```diff
@@ -793,7 +793,7 @@ export class RecentEventsService extends StatefulService<IRecentEventsState> {
 
   showFilterMenu() {
     this.windowsService.showWindow({
-      componentName: 'EventFilterMenu',
+      componentName: 'MiniFeed',
```
gettinToasty: this looks like a mistake

kuhasuki: Ah yep good catch

app/services/game-overlay/index.ts
kuhasuki: Probably not, will remove

app/components/windows/GameOverlayEventFeed.m.less
kuhasuki: I didn't know that. I'll stop doing this everywhere

app/services/game-overlay/index.ts
kuhasuki: isFullScreen makes the window title bar go away so it is necessary after all, alwaysOnTop seems to happen anyway so that can be removed

PR #2266, opened 2019-11-07
fix widget themes notification

kuhasuki:
Didn't know where to put the function so I added it to notification service.

### DIFF
--- a/app/components/pages/BrowseOverlays.vue.ts
+++ b/app/components/pages/BrowseOverlays.vue.ts
@@ -90,10 +90,8 @@ export default class BrowseOverlays extends Vue {
       showTime: false,
       message: $t('Widget Theme installed & activated. Click here to manage your Widget Profiles.'),
       action: this.jsonrpcService.createRequest(
-        Service.getResourceId(this.navigationService),
-        'navigate',
-        'Dashboard',
-        { subPage: 'widgetthemes' },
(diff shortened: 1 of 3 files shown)


gettinToasty:
hmm this seems like kinda a weird/funky place to me, why not just get the magic link service directly in the notification?

kuhasuki:
> hmm this seems like kinda a weird/funky place to me, why not just get the magic link service directly in the notification?

Magic link service only return the url. I'm not sure what I can/ can't do in the action property of a notification. The typedef says it must be  IJsonRpcRequest. Agree that notification service might not be the best place, happy to consider other services that logically make more sense. None come to mind for me right away.

gettinToasty:
what i mean is adding `openWidgetThemesMagicLink` to `magicLinksService` and doing this
```
this.jsonrpcService.createRequest(
  Service.getResourceId(this.magicLinksService),
  'openWidgetThemesMagicLink',
)
```

kuhasuki:
> what i mean is adding `openWidgetThemesMagicLink` to `magicLinksService` and doing this
> 
> ```
> this.jsonrpcService.createRequest(
>   Service.getResourceId(this.magicLinksService),
>   'openWidgetThemesMagicLink',
> )
> ```

Yeah I think that would be fine. Will move the function to magic link service

PR #2273, opened 2019-11-12
enable logging in game overlays

summeroff:


app/services/game-overlay/index.ts
```diff
@@ -105,7 +105,13 @@ export class GameOverlayService extends PersistentStatefulService<GameOverlaySta
   }
 
   async initializeOverlay() {
-    overlay.start();
+    let crashHandlerLogPath = '';
+    if (process.env.NODE_ENV !== 'production' || !!process.env.SLOBS_PREVIEW) {
```
andycreeth: So to be clear, this logging will only happen in dev and preview.  Is that what we want for now?

summeroff: Yes. It will give us information from preview users and not affect regular users too much.

andycreeth:
For whatever reason, tests never started on this PR.  Trying now.

PR #2303, opened 2019-11-18
Hide themes onboarding for those who have existing scene collections

gettinToasty:


test/scene-collections.ts
```diff
@@ -35,6 +35,7 @@ useSpectron({
       path.join(sceneCollectionsPath, 'manifest.json'),
     );
   },
+  skipThemes: true,
```
holiber: why `skipOnboarding: true`  doesn't work here?

gettinToasty: if you have an existing scene collection (like we do for this test) it doesnt show the themes page of onboarding but still shows the rest

holiber: Ok. Looks like we can handle this without additional flags
```
if (await t.context.app.client.isVisible('p=Skip')) { // the themes selector visible only if we already have scene collections 
     await t.context.app.client.click('p=Skip');
 }
```

PR #2310, opened 2019-11-19
workaround for legacy apps that modify the stream key

andycreeth:


app/components/StartStreamingButton.vue.ts
```diff
@@ -97,6 +89,38 @@ export default class StartStreamingButton extends Vue {
...
+        this.streamSettingsService.isSafeToModifyStreamKey()
+      );
+    }
+
+    if (this.isYoutube) {
+      return (
+        this.streamSettingsService.protectedModeEnabled &&
+        this.streamSettingsService.isSafeToModifyStreamKey()
+      );
+    }
```
gettinToasty: this seems a little like unnecessary repetition here-- are you just futureproofing?

andycreeth: This was mainly for my own sanity when writing it.  This was originally a very large boolean expression.  For me, the easiest way to verify this was correct was to break it out by what I expected to happen for each platform.  It's now far from a normalized boolean expression, but it was far easier for me to verify the logic was correct, and hopefully far easier to read too.

gettinToasty: makes sense :+1: for the sake of clarity i think its fine

PR #2311, opened 2019-11-19
stream-settings: forbid changing stream settings when user is live

holiber:
image


holiber:
Tests passed locally, merging

PR #2332, opened 2019-12-02
Modular Editor pt 1

gettinToasty:


gettinToasty:
gonna invoke campground rules on this codeclimate-- leaving it better than i found it 😂

app/services/layout.ts
```diff
@@ -0,0 +1,76 @@
...
+    this.state.resizes = RESIZE_DEFAULTS[layout];
+  }
+
+  @mutation()
+  SLOT_ELEMENT(element: ELayoutElement, slot: '1' | '2' | '3' | '4' | '5' | '6') {
+    this.state.slottedWidgets[element] = slot;
+  }
+
+  @mutation()
+  SET_RESIZE(bar: 'bar1' | 'bar2', size: number) {
```
holiber: We should be carefully thinking about what we should save in PersistingStatefulsService to avoid ugly migrations in the future. Let try to answer the next questions before saving these values to the persistence state for all layouts.
- Are all layouts are going to have `bar1` and `bar2`. Maybe it's better to not cap the number of bars with this 2
- Is `size` always `vertical` or `horizontal` size? Maybe it's better to use `width` and `height`?
- Should we store settings for each layout separately? For example `SET_RESIZE(layoutName, layoutOptions)`

app/components/pages/Studio.tsx
```diff
@@ -0,0 +1,10 @@
+import Vue from 'vue';
+import { Component } from 'vue-property-decorator';
+import Editor from 'components/pages/editor/Editor';
+
+@Component({})
+export default class Studio extends Vue {
+  render() {
+    return <Editor />;
+  }
+}
```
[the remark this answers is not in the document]
gettinToasty: this is just to maintain backwards compatability without changing around the navigation service. it's temporary while modularity remains invisible

app/services/layout.ts
gettinToasty: we decided at least for an initial scope that layouts will have either 1 or 2 resize bars. if we add more we won't need to migrate anything and if a layout doesn't have a 2nd bar it should be fine.

size is either width or height depending on the layout, which is why i left it as size.

that's not unreasonable, but it seems like saving a lot of extra data to state for no real reason. the user isn't expected to change layout often or at all once they have a workflow they like

holiber:
LGTM 👍

PR #2336, opened 2019-12-04
clipboard: don't allow to duplicate sources with `donNotDuplicateFlag`

holiber:


app/services/clipboard/clipboard.ts
```diff
@@ -197,6 +197,18 @@ export class ClipboardService extends StatefulService<IClipboardState>
...
 
+  canDuplicate(): boolean {
+    if (this.hasItemsInUnloadedClipboard()) return true;
+    if (!this.hasItems()) return false;
+    const hasNoduplicapableSource =
+      this.scenesService
+        .getScene(this.state.itemsSceneId)
+        .getSelection(this.state.sceneNodesIds)
+        .getSources()
+        .filter(source => source.doNotDuplicate).length > 0;
```
gettinToasty: not blocking but you can do `.some(source => source.doNotDuplicate)` here to make it a little easier to parse

PR #2340, opened 2019-12-04
Add restream onboarding step

gettinToasty:


app/components/pages/Onboarding.tsx
```diff
@@ -80,6 +83,9 @@ export default class OnboardingPage extends TsxComponent<{}> {
       ) {
         this.stepsState.push({ complete: false });
       }
+      if (this.onboardingService.isTwitchAuthed) {
```
andycreeth: This is a little confusing, but if this is the logic that enables the multistream step can we change this to check `restreamService.canEnableRestream` instead?  I know the only thing it checks right now is whether the user is Twitch authed, but I think it's good to have all restream UI looking at the same getter.

app/components/pages/onboarding-steps/Multistream.tsx
```diff
@@ -0,0 +1,127 @@
...
+import { UserService } from 'services/user';
+import { NavigationService } from 'services/navigation';
+import { $t } from 'services/i18n';
+import { RestreamService } from 'services/restream';
+import { StreamSettingsService } from 'services/settings/streaming';
+import { ToggleInput } from 'components/shared/inputs/inputs';
+import styles from './Multistream.m.less';
+
+class MultistreamProps {
+  continue: () => void = null;
```
andycreeth: Is `null` the correct default prop here or should it be `() => {}`?

app/components/pages/Onboarding.tsx
gettinToasty: i guess that makes sense, basically futureproofing against expanding restream to other platforms?

app/components/pages/onboarding-steps/Multistream.tsx
gettinToasty: not sure it makes a huge difference since it's a required prop but i can make that change

andycreeth: Ah ok.  Why is a default value needed at all if it's a required prop?

gettinToasty: the way `createProps` works we need any prop to have non-undefined values

PR #2354, opened 2019-12-09
Dont show no pages warning if a user has pages

gettinToasty:


andycreeth:
This should be based against preview

PR #2356, opened 2019-12-10
Add Layout Editor

gettinToasty:


app/components/editor/layouts/Classic.tsx
```diff
@@ -0,0 +1,56 @@
...
+import BaseLayout, { LayoutProps } from './BaseLayout';
+import { createProps } from 'components/tsx-component';
+import { Component } from 'vue-property-decorator';
+import ResizeBar from 'components/shared/ResizeBar.vue';
+import styles from './Layouts.m.less';
+
+@Component({ props: createProps(LayoutProps) })
+export default class Classic extends BaseLayout {
+  mounted() {
+    super.mountResize();
```
andycreeth: Why not `this.mountResize`?  Seems incorrect to use super in this case, since you should allow `mountResize` to potentially be overridden.  (This comment applies to all layout implementations)

app/components/editor/layouts/Classic.tsx
```diff
@@ -0,0 +1,56 @@
...
+    this.$emit('totalWidth', ['1', ['2', '3', '4']]);
+  }
+  destroyed() {
+    super.destroyResize();
```
andycreeth: Same comment.

app/services/layout.ts
```diff
@@ -15,63 +22,135 @@ export enum ELayoutElement {
...
+  [ELayout.OnePane]: { bar1: 500 },
+};
+
+const ELEMENT_MINS = {
+  [ELayoutElement.Display]: { x: 350, y: 250 },
+  [ELayoutElement.LegacyEvents]: { x: 430, y: 150 },
+  [ELayoutElement.Mixer]: { x: 230, y: 150 },
+  [ELayoutElement.Minifeed]: { x: 330, y: 150 },
+  [ELayoutElement.Sources]: { x: 220, y: 150 },
+  [ELayoutElement.Scenes]: { x: 200, y: 150 },
```
andycreeth: I think a lot of this stuff should really be pushed into the element classes themselves.  If they all subclass a common abstract class or implement a common interface, you can require all elements to define these values in their implementation.  This is cleaner and is better OO design, because you are clearly defining the requirements/shape of what a element implementation looks like.  It's much easier for a developer down the road to add a new element.  I think this comment also applies to layouts.  Things like the name of the layout should be an attribute on the layout itself rather than in a map object.

PR #2359, opened 2019-12-10
strictNullsCheck migration: core

holiber:
To enable strict nulls checking for files in the next PR you need to add a new json file into the `strict-null-check-files` folder with next shape:
```
{
  "ts": [
    "app/services/service-name/**/*.ts"
  ],
  "tsx": [
    "app/components/component-name/**/*.tsx"
  ]
}
```

### DIFF
--- a/app/services/core/persistent-stateful-service.ts
+++ b/app/services/core/persistent-stateful-service.ts
@@ -20,7 +20,7 @@ export abstract class PersistentStatefulService<TState extends object> extends S
   static defaultState = {};
 
   static get initialState() {
-    const persisted = JSON.parse(localStorage.getItem(this.localStorageKey)) || {};
+    const persisted = JSON.parse(localStorage.getItem(this.localStorageKey) as string) || {};
 
     return merge({}, this.defaultState, persisted);
   }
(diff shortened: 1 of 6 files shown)


app/services/core/persistent-stateful-service.ts
```diff
@@ -20,7 +20,7 @@ export abstract class PersistentStatefulService<TState extends object> extends S
   static defaultState = {};
 
   static get initialState() {
-    const persisted = JSON.parse(localStorage.getItem(this.localStorageKey)) || {};
+    const persisted = JSON.parse(localStorage.getItem(this.localStorageKey) as string) || {};
```
andycreeth: Is type casting really the right way to handle this?  `localStorage.getItem` really can return `null`.

holiber: The right way here is to do nothing.
`JSON.parse` works with `null` without exceptions. However, most of native .ts` files don't assume that we use the `strictNullsCheck` mode

andycreeth: Ah yes I see.  `JSON.parse(null)` just returns `null`.

PR #2365, opened 2019-12-12
Fix for selective recording test

summeroff:
Software encoding in tests with default setting takes more than 100% of CPU on azure CI servers. It leads to longer stopping of recording as it have to complete processing recorded frames before stopping. 
Next libobs release has it even slower, and it may exceed 15sec. 

Timeouts for stopping of recording increased to make tests pass.

### DIFF
--- a/test/recording.ts
+++ b/test/recording.ts
@@ -34,7 +34,7 @@ test('Recording', async t => {
 
     // Stop recording
     await app.client.click('.record-button');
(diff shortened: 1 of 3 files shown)


holiber:
Try to setup a lower resolution

```
// set low resolution to prevent intensive CPU usage
await setOutputResolution(t, '100x100');
```

summeroff:
This test fix has been merged in another PR together with OSN update what triggers tests timeouts.

PR #2374, opened 2019-12-13
linter: migrate tslint to eslint

holiber:
I disabled some new ESlint's rules that affects many files to be invalid.
If some rule affects not a big amount of files than I fixed them or added `eslint-disable` directive


app/components/shared/Hotkey.vue.ts
```diff
@@ -45,14 +45,15 @@ export default class HotkeyComponent extends TsxComponent<HotkeyProps> {
...
+      /*eslint-disable*/
       event instanceof MouseEvent
         ? {
             1: 'MiddleMouseButton',
             3: 'X1MouseButton',
             4: 'X2MouseButton',
           }[event.button]
         : event.code;
-
+      /*eslint-enable*/
```
andycreeth: Will the `eslint-disable` run until the end of the file if this is accidentally removed?  Seems a little dangerous, but hopefully eslint throws an error if you omit the closing tag?

app/components/windows/EditStreamInfo.vue.ts
```diff
@@ -144,10 +144,12 @@ export default class EditStreamInfo extends Vue {
         required: true,
         disabled: this.updatingInfo,
         description: this.isFacebook
+          /* eslint-disable */
```
andycreeth: What doesn't eslint like about this?  Can't we just format it the way it wants?

app/services/api/jsonrpc/jsonrpc.ts
```diff
@@ -107,6 +116,7 @@ export class JsonrpcService extends Service implements IJsonrpcServiceApi {
     data: any;
     isRejected?: boolean;
   }): IJsonRpcResponse<IJsonRpcEvent> {
+    // eslint-disable-next-line
     return JsonrpcService.createResponse.apply(this, arguments);
```
andycreeth: Curious why we need to disable so much in this file.

.eslintrc.js
```diff
@@ -0,0 +1,115 @@
...
+    // We do imports where our files are suffixed .vue and this rule would expect we import as e.g. MainVue
+    "import-name": OFF,
+    // TODO: fix instances instead
+    "no-increment-decrement": OFF,
+    // This is no longer required on refactor/mutation-names branch TODO: remove
+    "function-name": OFF,
+    "no-boolean-literal-compare": OFF,
+
+    // these rules have been disable after the migration from tslint
+    // TODO: make decision about what rules we should enable
```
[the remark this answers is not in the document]
holiber: yes. If you enable one of these rules you will see many errors in the codebase

.eslintrc.js
```diff
@@ -0,0 +1,115 @@
...
+    parser: '@typescript-eslint/parser',
+    sourceType: 'module',
+  },
+  plugins: [
+    'vue',
+    '@typescript-eslint',
+    'prettier'
+  ],
+  rules: {
+    // TODO: enable prettier when the prettier plugin will support optional chaning
```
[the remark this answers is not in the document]
holiber: should be fixed

app/components/windows/EditStreamInfo.vue.ts
holiber: `ESlint` requires 12 spaces of indentation but `Eslint:prettier` plugin requires 14 spaces here.
That looks like conflicting rules that I can't fix in the code. This needs more investigation in the prettier config, but it seems like it's only one place in the code where we have this conflict.

app/services/api/jsonrpc/jsonrpc.ts
holiber: Using of `arguments` is not allowed. Each disabled line contains the `arguments` keyword

app/components/shared/Hotkey.vue.ts
holiber: Unfortunately, it doesn't throw an error

PR #2388, opened 2019-12-17
new feature: auto game capture

holiber:


app/components/StudioEditor.vue.ts
```diff
@@ -358,6 +359,16 @@ export default class StudioEditor extends TsxComponent {
       rect.crop,
       { x: rect.x, y: rect.y },
     );
+
+    // set the game_capture's auto resize to false
+    if (source.type === 'game_capture') {
+      this.editorCommandsService.executeCommand('EditSourcePropertiesCommand', source.id, [
```
andycreeth: This shouldn't be done as a command.  This will result in a new entry in the undo history, which will not combine with the command above it.  This breaks the "rule" of a single user action resulting in a single new entry in the undo history.  I think unfortunately we will need to roll this behavior into the `CropItemsCommand` if it is required whenever we crop the game capture.

app/services/sources/sources.ts
```diff
@@ -327,6 +328,13 @@ export class SourcesService extends StatefulService<ISourcesState> implements IS
       resolvedSettings.video_device_id = this.defaultHardwareService.state.defaultVideoDevice;
     }
 
+    // setup the list of supported games for the auto mode in the game_capture
+    if (type === 'game_capture') {
+      resolvedSettings.auto_capture_list_path = `${electron.remote.app.getPath(
+        'userData',
+      )}/game_capture_list.lst`;
+    }
```
andycreeth: What happens if this file doesn't exist at this point?  Does the game capture gracefully handle this?  Will it find the list if it comes into existence later?  What if the file is only partially downloaded?  (We are not awaiting the download, not sure if that was intentional)

summeroff: game capture expect for a file to be readable at the moment settings set. it is better to set this option again after it downloaded.

andycreeth: K.  We definitely need to be waiting on this download then, and it needs to be done much earlier in the initialization process.

app/services/app/app.ts
```diff
@@ -103,9 +103,20 @@ export class AppService extends StatefulService<IAppState> {
...
+      // associated with the user in sentry.
+      this.userService.validateLogin(),
+
+      // this config should be downloaded before any game-capture source has been added to the scene
+      this.downloadAutoGameCaptureConfig(),
+    ]).catch(e => {
+      // probably the internet is disconnected
+    });
+
+    this.downloadAutoGameCaptureConfig();
```
andycreeth: I don't think we want to be doing this twice.

holiber: Good catch. Updated

app/services/sources/sources.ts
```diff
@@ -327,6 +329,21 @@ export class SourcesService extends StatefulService<ISourcesState> implements IS
...
+      // setup the list of supported games for the auto mode in the game_capture
+      resolvedSettings.auto_capture_list_path = path.join(
+        electron.remote.app.getPath('userData'),
+        'game_capture_list.lst',
+      );
+
+      // setup the placeholder image
+      resolvedSettings.auto_placeholder_image = path.join(
+        electron.remote.app.getAppPath(),
+        'media/images/game-capture/capture-placeholder.png',
```
andycreeth: Pretty sure this needs to be `require`d to work with the app when it is packaged.

PR #2392, opened 2019-12-20
Fix Facebook scheduling error message

gettinToasty:


app/services/platforms/facebook.ts
```diff
@@ -16,6 +16,7 @@ import { IListOption } from '../../components/shared/inputs';
 import { $t } from 'services/i18n';
 import { StreamSettingsService } from 'services/settings/streaming';
 import { Subject } from 'rxjs';
+import { promises } from 'dns';
```
andycreeth: Is this an accidental import?

gettinToasty: yes lol

PR #2423, opened 2020-01-13
strictNullsCheck migration: sources and scenes

holiber:


app/services/scenes/scene-folder.ts
```diff
@@ -26,16 +26,14 @@ export class SceneItemFolder extends SceneItemNode {
 
     const state = this.scenesService.state.scenes[sceneId].nodes.find(item => {
       return item.id === id;
-    });
+    }) as ISceneItemFolder;
```
andycreeth, line 29: Wouldn't it be better to handle `undefined` rather than just casting here?

app/services/scenes/scene-item.ts
```diff
@@ -93,11 +93,11 @@ export class SceneItem extends SceneItemNode {
   }
 
   getScene(): Scene {
-    return this.scenesService.views.getScene(this.sceneId);
+    return this.scenesService.views.getScene(this.sceneId) as Scene;
```
andycreeth, line 96: Shouldn't we just modify the return type to `Scene | undefined` if we're not handling the `undefined` case here?

app/services/scenes/scene-item.ts
```diff
@@ -93,11 +93,11 @@ export class SceneItem extends SceneItemNode {
...
   }
 
   get source() {
-    return this.sourcesService.views.getSource(this.sourceId);
+    return this.sourcesService.views.getSource(this.sourceId) as Source;
```
andycreeth, line 100: Same

app/services/scenes/scene.ts
```diff
@@ -71,39 +72,39 @@ export class Scene {
...
-    return this.getNodes().find(node => node.name === name);
+  getNodeByName(name: string): TSceneNode | null {
+    return this.getNodes().find(node => node && node.name === name) || null;
   }
 
   getItems(): SceneItem[] {
     return this.state.nodes
       .filter(node => node.sceneNodeType === 'item')
-      .map(item => this.getItem(item.id));
+      .map(item => this.getItem(item.id)) as SceneItem[];
```
andycreeth, line 95: Is this cast removing `null`s?  If so can we use a `compact` or something?

app/services/scenes/scene.ts
```diff
@@ -71,39 +72,39 @@ export class Scene {
...
   }
 
   getFolders(): SceneItemFolder[] {
     return this.state.nodes
       .filter(node => node.sceneNodeType === 'folder')
-      .map(item => this.getFolder(item.id));
+      .map(item => this.getFolder(item.id)) as SceneItemFolder[];
```
andycreeth, line 101: Same

app/services/scenes/scene-folder.ts
```diff
@@ -26,16 +27,20 @@ export class SceneItemFolder extends SceneItemNode {
 
     const state = this.scenesService.state.scenes[sceneId].nodes.find(item => {
       return item.id === id;
-    });
+    })!;
```
andycreeth, line 30: I think it's incorrect to assert not-null here.  It is entirely possible that I could do `new SceneItemFolder(sceneId, badNodeId)`.  We should handle with a runtime exception.

app/services/sources/sources.ts
```diff
@@ -406,7 +406,9 @@ export class SourcesService extends StatefulService<ISourcesState> {
   }
 
   refreshSourceAttributes() {
-    const activeItems = this.scenesService.views.activeScene.getItems();
+    const activeScene = this.scenesService.views.activeScene as Scene;
```
andycreeth, line 409: I think this cast can be removed now since you're handling the null.

app/services/sources/sources.ts
```diff
@@ -241,7 +240,7 @@ export class SourcesService extends StatefulService<ISourcesState> {
      * otherwise OBS thinks it's still attached
      * and won't release it. */
     if (source.channel !== void 0) {
-      obs.Global.setOutputSource(source.channel, null);
+      obs.Global.setOutputSource(source.channel, (null as unknown) as obs.ISource);
```
andycreeth, line 243: We probably need to update the type definitions inside OSN for this one since `null` is a valid value.  No action here, this is fine for now.

app/services/sources/sources.ts
```diff
@@ -55,13 +55,13 @@ interface IObsSourceCallbackInfo {
 
 class SourcesViews extends ViewHandler<ISourcesState> {
   get sources(): Source[] {
-    return Object.values(this.state.sources).map(sourceModel =>
-      this.getSource(sourceModel.sourceId),
+    return Object.values(this.state.sources).map(
+      sourceModel => this.getSource(sourceModel.sourceId) as Source,
```
andycreeth, line 59: Should be `!`?

app/services/sources/sources.ts
```diff
@@ -72,7 +72,7 @@ class SourcesViews extends ViewHandler<ISourcesState> {
     const sourceModels = Object.values(this.state.sources).filter(source => {
       return source.name === name;
     });
-    return sourceModels.map(sourceModel => this.getSource(sourceModel.sourceId));
+    return sourceModels.map(sourceModel => this.getSource(sourceModel.sourceId) as Source);
```
andycreeth, line 75: Should be `!`?

app/services/sources/sources.ts
```diff
@@ -193,8 +193,7 @@ export class SourcesService extends StatefulService<ISourcesState> {
     const obsInput = obs.InputFactory.create(type, id, obsInputSettings);
 
     this.addSource(obsInput, name, options);
-
-    return this.views.getSource(id);
+    return this.views.getSource(id) as Source;
```
andycreeth, line 196: Should be `!`?

app/services/sources/sources.ts
```diff
@@ -213,7 +212,7 @@ export class SourcesService extends StatefulService<ISourcesState> {
       isTemporary: options.isTemporary,
       propertiesManagerType: managerType,
     });
-    const source = this.views.getSource(id);
+    const source = this.views.getSource(id) as Source;
```
andycreeth, line 215: Should be `!`?

PR #2430, opened 2020-01-14
Refactor minimum size handling in modular editor

gettinToasty:


app/components/editor/layouts/BaseLayout.ts
```diff
@@ -32,6 +37,37 @@ export default class BaseLayout extends TsxComponent<LayoutProps> {
...
+    bar2Slots?: ILayoutSlotArray,
+  ) {
+    const rest = await this.calculateMinimum(restSlots);
+    const bar1 = await this.calculateMinimum(bar1Slots);
+    const bar2 = await this.calculateMinimum(bar2Slots);
+    this.mins = { rest, bar1, bar2 };
+  }
+
+  async minsFromSlot(slot: LayoutSlot) {
+    await this.$nextTick();
```
holiber: this single line makes multiple methods here `async` and adds multiple `this.$nextTick()` calls to the `setMins` method. Can we avoid this line?

gettinToasty: that's the core tradeoff of this refactor-- if we want to access elements sizes from the element itself we need to wait a render cycle before being able to access the `componentInstance` of those components

app/components/editor/elements/BaseElement.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  get element() {
+    return <div />;
+  }
+
+  @Watch('height')
+  @Watch('width')
+  calcBelowMins() {
+    if (!this.$el) return;
+    this.belowMins = this.height + 20 < this.mins.y || this.width + 20 < this.mins.x;
```
andycreeth: What's the significance of 20?  Is that the size of the padding?  Can you add a comment explaining?

app/components/editor/layouts/BaseLayout.ts
```diff
@@ -32,6 +38,39 @@ export default class BaseLayout extends TsxComponent<LayoutProps> {
...
+    const rest = await this.calculateMinimum(restSlots);
+    const bar1 = await this.calculateMinimum(bar1Slots);
+    const bar2 = await this.calculateMinimum(bar2Slots);
+    this.mins = { rest, bar1, bar2 };
+  }
+
+  async minsFromSlot(slot: LayoutSlot) {
+    // Before we can access the componentInstance at least one render cycle needs to run
+    if (!this.firstRender) await this.$nextTick();
+    this.firstRender = true;
```
andycreeth: Kinda jank but can't think of a better way for now.

PR #2436, opened 2020-01-15
tsx translations

holiber:
The main Idea that you assign each part of string to a variable.
```
        {{ $t('[channelError]There was an error fetching your channel information. You can try') }}
        <a class="description-link" @click="populateInfo">{{
          $t('[channelError]fetching the information again')
        }}</a
        >, {{ $t('[channelError]or you can') }}
        <a class="description-link" @click="() => goLive(true)">{{ $t('[channelError]just go live.') }}</a>
```
And use this variable as a key in the json  dictionary

```
"channelError": "There was an error fetching your channel information. You can try <link> fetching the information again </link> or you can <link> just go live.</link>",
```


andycreeth:
We can merge this in if you resolve conflicts.  I like the XML tags approach better.

gettinToasty:
this is still an unsatisfactory change that doesn't solve the core issue imo which is it chunks up the language in a way that is non-configurable to the translator. consider the following:
`There was an error fetching your channel information. You can try <link>fetching the information again</link> or you can <link>just go live.</link>`
this works for this specific sentence but what if the sentence has to be changed to this in a foreign language?
`There was an error fetching your channel information. You can try these two options <link>fetching the information again</link> <link>just go live.</link>`
sentence rearrangement is a possibility especially for eastern languages and this will break translations in those cases since we don't responsively change the templating in the component itself

andycreeth:
@gettinToasty Good call.  I think the placeholder tags need to be numbered or labeled somehow.  So it wouldn't be `<link></link>` for all of them.  It would be `<1></1>` and `<2></2>` or even better, `<fetch-again-link></fetch-again-link>` and `<go-live-link></go-live-link>`

holiber:
close in favor of
https://github.com/stream-labs/streamlabs-obs/pull/2632

PR #2447, opened 2020-01-21
streaming: don't update stream settings when it's not needed

holiber:
Backend can't reconnect the stream if we change stream settings when we're live. We always do it for YT


app/services/settings/streaming/stream-settings.ts
```diff
@@ -94,6 +94,7 @@ export class StreamSettingsService extends PersistentStatefulService<IStreamSett
    * setup all stream-settings via single object
    */
   setSettings(patch: Partial<IStreamSettings>) {
+    console.log('save settings', patch);
```
gettinToasty: leftover console log

PR #2448, opened 2020-01-21
Fix crash on theme toggle while chat is open

gettinToasty:
We get into a bad state not properly awaiting chat reloading it's url when the theme changes which causes crashes on resize

### DIFF
--- a/app/services/chat.ts
+++ b/app/services/chat.ts
@@ -38,8 +38,6 @@ export class ChatService extends Service {
 
       if (!this.chatUrl) return;
 
-      this.customizationService.setLiveDockCollapsed(true);
-
       // chat url changed to a new valid url, init or reload chat
       if (oldChatUrl) {
         this.deinitChat();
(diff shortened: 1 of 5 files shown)


app/services/chat.ts
```diff
@@ -94,8 +94,9 @@ export class ChatService extends Service {
         nodeIntegration: false,
       },
     });
-
-    this.navigateToChat();
+    this.windowsService.updateStyleBlockers('main', true);
+    await this.navigateToChat();
+    this.windowsService.updateStyleBlockers('main', false);
```
andycreeth: What's the purpose of disabling style blockers here?  Is there a change this raises an exception and we get stuck with style blockers hidden?  Can this take a long time (servers time out after 60 seconds of loading) and we are stuck with style blockers hidden for that all that time?  This change makes me nervous.

gettinToasty: `navigateToChat` appears to catch errors from what i saw in the codebase, but the reason for this is rendering the browserWindow in chat while performing this navigation is what is causing the crashy state

andycreeth: It only catches 1 very specific type of error that happens when the page redirects, which is expected.  I think that style blockers is not the right thing to use here, and it needlessly hides many other things.  We should just hide the chat during this time.

gettinToasty: i asked if that was a viable fix in FE chat but got no responses 😅 will make that change

PR #2455, opened 2020-01-22
fix tests: Adding and removing some sources

holiber:


test/sources.js
```diff
@@ -39,9 +40,9 @@ test('Adding and removing some sources', async t => {
 
     await selectSource(t, sourceName);
     await clickRemoveSource(t);
-
-    t.false(await sourceIsExisting(t, sourceName));
+    await waitForSourceExist(t, sourceName, true);
   }
+  t.pass();
```
andycreeth: Is this needed?  There's still a true assertion in the for loop above, which means test should not run without assertions.

holiber: Yea, we can ditch it

PR #2464, opened 2020-01-28
use error message for outdated drivers

summeroff:
required osn with https://github.com/stream-labs/obs-studio-node/pull/603


app/services/streaming/streaming.ts
```diff
@@ -542,15 +542,23 @@ export class StreamingService extends StatefulService<IStreamingServiceState>
...
       } else {
         // -4 is used for generic unknown messages in OBS. Both -4 and any other code
         // we don't recognize should fall into this branch and show a generic error.
         if (info.error) {
           errorText = info.error;
         } else {
-          linkToDriverInfo = true;
           errorText = $t(
-            'An error occurred with the output. This is usually caused by out of date video drivers. Please ensure your Nvidia or AMD drivers are up to date and try again.',
+            'An error occurred with the output.',
```
gettinToasty: i feel like there's gotta be something better we can say here

EddyGharbi:
Will merge once OSN update is in.

PR #2465, opened 2020-01-29
buffer all mutations in the services_worker window

holiber:
Improve performance, get rid of resize junk and some console errors.

### DIFF
--- a/app/store/index.ts
+++ b/app/store/index.ts
@@ -7,6 +7,7 @@ import { ServicesManager } from '../services-manager';
 import { IMutation } from 'services/api/jsonrpc';
 import Util from 'services/utils';
 import { InternalApiService } from 'services/api/internal-api';
+import cloneDeep from 'lodash/cloneDeep';
 
 Vue.use(Vuex);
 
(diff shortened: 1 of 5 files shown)


app/store/index.ts
```diff
@@ -31,6 +31,20 @@ const plugins: any[] = [];
...
+ * Add mutation to the queue so we can send it to the renderer windows along with other
+ * pending mutations.
+ * This prevents multiple re-renders of Vue components for each single mutation.
+ */
+function sendMutationToRenderWindows(mutation: IMutation) {
+  mutationsQueue.push(mutation);
+  setTimeout(() => {
+    ipcRenderer.send('vuex-mutation', JSON.stringify(mutationsQueue));
+    mutationsQueue.length = 0;
+  });
```
gettinToasty: i think if we want this to sync with renders we might want to use `requestAnimationFrame` here since if a render takes more than 16ms this will cause a re-render

gettinToasty: not sure if that would matter in a cross-window case like this though

holiber: the main idea here is that one synchronous part of code in the worker window should cause only one re-rerender step of Vue components. So we don't call re-render for each mutation if they come as a group of mutations

PR #2472, opened 2020-01-30
Add FB Share button

gettinToasty:


app/components/windows/EditStreamInfo.vue.ts
```diff
@@ -489,6 +489,18 @@ export default class EditStreamInfo extends Vue {
     return !this.channelInfo && !this.infoError;
   }
 
+  get sharePageUrl() {
+    if (!this.isFacebook) return;
+    const pageName = this.facebookService.state.facebookPages.options.find(
+      option => option.value === this.channelInfo.facebookPageId,
+    )?.title;
+    const formattedPageName = pageName
```
holiber: `pageName` could be undefined but you're trying to call `match` on it

gettinToasty: in this case `this.channelInfo.facebookPageId` is directly derived from these options so that's actually impossible

gettinToasty: on second thought it's probably safer to guard against that

PR #2490, opened 2020-02-14
use local scale to make autofit in auto gamecapture mode

summeroff:
required osn PR https://github.com/stream-labs/obs-studio-node/pull/608


gettinToasty:
this looks like it turns on autofit for every game capture?

summeroff:
kind of. it will be handled inside obs code transparently for upper levels. 
it is not implement autofit it is also gonna keep fixed scale between games( which can have different window size ). 
so workaround with disabling autofit after user manually resize no more required.

gettinToasty:
@summeroff are the BE changes in for this?

summeroff:
not yet https://github.com/stream-labs/obs-studio-node/pull/608

andycreeth:
@summeroff Please go ahead and merge this into staging whenever the backend is updated.

EddyGharbi:
Will merge once OSN update is in

PR #2500, opened 2020-02-26
Draw extra frames for volmeters

holiber:
Render 60fps voltmeters via interpolation

### DIFF
--- a/app/components/MixerVolmeter.vue.ts
+++ b/app/components/MixerVolmeter.vue.ts
@@ -316,7 +316,7 @@ export default class MixerVolmeter extends TsxComponent<MixerVolmeterProps> {
     this.gl.uniform2f(this.scaleLocation, 1, CHANNEL_HEIGHT);
     this.gl.uniform2f(this.translationLocation, 0, channel * (CHANNEL_HEIGHT + PADDING_HEIGHT));
 
-    const prevPeak = this.prevPeaks ? this.prevPeaks[channel] || peak : peak;
+    const prevPeak = this.prevPeaks && this.prevPeaks[channel] ? this.prevPeaks[channel] : peak;
     const timeDelta = performance.now() - this.lastEventTime;
     let alpha = timeDelta / this.interpolationTime;
     if (alpha > 1) alpha = 1;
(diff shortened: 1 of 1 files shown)


app/components/MixerVolmeter.vue.ts
```diff
@@ -266,7 +292,15 @@ export default class MixerVolmeter extends TsxComponent<MixerVolmeterProps> {
 
     this.gl.uniform2f(this.scaleLocation, 1, CHANNEL_HEIGHT);
     this.gl.uniform2f(this.translationLocation, 0, channel * (CHANNEL_HEIGHT + PADDING_HEIGHT));
-    this.gl.uniform1f(this.volumeLocation, this.dbToUnitScalar(peak));
+
+    const prevPeak = this.prevPeaks ? this.prevPeaks[channel] || peak : peak;
```
gettinToasty: i think this can just be `this.prePeaks?[channel] || peak`

app/components/MixerVolmeter.vue.ts
```diff
@@ -266,7 +292,15 @@ export default class MixerVolmeter extends TsxComponent<MixerVolmeterProps> {
...
+    const timeDelta = performance.now() - this.lastEventTime;
```
gettinToasty: this looks like the first time `lastEventTime` is referenced but it's uninitialized, so wouldn't this be `performance.now() - undefined` in the first cycle?

app/components/MixerVolmeter.vue.ts
```diff
@@ -110,6 +128,38 @@ export default class MixerVolmeter extends TsxComponent<MixerVolmeterProps> {
...
+  }
+
+  /**
+   * Render volmeters with FPS capping
+   */
+  private onRequestAnimationFrameHandler() {
+    // init first rendering frame
+    if (!this.frameNumber) {
+      this.frameNumber = -1;
+      this.firstFrameTime = performance.now();
```
andycreeth: The callback to `requestAnimationFrame` is passed a `DOMHighResTimeStamp` in its arguments that is meant to be used for calculations like this.  Seems more logical and possibly higher performance to use that rather than making several `performance.now()` calls inside this function.

PR #2521, opened 2020-03-13
Add Prime Flow to Layout Editor v2

gettinToasty:


app/app.g.less
```diff
@@ -420,6 +420,22 @@ select {
...
+.sidenav-slide-enter,
+.sidenav-slide-leave-to {
+  transform: translateY(-20px);
+  opacity: 0;
+}
+
+// .sidenav-slide-enter-to {
+//   // transform: translateY(0);
+//   opacity: 1;
+// }
```
andycreeth: Should this be removed?

app/services/user/index.ts
```diff
@@ -125,6 +120,11 @@ export class UserService extends PersistentStatefulService<IUserServiceState> {
     Vue.delete(this.state, 'auth');
   }
```
andycreeth: This mutations should probably additionally set `isPrime` to false now.

app/services/user/index.ts
```diff
@@ -178,11 +180,17 @@ export class UserService extends PersistentStatefulService<IUserServiceState> {
...
     super.init();
     this.MIGRATE_AUTH();
     this.VALIDATE_LOGIN(false);
     this.SET_AUTH_STATE(EAuthProcessState.Idle);
+    await this.setPrimeStatus();
+    this.lifecycle = await this.withLifecycle({
+      init: this.subscribeToSocketConnection,
+      destroy: this.unsubscribeFromSocketConnection,
+      context: this,
+    });
```
andycreeth: This stuff should be moved out of `init`.  Nothing is waiting on this promise, and this isn't where we are currently fetching information about the current user.  This would be a better fit for the `login` method, which handles setting up the state of the currently logged in user, and is called every time the app starts up and every time a new user logs in.  It's also waited on early in the app startup process so you don't need to worry about race conditions like you do here.

PR #2532, opened 2020-03-23
Optimize re-order scene-items

summeroff:


holiber:
Close in favor of https://github.com/stream-labs/streamlabs-obs/pull/2608

PR #2537, opened 2020-03-26
List devices without activating

summeroff:
Vladimir Sumarov	Update libobs to v24.17.53-dna1


summeroff:
new backend implementation do not need this changes.

PR #2545, opened 2020-04-01
SLOBS optimizations

holiber:


app/services/scene-collections/scene-collections.ts
```diff
@@ -668,16 +670,12 @@ export class SceneCollectionsService extends Service implements ISceneCollection
     if (this.autoSavePromise) await this.autoSavePromise;
   }
 
-  private async setActiveCollection(id: string) {
+  private async setActiveCollection(id: string, informServer = true) {
```
andycreeth, line 673: What is `informServer doing?  I don't see it being used.

app/services/scene-collections/scene-collections.ts
```diff
@@ -668,16 +670,12 @@ export class SceneCollectionsService extends Service implements ISceneCollection
...
     const collection = this.collections.find(coll => coll.id === id);
 
     if (collection) {
       if (collection.serverId && this.userService.isLoggedIn) {
-        try {
-          await this.serverApi.makeSceneCollectionActive(collection.serverId);
-        } catch (e) {
-          console.warn('Failed setting active collection');
-        }
```
andycreeth, line 680: Is there a reason for removing the `try/catch` here?

electron-builder/base.config.js
```diff
@@ -37,7 +38,9 @@ const base = {
   },
   extraMetadata: {
     env: 'production'
-  }
+  },
+  afterPack: 'node_modules/streamlabs-obs-launcher/after_pack.js',
+  afterSign: 'node_modules/streamlabs-obs-launcher/after_sign.js'
```
andycreeth, line 43: Changes to this file should be removed from this PR.

updater/bootstrap.ts
```diff
@@ -182,53 +177,39 @@ async function fetchUpdater(info: IUpdateInfo): Promise<string | null> {
...
-  if (response.statusCode !== 200) {
-    console.log(`Failed to fetch version information - ${response.statusCode}`);
+  const response = await fetch(`${info.baseUrl}/${info.versionFileName}`);
 
+  if (response.status !== 200) {
+    console.log(`Failed to fetch version information - ${response.status}`);
     return null;
   }
 
-  return response.body;
```
andycreeth, line 231: Can you extract all changes to `bootstrap.ts` into a separate PR?  This is the riskiest file in the entire app to change as if it's broken we cannot push new updates.  Any modifications to this file require extremely extensive testing.  I want to do that testing - but I'm not sure we have time to give it the attention it deserves for the next release.

app/services/scene-collections/scene-collections.ts
holiber: Forgot to delete it from some experiments

app/services/scene-collections/scene-collections.ts
holiber: After removing "await" keyword it doesn't make sense. I'll add ".catch()" instead

andycreeth: Oh of course.  Wasn't reading the change correctly.  ty.

PR #2548, opened 2020-04-02
Minimal bundle update infrastructure.

andycreeth:
Allows updating of JS bundles outside normal release cycle.


updater/bundle-updater.ts
```diff
@@ -0,0 +1,95 @@
+import * as request from 'request';
```
holiber, line 1: Can you switch to 'node-fetch'? I'm trying to get rid of 'request' library

andycreeth: I'll refactor it to node-fetch once your PR adding it is merged, which I think we can probably go ahead and merge now.

PR #2549, opened 2020-04-02
bootsrap.ts: Use "node-fetch" instead of "request"

holiber:
This change decreases the startup time


updater/bootstrap.ts
```diff
@@ -182,53 +167,39 @@ async function fetchUpdater(info: IUpdateInfo): Promise<string | null> {
...
-      })
-      .pipe(outStream);
-
-    outPipe.on('close', () => {
-      resolve(updaterPath);
-    });
-
-    outPipe.on('error', error => {
-      reject(error);
+    fetch(`${info.baseUrl}/${updaterName}`).then(response => {
```
andycreeth, line 178: If the promise returned by `fetch` rejects, then the outer promise never resolves or rejects.  It's the same bug that you fixed with the `downloadFile` infinite loader.

holiber: Good catch

andycreeth:
Reopening to get tests triggered.

PR #2584, opened 2020-04-16
external-api: improvements for streamdeck

holiber:
- Add `@Expensive` decorator to limit the frequency of some heavy API calls
- Optimize dragging and resize API events
- Add tests for streamdeck

### DIFF
--- a/test/api/scenes.ts
+++ b/test/api/scenes.ts
@@ -208,8 +208,11 @@ test('SceneItem.addFile()', async t => {
   const scenesService = client.getResource<ScenesService>('ScenesService');
   const scene = scenesService.activeScene;
 
+  console.log('inited');
   scene.clear();
+  console.log('cleared');
   scene.addFile(dataDir);
+  console.log('added');
 
   t.true(
     sceneBuilder.isEqualTo(`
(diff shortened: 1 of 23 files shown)


test/api/scenes.ts
```diff
@@ -227,3 +227,11 @@ test('SceneItem.addFile()', async t => {
   `),
   );
 });
+
+test('Try to make a not existing scene active', async t => {
+  const client = await getClient();
+  const scenesService = client.getResource<ScenesService>('ScenesService');
+  await sleep(30000);
```
andycreeth, line 234: Yikes 30 seconds?  I hate to see a single test taking this long.  Is this valuable enough for that?

test/streamdeck.ts
```diff
@@ -0,0 +1,320 @@
+import { test, TExecutionContext, useSpectron } from './helpers/spectron';
+import { ApiClient, getClient } from './helpers/api-client';
+import { IJsonRpcResponse } from '../app/services/api/jsonrpc';
+import { cloneDeep } from 'lodash';
+
+useSpectron({ pauseIfFailed: true });
+
+// Test all API requests that streamdeck device does right after connection
+// tested on Streamdeck v4.5.0
+test('Connect streamdeck device', async t => {
```
andycreeth, line 10: Thanks for writing such thorough tests

test/api/scenes.ts
holiber: It seems like my last `git push` that removes that failed. Should be fixed

PR #2594, opened 2020-04-22
internal-api: throw a runtime error if using destroyed objects

holiber:
This code now throws a human-readable error:
```ts
const scene = scenesService.getScene(id);
scene.remove();
scene.getItems(); // <-- throws "Trying to call the method "${key}" on destroyed object "${this._resourceId}"
```


app/services/api/external-api/scenes/selection.ts
```diff
@@ -67,6 +71,7 @@ export class Selection implements ISceneItemActions {
    * Select items. Previous selected items will be unselected
    */
   select(ids: string[]): Selection {
+    console.log('try to select');
```
andycreeth, line 74: Remove?

PR #2604, opened 2020-04-28
addFile was using hardcoded Windows separators and doesn't work on Macs.

ccapndave:
I found that adding a directory using the remote API on Mac doesn't work because the separator is wrong.  I've updated it to use the `path` module so that its cross-platform.


andycreeth:
We're in kind of a complicated branching situation with mac right now (should be resolved soon).  In the meantime, I changed the base branch to `mac-andy` which will allow it go through our normal QA channels before hitting the stable branch.  Can you rebase and then I'll be able to merge.  Thanks!

ccapndave:
I've rebased it.  I think :)  Let me know if there is a problem.  Cheers!

andycreeth:
Perfect, thanks!

PR #2607, opened 2020-04-28
optimize scene items

holiber:
frontend-only optimizations for scene-items

- Use a dictionary as storage for `sceneNodes`
- Cache instances of `sceneNodes` to minimize the amount of `new` calls
- reduce the amount of redundant `flushMutations` calls

### DIFF
--- a/app/services/scenes/scenes.ts
+++ b/app/services/scenes/scenes.ts
@@ -5,7 +5,7 @@ import { Subject } from 'rxjs';
 import { mutation, StatefulService } from 'services/core/stateful-service';
 import { TransitionsService } from 'services/transitions';
 import { WindowsService } from 'services/windows';
-import { Scene, SceneItem } from './index';
+import { Scene, SceneItem, TSceneNode } from './index';
 import { ISource, SourcesService, ISourceAddOptions } from 'services/sources';
 import { Inject } from 'services/core/injector';
 import * as obs from '../../../obs-api';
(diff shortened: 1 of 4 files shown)


app/services/scenes/scenes.ts
```diff
@@ -199,12 +215,31 @@ export class ScenesService extends StatefulService<IScenesState> {
   @Inject() private sourcesService: SourcesService;
   @Inject() private transitionsService: TransitionsService;
 
+  protected init() {
+    // subscribe to itemAdded and itemRemoved event to sync cachedNodes
+    this.itemAdded.subscribe(itemModel => {
+      this.addItemToCache(itemModel.sceneId, itemModel.id);
+    });
```
andycreeth, line 222: Do we really want to be proactively saturating the cache?  Isn't this expensive?

holiber: I found that we a lot of calls `new SceneItem` when loading a scene-collection. We have 2 `applyProxy` method in the constructor that takes some significant time because it iterates through each property of SceneItem.

andycreeth: Yeah I understand that.  Normally with a cache though you wouldn't add an object to the cache until it is being requested.  Here you are adding it as soon as it is created.  Maybe we always immediately request the object so it's effectively the same.  But the behavior struck me as weird.

PR #2608, opened 2020-04-28
Optimize nodes re-order

holiber:
Use optimized backend re-ordering for SceneNodes


package.json
```diff
@@ -213,4 +213,4 @@
   "optionalDependencies": {
     "node-win32-np": "1.0.6"
   }
-}
\ No newline at end of file
+}
```
andycreeth, line 216: Can you remove this change please

PR #2611, opened 2020-04-29
Added performance mode option in source menu

apap04:
Just like in OBS, you can enable/disable performance mode by right-clicking on the source instead of the scene. This is easier instead of doing this on the scene.


apap04:
Sounds good, thanks! 🙂

PR #2614, opened 2020-04-30
screentests: fix githubcheck

holiber:


test/screentest/runner.js
```diff
@@ -22,6 +22,9 @@ const args = process.argv.slice(2);
 
 (async function main() {
 
+  console.log('pause');
+  await new Promise(r => setTimeout(r));
```
andycreeth, line 26: Can you add a comment on why this defer is needed?

holiber: it's here from another stashed change. Will remove it

holiber:
Close in favor of https://github.com/stream-labs/streamlabs-obs/pull/2633

PR #2650, opened 2020-05-08
Fix resizing from the top of the screen

andycreeth:
I think the drag region was changed to the entire titlebar component recently.  We can't do this because if the drag region extends fully to the edge of the window it becomes impossible to resize the window from that edge.


gettinToasty:
this change was made to support dragging on the prime confirmation window, which doesn't have a title and as a result isn't draggable. i can look into an alternative solution!

PR #2668, opened 2020-05-18
Wrap donation messages in minifeed

gettinToasty:


app/components/RecentEvents.m.less
```diff
@@ -101,6 +105,7 @@
   opacity: 0.7;
   overflow: hidden;
   text-overflow: ellipsis;
+  white-space: normal !important;
```
andycreeth, line 108: What rule is this overriding?  Something global?

gettinToasty: `.cell span` which beats it in specificity slightly but applies to every other span in the cell

PR #2680, opened 2020-05-22
external-api: request token for local websoket clients

holiber:


holiber:
replaced with https://github.com/stream-labs/streamlabs-obs/pull/2687

PR #2694, opened 2020-05-28
sentry: save the API request

holiber:
Save API request and the Id of the window that emitted this request.

### DIFF
--- a/app/services/api/jsonrpc/jsonrpc-api.ts
+++ b/app/services/api/jsonrpc/jsonrpc-api.ts
@@ -20,6 +20,7 @@ export interface IJsonRpcRequest {
     fetchMutations?: boolean;
     compactMode?: boolean;
     noReturn?: boolean;
+    windowId?: string;
   };
 }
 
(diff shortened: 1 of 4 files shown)


app/services/api/jsonrpc/jsonrpc-api.ts
```diff
@@ -20,6 +20,7 @@ export interface IJsonRpcRequest {
     fetchMutations?: boolean;
     compactMode?: boolean;
     noReturn?: boolean;
+    windowId?: string;
```
andycreeth, line 23: Why are we adding this to the interface when I don't see it used anywhere?

holiber: It uses here https://github.com/stream-labs/streamlabs-obs/pull/2694/files#diff-6287f3fd9c57c5fc50831cd0ad9ca72aR146

andycreeth: Don't know how I missed that haha

PR #2707, opened 2020-06-01
external API: update docs

holiber:


package.json
```diff
@@ -202,4 +202,4 @@
     "webpack-manifest-plugin": "2.2.0",
     "webpack-merge": "4.2.1"
   }
-}
\ No newline at end of file
+}
```
andycreeth, line 205: Why is this change in here?

holiber: I had to reinstall `typedock` and then manually removed it from the `package.json`

holiber: I bet my IDE applied to autoformatting after that

andycreeth: Next time please remove the change before merging.

PR #2709, opened 2020-06-02
Revert "Revert external API streamdeck changes"

holiber:
Reverts stream-labs/streamlabs-obs#2705 + fixes and tests


package.json
```diff
@@ -202,4 +202,4 @@
     "webpack-manifest-plugin": "2.2.0",
     "webpack-merge": "4.2.1"
   }
-}
\ No newline at end of file
+}
```
andycreeth, line 205: Please remove this change before merging.  Your IDE is fighting with our release script and it's going to really muddy up the git history.

test/api/sources.ts
```diff
@@ -57,8 +58,12 @@ test('Source events', async t => {
...
-  event = await client.fetchNextEvent();
-  t.is(event.data.name, 'audio3');
+  // // check `sourceUpdated` event when renaming a source
+  // source1.setName('audio3');
+  // event = await client.fetchNextEvent();
+  //
+  // // the remote control app requires these fields to be in the event
+  // t.is(event.data.name, 'audio3');
+  // t.is(event.data.configurable, true);
+  // t.truthy(event.data.resourceId);
```
andycreeth, line 68: Is there a reason we're keeping these but commented out?  Seems like this is supposed to be testing for the regression you just fixed.

holiber: Didn't push the last commit. Good catch

PR #2726, opened 2020-06-05
Fix synchronous IPC while resizing

gettinToasty:


app/services/layout/index.ts
```diff
@@ -56,6 +56,42 @@ class LayoutViews extends ViewHandler<ILayoutServiceState> {
...
+  }
+
+  aggregateMinimum(orientation: 'x' | 'y', slots: IVec2Array) {
+    const minimums = slots.map(mins => {
+      if (mins) return mins[orientation];
+      return 10;
+    });
+    if (!minimums.length) minimums.push(10);
+    return minimums.reduce((a: number, b: number) => a + b);
+  }
```
andycreeth, line 94: Nice.  I didn't really anticipate views being used for shared logic that isn't based on state (i.e. pure helper functions) but I think it's a fine pattern.  Another pattern (which we don't utilize enough) is just exporting a function from a services file.  If it's just a pure function that doesn't interact with any state at all, it's totally fine to just define the function outside the service and export it.

PR #2739, opened 2020-06-16
Fix Add/Remove Scene at small window sizes

gettinToasty:


gettinToasty:
only relevant change is in the CSS, rest of it is just prettier auto-format

PR #2750, opened 2020-06-22
Add active state to selective recording icon in prime themes

gettinToasty:


app/services/layout/layout-data.ts
```diff
@@ -37,49 +37,49 @@ type ILayoutData = {
 
 export const LAYOUT_DATA: ILayoutData = {
   [ELayout.Default]: {
-    resizeDefaults: { bar1: 156, bar2: 240 },
+    resizeDefaults: { bar1: 20, bar2: 40 },
```
holiber, line 40: why the numbers are so different. Are they percentages now?

gettinToasty: whoops those changes werent meant to be carried over, will fix

app/components/editor/elements/SourceSelector.vue
```diff
@@ -135,7 +157,7 @@ i.disabled {
 }
 
 .icon--active {
-  color: var(--teal);
+  color: var(--icon-active);
```
andycreeth, line 160: Lol it's maybe slightly counterintuitive that `--teal` is set to `@goldenrod` in `themes.g.less`.  We should probably rename that variable at some point.

gettinToasty: yea i don't disagree-- maybe something like `--brand-primary` or `--primary-accent` would be better

PR #2754, opened 2020-06-25
Fix/Refactor Alert and EventList Animations

trevorrl:
Primary goal was to fix a bug where if you set your Alert animations to specific values, they would get bugged and play forever. I got a bit carried away, refactored the AnimationInput.vue.ts, but I fixed the bug :)


trevorrl:
> Am I right that the only difference between `inAnimations()` and `eventInAnimation()` is that one of them includes common animations?

There are other differences as well. Primarily the Event List has these "common" animations, but it also DOESN'T include a couple that are compatible with Alert Box (Fade In Right Big and Flip In specifically). 

I made these lists to mirror animations available on the site, and couldn't think of a smart way to combine them without accidentally including animations that weren't supported on one or the other.

PR #2757, opened 2020-06-30
Buffer mutations for actions that return

andycreeth:


andycreeth:
We replaced it because we need to avoid applying double mutations in both directions now - It is now possible that a synced mutation will arrive *before* the buffered mutations.  Yes it is technically a memory leak, but it should be small.  I will refactor out the leak eventually.

PR #2765, opened 2020-07-06
Add channel point notifications to multichat

GtgdJames:
Channel point rewards only visible in Twitch-only chat currently. Switching to multichat stops channel point notifications from showing.


andycreeth:
Hey - it looks like maybe you didn't open this PR with the correct branches.  I'm going to close this, but feel free to open up another PR if you have changes you would like to submit.

PR #2771, opened 2020-07-08
Improve crash analytics

summeroff:
required osn PR https://github.com/stream-labs/obs-studio-node/pull/687

### DIFF
--- a/app/services/crash-reporter.ts
+++ b/app/services/crash-reporter.ts
@@ -27,7 +27,7 @@ enum EAppState {
 
 interface ICrashReporterState {
   code: EAppState;
-  detected: string; // what was detected before close happens: ipc freez, backend crash, electron window unresponcive
+  detected: string; // The Backend event that was detected: IPC freeze or backend crash or electron window unresponsive.
   version: string; // SLOBS version
 }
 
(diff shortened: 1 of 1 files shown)


app/services/crash-reporter.ts
```diff
@@ -27,6 +27,7 @@ enum EAppState {
 
 interface ICrashReporterState {
   code: EAppState;
+  detected: string; // what was detected before close happens: ipc freez, backend crash, electron window unresponcive
```
EddyGharbi, line 30: I would rephrase the comment to: `The Backend event that was detected when the app close event was triggered: IPC freeze or backend crash or electron window unresponsive.`

app/services/crash-reporter.ts
```diff
@@ -94,7 +95,14 @@ export class CrashReporterService extends Service {
...
 
   private writeStateFile(code: EAppState) {
-    this.appState = { code, version: this.version };
+    this.appState = { code, version: this.version, detected: '' };
+    this.appState = this.readStateFile();
+    this.appState.code = code;
+    this.appState.version = this.version;
+    if (this.appState.code !== EAppState.Starting) {
+      this.appState.detected = '';
+    }
```
holiber, line 104: this could be simplified to
```
this.appState = {...this.readStateFile, code};
if (this.appState.code !== EAppState.Starting) {
      this.appState.detected = '';
}
```

summeroff: It does not let me to use ... in this way. 

Removed unnecessary preinitialization.

PR #2784, opened 2020-07-14
tests: fix the theme installation test

holiber:


andycreeth:
Merging this in now so everyone's test runs can benefit.  Thanks for fixing.

PR #2787, opened 2020-07-14
Port Streamlabels to new API

gettinToasty:


app/components/custom-source-properties/StreamlabelProperties.vue.ts
```diff
@@ -19,20 +23,23 @@ export default class StreamlabelProperties extends Vue {
...
+    return Object.values(this.definitions);
   }
 
+  definitions: IStreamlabelSet = null;
   currentlySelected: IStreamlabelDefinition = null;
   labelSettings: IStreamlabelSettings = null;
 
-  created() {
+  async created() {
+    this.definitions = await this.streamlabelsService.fetchDefinitions();
```
andycreeth, line 35: Just want to double check you've handled race conditions here.  Vue doesn't wait for any Promise returned by `created`.  It's totally fine to make it an `async` function so you can use `await` like this.  You just need to keep in mind that Vue will immediately continue with mounting the component.  So the component can fully render while this request is in flight.  Just wanted to double check that is handled.

gettinToasty: yep i have a guard clause in `statOptions` to make sure the rendering doesn't error out, and a semiredundant guard clause in `refreshPropertyValues` to do basically the same thing. thanks for double checking!

PR #2807, opened 2020-07-21
Better handling for "post a Tweet" issues

holiber:


app/components/windows/go-live/GoLiveChecklist.tsx
```diff
@@ -58,7 +58,11 @@ export default class GoLiveChecklist extends TsxComponent<Props> {
 
   private getHeaderText() {
     if (this.error) {
-      return $t('Something went wrong');
+      if (this.view.info.checklist.startVideoTransmission === 'done') {
+        return $t('You stream has been started with issues');
```
gettinToasty, line 62: this should probably be something more clear like `Your stream has started, but there were issues with other actions taken`

andycreeth: ^ Yeah Sean's suggestion is good.  Let's use that copy.

PR #2809, opened 2020-07-21
Fix yt grandfathering

andycreeth:


app/components/windows/go-live/GoLiveError.tsx
```diff
@@ -125,7 +125,9 @@ export default class GoLiveError extends TsxComponent<{}> {
     assertIsDefined(error.platform);
     const platformName = getPlatformService(error.platform).displayName;
     return (
-      <ErrorLayout message={$t('Multistream to YouTube is a Prime feature', { platformName })}>
+      <ErrorLayout
+        message={$t('Multistreaming to these platforms requires Prime', { platformName })}
```
holiber, line 129: `{ platformName }` is not needed now

PR #2823, opened 2020-07-23
remove mixer login

andycreeth:
image


holiber:
Does that mean we still keep logged in users?

andycreeth:
@holiber Yeah for now we are not actively logging out users.  But we have sent notifications to all users logged into mixer telling them what they need to do.  For now hopefully this is sufficient.

PR #2829, opened 2020-07-24
Get rid of services injectors in stream-view.ts

holiber:


app/services/restream.ts
```diff
@@ -69,26 +69,32 @@ export class RestreamService extends StatefulService<IRestreamState> {
...
+       * This determines whether the user can enable restream
+       * Requirements:
+       * - Has prime, or
+       * - Has a grandfathered status enabled
+       */
+      get canEnableRestream() {
+        const userView = this.getServiceViews(UserService);
+        return userView.isPrime || (userView.auth && this.state.grandfathered);
+      }
+    }
```
andycreeth, line 84: Can you explain why you're starting the pattern of defining the view handler inside the getter like this?

andycreeth: @holiber bump

holiber: I have only one method here and we don't export this class outside the file. So just don't see a big difference between declaring this class separately

andycreeth:
Can you rebase this onto master?  It only changes frontend files so it's best if it goes straight to master.

andycreeth:
Seems like there is now a merge conflict.

PR #2830, opened 2020-07-25
Minor proposal on lines 33 & 41

kant:


andycreeth:
Thanks for submitting this, but we are actually no longer on a fork of electron, so this document is no longer needed.  We will remove it.  Thanks.

PR #2854, opened 2020-08-06
Youtube stream category

holiber:
Select a category for YT stream


holiber:
Close in favor https://github.com/stream-labs/streamlabs-obs/pull/3038

PR #2855, opened 2020-08-06
Repair scene collection after theme installation

holiber:


andycreeth:
I'm now seeing the error immediately upon installing the theme on this branch.  I'm wondering if the act of repairing is actually causing us to log the error.

holiber:
Ther repair code has `console.log` inside. What kind of error do you see?

andycreeth:
I'm seeing the original error about a loop in folders structure

andycreeth (approved):

PR #2861, opened 2020-08-07
fix: remove the duration field from stinger transition

holiber:


andycreeth:
Can you rebase this onto master.

PR #2870, opened 2020-08-11
Add subcount viewer

kddove85:


holiber:
Hello. Could you please add a detailed description of the feature you're suggesting to implement here?

andycreeth:
Hi, thank you for your interest in contibuting to Streamlabs OBS.  While we can see why having a sub counter inside streamlabs would be useful to some people, we don't feel that adding it here fits with our product vision and roadmap, so I am closing your PR.  Thanks again.

PR #2875, opened 2020-08-12
Add reactive scrollbar

gettinToasty:


app/components/shared/HScroll.vue.ts
```diff
@@ -1,69 +0,0 @@
...
-  canScroll: boolean;
-  canScrollLeft: boolean;
-  canScrollRight: boolean;
-}
-
-/**
- * horizontal scrolling wrapper
- */
-@Component({})
-export default class HScroll extends Vue {
```
andycreeth, line 16: Never knew we had this component.  What was it used for?

gettinToasty: it was used when we had the top nav bar-- with the apps nav in that situation

PR #2876, opened 2020-08-13
Duplicate items with filters

holiber:


andycreeth:
This is an undo-able action, so you may need to make changes to the editor command as well.  Can you double check that.

holiber:
I tested the `undo` feature, it works

PR #2899, opened 2020-08-25
Manually handle sentry sampling, and send to logs

andycreeth:
Rather than trying to fight sentry eating all the exceptions, this lets them eat 100% of exceptions (with 100% sample rate), then we can tie into the `beforeSend` handler to send to the local log file.  Sample rate is implemented at the very end manually by returning null from the `beforeSend`.

Also fixes the `Hub.withScope` sentry error that resulted from calling `console.error` with a javascript object.

Also fixes an extremely high volume exception related to an extremely rare edge case where we cannot initialize the WebGL rendering for volmeters, but keep retrying over and over.


app/components/MixerVolmeter.vue.ts
```diff
@@ -118,21 +118,31 @@ export default class MixerVolmeter extends TsxComponent<MixerVolmeterProps> {
...
+
+        // Get ready to lose this conext if too many are created
+        if (window['activeWebglContexts'] == null) window['activeWebglContexts'] = 0;
+        window['activeWebglContexts'] += 1;
+        this.$refs.canvas.addEventListener('webglcontextlost', this.handleLostWebglContext);
+      } catch (e) {
+        console.error('Failed to initialize WebGL rendering, falling back to Canvas 2d', e);
+        this.gl = null;
+        this.initCanvas2dRendering();
+      }
```
holiber, line 132: This code will be not actual after merging Volmeters on a shared canvas

andycreeth: Good to know

PR #2907, opened 2020-08-28
Add Prime Expiration flow

gettinToasty:
Made it a part of onboarding just for organizational purposes, will only ever show as a singleton step though


app/components/pages/onboarding-steps/PrimeExpiration.tsx
```diff
@@ -0,0 +1,60 @@
...
+        >
+          <h2>{$t('Resubscribe to Prime')}</h2>
+        </div>
+        <a onClick={() => this.leavePage()} style="text-decoration: underline; margin-top: 16px;">
+          {$t('No Thanks')}
+        </a>
+      </div>
+    );
+  }
+}
```
holiber, line 60: Can you register new files in the `strict-null-check-files` folder

PR #2909, opened 2020-08-29
Prevent the stream from launching if machine is locked

AuroraAri:
In certain (very specific) situations, Streamlabs OBS can be commanded to start a stream while the host machine is locked. This should be prevented, and this PR aims to do so. By using Electron's [powerMonitor.getSystemIdleState](https://www.electronjs.org/docs/api/power-monitor), we can find out whether or not the machine is locked when the "go live" attempt is made, and if so, prevent it from doing so.

### DIFF
--- a/app/components/windows/go-live/GoLiveError.tsx
+++ b/app/components/windows/go-live/GoLiveError.tsx
@@ -240,7 +240,7 @@ export default class GoLiveError extends TsxComponent<{}> {
       </ErrorLayout>
     );
   }
-  
+
   private renderMachineLockedError(error: IStreamError) {
     return (
       <ErrorLayout error={error}>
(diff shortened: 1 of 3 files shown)


andycreeth:
Hey!  Thanks for your submission.  Can you help me understand how this would happen?  What causes Streamlabs OBS to start streaming when the system is locked?

AuroraAri:
Sure thing! My Stream Deck does not appear to be locking properly when I lock my computer, and I have a "Go Live" action tied to SLOBS. As a result, and only when the confirmation dialog is turned off, SLOBS can go live while the computer is locked. It's definitely an issue on their end, and I've reported it to them, but they're not awesome and open-source! :) I know this is quite the edge case and a comedy of errors before it does happen, but better safe than sorry!

andycreeth:
That makes sense.  Thanks for submitting this.  I'll have our QA team look into testing this next week when we have a little more time.  In the meantime - I see some code formatting issues that will prevent this from building on our CI servers.  Can you push a commit to resolve those?  They're most likely auto-fixable, which you can resolve with `yarn eslint --fix`.

AuroraAri:
Done, and apologies! I don't have as much experience with yarn and CI, so that slipped my mind. Let me know if I can be of further help! :)

holiber:
Going to run tests for this PR here https://github.com/stream-labs/streamlabs-obs/pull/2960

holiber:
Changes have been merged in #2960

PR #2927, opened 2020-09-08
Fix settings window positioning

gettinToasty:


andycreeth (changes requested):
You can pass the `contentStyles` prop to `modal-layout` to remove the padding from the content like this: `:contentStyles="{ padding: '0px' }"`.  Then you can put back the `height: 100%` and remove the negative margin.  I'm not a fan of using this `position: absolute` hack to stretch the content.  It feels lazy and non idiomatic.

andycreeth (approved):

PR #2930, opened 2020-09-09
Revert Scrollbars

gettinToasty:


andycreeth:
Thanks for preparing this.  Luckily I think we can safely close this.

PR #2934, opened 2020-09-09
Add support for prime audio in MediaGallery

gettinToasty:


app/services/media-gallery/media-gallery.ts
```diff
@@ -154,23 +141,32 @@ export class MediaGalleryService extends Service {
     return new Request(url, { ...options, headers });
   }
 
+  private async fetchStockMedia(): Promise<{
+    audios: Array<IMediaGalleryFile>;
+    images: Array<IMediaGalleryFile>;
```
andycreeth, line 146: I don't think there is a functional difference, but why the `Array<>`  syntax as opposed to just `IMediaGalleryFile[]`.  We have previously exclusively used the latter syntax.

app/services/media-gallery/media-gallery.ts
```diff
@@ -6,15 +6,15 @@ import { UserService } from 'services/user';
 import { HostsService } from 'services/hosts';
 import { WindowsService } from 'services/windows';
 import uuid from 'uuid';
-import { stockImages, stockSounds } from './stock-library';
 import { $t } from '../i18n';
 
 export interface IMediaGalleryFile {
   href: string;
-  fileName: string;
+  filename: string;
```
andycreeth, line 13: Is this a breaking change to the API?  We aren't trying to coordinate deploys here are we?  Because that won't work well.

app/services/media-gallery/media-gallery.ts
gettinToasty: i think its when we switched from tslint to eslint but my IDE complains about the latter now

app/services/media-gallery/media-gallery.ts
gettinToasty: we didn't use an api before stock files were hard coded

andycreeth (approved):

app/services/media-gallery/media-gallery.ts
andycreeth: Oh right of course.  Got it.

PR #2998, opened 2020-09-29
Selection service refactor

andycreeth:
The goal of this refactor is to stop mixing in the `Selection` class into the `SelectionService`.  The primary motivation is that working with the `SelectionService` as a `Selection` causes a lot of synchronous IPC due to each call being a service call.  The secondary motivation is that the pattern is a little confusing, the methods get out of sync with `Selection`, and it results in a bad IDE experience when using "Go To Definition".

This PR introduces the concept of the `GlobalSelection`.  This is a class that inherits from `Selection` but has a couple key differences:
- It's `state` represents the current state in the `SelectionService`, and represents what the user has selected in the UI
- When selecting anything on a `GlobalSelection`, it delegates the selection to the `SelectionService`, and that change is reflected in the UI.

You can always get a reference to a `GlobalSelection` from the views on the `SelectionService`.  So for example, to get the number of user selected sources:
Before: `this.selectionService.getSize()`
After: `this.selectionService.views.globalSelection.getSize()`

The main advantage is that the second call can be serviced directly from the Vuex store with no IPC.

This PR also implements the concept of a frozen `Selection`.  You can now call `freeze()` on any `Selection` or `GlobalSelection`, which will prevent the selected items from changing.  This is most useful in the context of the `GlobalSelection` in conjunction with undo/redo.  Most of the undo/redo commands save a `Selection` in memory to remember what items the command operated on.  If a `GlobalSelection` is passed, this is not good for storing in memory because the underlying selection can change when the user selects something.  undo/redo gets around this by simply calling `freeze()` on the selection when storing it long term.


app/services/selection/global-selection.ts
```diff
@@ -0,0 +1,91 @@
...
+        this.editorCommandsService.executeCommand('RemoveNodesCommand', this.clone());
+      });
+  }
+
+  select(items: TNodesList) {
+    if (this.isFrozen) {
+      throw new Error('Attempted to modify frozen selection');
+    }
+
+    // TODO: Allow calling of actions from the worker window
```
holiber, line 75: I didn't know that it's not allowed =(

holiber (approved):
We don't need `Utils.applyMixins(SelectionService, [Selection]);` in the `selection/index.ts` now

app/services/selection/global-selection.ts
andycreeth: Yeah right now if you try to call an action from the worker window you will just get a `null` exception.  I need to add a system so that it just executes synchronously from the worker window.

PR #3004, opened 2020-10-01
Update osn to v0.10.10

EddyGharbi:
EddyGharbi	Update to obs 26


holiber:
Flaky tests should go after merging with the master. But the recording test looks like a real fail.

PR #3005, opened 2020-10-01
remove mouse throttling in the editor

andycreeth:
This removes the blanket 20ms throttle on sending mouse move events to the worker process.  Instead, it does a round trip coordination with the worker process to avoid overwhelming it with events.  This should result in faster processing of events on fast machines, and avoid us still overwhelming the worker process on very slow machines.


holiber:
merging with master should fix the multistream test

holiber (approved):

PR #3038, opened 2020-10-14
Upgrade YT flow

holiber:
- new broadcasts now come with an `Auto-start` option
- added `Stream Latency`, `Privacy Status`, `Category` and `Additional Settings` 
- the broadcast transition code is deleted
- it's possible to stream into one event twice if the `Auto-stop` option has been disabled

image


app/services/platforms/base-platform.ts
```diff
@@ -60,6 +61,21 @@ export abstract class BasePlatformService<T extends IPlatformState> extends Stat
     );
   }
 
+  protected syncSettingsWithLocalStorage() {
```
andycreeth, line 64: Can you explain what is happening here?  Why not just use `PersistentStatefulService`?  Is this so we can only store a subset of settings in localStorage?  Would it be worth adding the functionality to only store certain settings when using `PersistentStatefulService` instead?

andycreeth: Oh I see now that this is a pattern we were already using in the `FacebookService` and you hoisted it up to the base platform service.  I suppose it's fine for now, we should probably come up with a better pattern at some point though.

app/components/windows/go-live/GoLiveChecklist.tsx
```diff
@@ -158,16 +153,25 @@ export default class GoLiveChecklist extends TsxComponent<Props> {
...
-    if (this.view.info.checklist.publishYoutubeBroadcast === 'not-started') return '';
-    const progressInfo = this.youtubeService.progressInfo;
-    return <span class={styles.pending}> {progressInfo.progress * 100}%</span>;
+  private renderWarning() {
+    if (!this.view.info.warning) return;
+    return (
+      <div class="section selectable">
+        <p>
+          {$t(
+            'Auto-start is disabled for you broadcast. You should manually publish you stream in the Youtube Studio',
```
andycreeth, line 167: Grammatical fixes:
`Auto-start is disabled for your broadcast. You should manually publish your stream from Youtube Studio`

andycreeth: Also, just to confirm - we are trying to set Auto-start on the broadcast from the API, so this should rarely/never happen, right?

andycreeth (changes requested):

app/components/windows/go-live/GoLiveChecklist.tsx
holiber: It happens only when you explicitly disable `Auto-start` from SLOBS.

andycreeth: Got it - sounds good.  Thanks!

PR #3046, opened 2020-10-19
implement jfetch helper

andycreeth:
This PR does 2 things primarily:
- Override the built-in `Response` type (which represents the return type of `fetch`) to have a much safer signature for `json()`: `json(): unknown` instead of `json(): any`.  This requires us to pay attention to the return types of JSON fetches, and forces us to cast it to an expected type before using it.
- Add a new utility function `jfetch` that automatically checks if the response was successful and contains JSON, and if so will parse the JSON and return it.  It is a generic function that assists in properly typing the return value when needed.  This should be looked at as a replacement for `handleResponse` in all cases where you are expecting a JSON response.


app/util/requests.ts
```diff
@@ -82,3 +81,20 @@ export function getChecksum(filePath: string) {
     file.on('error', e => reject(e));
   });
 }
+
+export function jfetch<TResponse>(request: RequestInfo, init?: RequestInit): Promise<TResponse> {
```
holiber, line 85: should we cast the default type to `unknown`?
`jfetch<TResponse=unknown>(..`

holiber (approved):

app/util/requests.ts
andycreeth: Yeah I originally had that, but the default type seemed to already be `unknown`.  I'm not sure how TypeScript decides the default type.  It can't hurt to be explicit though.

PR #3057, opened 2020-10-22
Move login to settings menu

gettinToasty:


andycreeth (approved):
Nice.  Let's make sure we communicate this one heavily to support when we roll it out.

PR #3058, opened 2020-10-22
Fetch streamlabel definitions on user login

gettinToasty:


app/services/streamlabels/index.ts
```diff
@@ -233,17 +246,19 @@ export class StreamlabelsService extends Service {
       .then(settings => this.updateSettings(settings));
   }
 
-  async fetchDefinitions(): Promise<IStreamlabelSet> {
+  fetchDefinitions(): void {
```
andycreeth, line 249: Were we already calling this on login?  I don't see this being called anywhere in this PR after you removed from the component.

andycreeth: Nvm see it was already being called elsewhere in this service.

andycreeth (approved):

app/services/streamlabels/index.ts
gettinToasty: yea it was called in init i think suring my initial implementation i was planning to do it there but things changed

PR #3078, opened 2020-10-29
goLiveWindow: Suggest logout when user is trying to disable the primary platform

holiber:
image


andycreeth (approved):
Looks great.  Thanks for setting up a generic component.

PR #3104, opened 2020-11-09
new NVENCE custom resolution

YDeltagon:
Hello, under new NVENC, it is impossible to change the output resolution, compare to NVENC and X264


andycreeth:
Hi, we are looking into this issue but I will close this PR now as it does not seem to be a contribution.

PR #3111, opened 2020-11-16
settings refactor for performance

andycreeth:


app/components/windows/settings/Hotkeys.vue.ts
```diff
@@ -52,18 +52,15 @@ export default class Hotkeys extends Vue {
   async mounted() {
     // We don't want hotkeys registering while trying to bind.
     // We may change our minds on this in the future.
-    this.hotkeysService.unregisterAll();
+    this.hotkeysService.actions.unregisterAll();
```
gettinToasty, line 55: do we want to wait for this to complete before running `getHotkeySet`?

holiber (approved):

gettinToasty (approved):

app/components/windows/settings/Hotkeys.vue.ts
andycreeth: I think it's probably ok if we don't wait.  The main reason we unbind is so that you don't accidentally trigger a hotkey while a user is rebinding one.  It won't affect the result of the `getHotkeySet` call.

PR #3115, opened 2020-11-16
Fix edge case where user hits Done in MediaGallery w/o selecting media

gettinToasty:


app/components/windows/MediaGallery.vue.ts
```diff
@@ -203,6 +203,7 @@ export default class MediaGallery extends Vue {
   }
 
   handleSelect() {
+    if (!this.selectedFile) return this.windowsService.closeChildWindow();
```
holiber, line 206: `this.windowsService.actions.closeChildWindow()`

holiber (approved):

PR #3121, opened 2020-11-18
better list naming

andycreeth:
Also, cleanup some chatbot stuff.

Should be safe to ship to master even though there are changes outside of `app/` since it's only deletions.


gettinToasty (approved):
LGTM, just wondering what the implications of non-app deletions being sent through the bundle update process would be

andycreeth:
They won't be released.  The important thing is just ensuring the compatibility between the webpack bundle and the currently released container.  Given it's only deletions it should be ok, but the bundle QA process will catch any major issues if there's something I overlooked.

PR #3124, opened 2020-11-19
Add analytics for facebook 

holiber:


gettinToasty:
i mentioned this in slack but i think that we should probably extract this into a separate method rather than tacking it onto the beastly goLive method

gettinToasty (approved):

PR #3125, opened 2020-11-19
Youtube thumbnail uploader

holiber:
image


gettinToasty:
more of an implementation question but why not just use our existing media gallery input here?

holiber:
The media-gallery-input has a different scenario of usage. It uploads an image to our server, however for this task I need to upload the image directly to YT. I  wish I could use only styles from the media-gallery-input

app/services/utils.ts
```diff
@@ -221,4 +221,14 @@ export default class Utils {
...
   }
+
+  static getReadableFileSizeString(fileSizeInBytes: number): string {
+    let i = -1;
+    const byteUnits = [' kB', ' MB', ' GB', ' TB', 'PB', 'EB', 'ZB', 'YB'];
+    do {
+      fileSizeInBytes = fileSizeInBytes / 1024;
+      i++;
+    } while (fileSizeInBytes > 1024);
+    return Math.max(fileSizeInBytes, 0.1).toFixed(1) + byteUnits[i];
```
iKBAHT, line 232: What if `i >= byteUnits.length`?

gettinToasty: while this is technically unclamped i think the probability of someone making and successfully uploading a 1025+YB file is extremely improbable, and the only place we use it has a max filesize of 2MB

app/services/platforms/youtube.ts
```diff
@@ -656,3 +694,9 @@ export class YoutubeService extends BasePlatformService<IYoutubeServiceState>
     this.state.categories = categories;
   }
 }
+
+export class YoutubeView extends ViewHandler<IYoutubeServiceState> {
+  getThumbnailUploadUrl(videoId: string) {
+    return;
+  }
+}
```
gettinToasty, line 702: is this unused?

gettinToasty (changes requested):
looks like there is some unused code here but otherwise seems good

gettinToasty (approved):

PR #3134, opened 2020-11-20
Rethink the custom-resolution input

holiber:
The old UI had some issues. There is a new one
resbuttons


andycreeth (approved):
Great fix.  Thank you for doing this.  Resolves the issue and makes the UX 100x better.

PR #3142, opened 2020-11-24
Update brand colors

gettinToasty:
Semi-big diff because of prettier formatting in StudioFooter


andycreeth (approved):
Looks good to me.  I think this is worth having trevor do a quick UI runthrough on all the different themes.  Would probably hold off merging this until Trevor takes a look.

PR #3157, opened 2020-12-04
updated recommended version of node to 14 from 8

loftwah:
I noticed the recommended version of node was still 8 and have updated it to the current active LTS version. Not a huge change but I thought I should fix it.

Cheers


andycreeth (approved):
Thank you

loftwah:
> Thank you

Any time, thanks for the great software!

PR #3162, opened 2020-12-08
Fix for .DS_Store and other invalid files by ignoring them

florin0x01:


app/services/obs-importer.ts
```diff
@@ -128,6 +128,9 @@ export class ObsImporterService extends StatefulService<{ progress: number; tota
 
   private async importCollection(collection: ISceneCollection) {
     const sceneCollectionPath = path.join(this.sceneCollectionsDirectory, collection.filename);
+    if (sceneCollectionPath.indexOf('.json') === -1) {
```
florin0x01, line 131: ignore files such as .DS_Store or other non json files by checking file extension

gettinToasty (approved):

PR #3166, opened 2020-12-09
Fix the resizing junk of the GoLive window

holiber:


app/components/windows/go-live/GoLiveSettings.tsx
```diff
@@ -110,7 +110,7 @@ export default class GoLiveSettings extends TsxComponent<GoLiveProps> {
         )}
 
         {/*RIGHT COLUMN*/}
-        <div style={{ width: '100%' }}>
+        <div style={{ width: '100%', display: 'flex' }}>
```
gettinToasty, line 113: does this need to have a `flex-direction: column`?

holiber: We can have only one visible child element per time for this div, So it doesn't matter which flex-direction we use

gettinToasty (approved):

PR #3171, opened 2020-12-09
Webpack / Toolchain Upgrades

andycreeth:
- Upgrade to webpack 5
- Switch from hard-source-plugin to built-in webpack caching
- Switch `awesome-typescript-loader` to `ts-loader` (atl is not maintained anymore)
- Migrate to `vue-loader` 15
- Upgrade eslint to latest
- Upgrade terser plugin to latest
- Get rid of `tsxconfig.json` and move to a consolidated single `tsconfig.json`
- Remove the strict nulls webpack config and use an env variable instead (this one required a complex merge that was easier to do this way)
- Remove eslint from webpack output to speed up builds.  Instead, use the editor plugin or `yarn eslint`
- We are now targeting ES6 instead of ES5, resulting in smaller bundles due to less polyfilling


webpack.base.config.js
```diff
@@ -77,47 +90,61 @@ module.exports = {
...
+            },
+          },
+        },
       },
       {
         test: /\.tsx$/,
         include: path.resolve(__dirname, 'app/components'),
-        loader: [
+        use: [
           'babel-loader',
```
gettinToasty, line 124: not actually sure we need babel-loader now that our target isn't es5, could boost build times a bit if we are able to get rid of this step

gettinToasty (approved):
one non-blocking question but otherwise didn't notice anything-- kinda skimmed the metacoding parts as i haven't really interacted with that part of the codebase

webpack.base.config.js
andycreeth: Ah got it.  I thought TS compiler was translating from TSX => JSX and then babel was translating JSX => JS.  I can check if that's actually the case.

gettinToasty: oh that might be the case actually, not sure i haven't taken a look in a while

PR #3178, opened 2020-12-11
Handshake version control

florin0x01:


app/app.ts
```diff
@@ -29,6 +29,7 @@ import { UserService, setSentryContext } from 'services/user';
 import { getResource } from 'services';
 import * as obs from '../obs-api';
 import path from 'path';
+import fs from 'fs';
```
andycreeth, line 32: Seems like this import isn't being used.

app/app.ts
```diff
@@ -37,10 +38,11 @@ import CustomLoader from 'components/CustomLoader';
 import process from 'process';
 import { MetricsService } from 'services/metrics';
 import { UsageStatisticsService } from 'services/usage-statistics';
+import { AlertLayoutInput } from 'components/widgets/inputs';
```
andycreeth, line 41: Looks like this isn't being used either

andycreeth:
Closing for now.  Please reopen and address comments if we still want this merged.

PR #3200, opened 2020-12-17
Fix YT test

holiber:
Since we have `enableAutoStart=true` flag for YT streams we close the GoLive window too fast and that can break the test


gettinToasty:
would it not be similar to just not have `enableAutoStart=true` for that test?

holiber:
We don't have the code for stream transition for YT. So we always close the GoLive window pretty fast anyway

andycreeth (approved):

gettinToasty (approved):

PR #3240, opened 2021-01-08
React migration

holiber:


app/components-react/shared/ModalLayout.tsx
```diff
@@ -0,0 +1,77 @@
...
+  // pick variables for the template
+  const { showControls, showCancel, children, showDone, disableDone, onSubmit } = props;
+
+  // render template
+  return (
+    <div className={wrapperClassNames}>
+      {/* CONTENT */}
+      <div className={css.modalLayoutContent}>{children}</div>
+
+      {/* CONTROLS */}
```
gettinToasty, line 60: wondering how custom controls and/or other named slots would work in a react-environment. does it just require us to reframe how we think about slots? perhaps pass them in as props?

app/components-react/shared/ModalLayout.tsx
```diff
@@ -0,0 +1,77 @@
...
+  showDone?: boolean;
+  disableDone?: boolean;
+  onSubmit?: (...args: unknown[]) => unknown;
+}
+type TProps = IModalLayoutProps & { children: ReactNode };
+
+/**
+ * A modal layout for showing dialogs
+ */
+export function ModalLayout(partialProps: TProps) {
```
gettinToasty, line 21: is there a reason to prefer es5 `function` syntax over `const ModalLayout = (props) => ...`?

app/components-react/shared/ModalLayout.tsx
```diff
@@ -0,0 +1,77 @@
...
+  // inject services
+  const { WindowsService, CustomizationService } = Services;
```
gettinToasty, line 23: love how simple this is tbh

app/components-react/shared/ModalLayout.tsx
```diff
@@ -0,0 +1,77 @@
...
+    hasTitleBar: true,
+    showCancel: true,
+    showControls: true,
+    showDone: true,
+    onSubmit: () => {},
+    ...partialProps,
+  };
+
+  // define a reactive state
+  const state = useVuex(() => ({ currentTheme: CustomizationService.currentTheme }));
```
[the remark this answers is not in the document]
gettinToasty: also is this referencing a `view` or direct service state?

app/components-react/windows/NameFolder.tsx
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * Modal for creating or re-naming a folder
+ */
+export default function NameFolder() {
+  // inject services
+  const { ScenesService, EditorCommandsService, WindowsService } = Services;
+
+  // define stateful variables and setters
+  const [name, setName] = useState('');
+  const [error, setError] = useState('');
```
gettinToasty, line 24: is it normal to use `useState` multiple times like this? feels a little like it could be collapsed into an object

app/components/shared/ReactComponent.tsx
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  beforeDestroy() {
+    ReactDOM.unmountComponentAtNode(this.$refs.container);
+  }
+
+  render() {
+    return <div ref="container" style={this.props.wrapperStyles}></div>;
+  }
+}
```
gettinToasty, line 40: this is super clever

app/components/shared/ReactComponent.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+@Component({
+  props: {
+    name: { default: 'NameFolder' },
+    wrapperStyles: { default: () => ({ height: '100%' }) },
+  },
+})
+export class NameFolder extends ReactComponent {}
```
gettinToasty, line 48: is this supposed to be here?

app/components-react/shared/ModalLayout.tsx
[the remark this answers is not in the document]
holiber: We can use `state`, `globalState` or `localState`. But since this component doesn't have a local state I used a simple `state` word.
The pattern I used is from the Redux library. They call the returning object as `selectedData`

app/components-react/shared/ModalLayout.tsx
holiber: In most cases using props for composition is the preferred way.
You may need to use refs when you need a child component to have access to a parent component.

app/components-react/windows/NameFolder.tsx
holiber: You may collapse it into a single object if you need. For me, it's was convenient to have separate setters here. So I can use more readable `setError(myError)` instead `setState({ error: myError, ...state })`

app/components/shared/ReactComponent.tsx
holiber: Yes. I think we can have a list of all wrapped react components in this file cuz you don't need to write any implementation code for them.

app/components-react/shared/ModalLayout.tsx
holiber: I could ask the opposite question. Is there a reason to prefer arrow functions instead of regular ones?

Named functions are always a preferred way to define a function if you don't need to re-bind `this`. Mostly because of 2 reasons.
1) You always see a function name in the debugger call stack instead of anononimous ones
2) Compiler sends the definition of the named function to the top of the file. So if you define `foo()` after `bar()` you can still call `foo()` inside `bar()`

andycreeth: I could be wrong, but I think these days chrome treats an anonymous function assigned to a `const` the same as a named function (with the variable name being the name of the function).  Either way, I agree with you and don't have strong preference here.

app/components-react/shared/ModalLayout.tsx
andycreeth: This is the way.

app/components-react/windows/NameFolder.tsx
andycreeth: Yeah in my very limited knowledge of hooks, each call to `setState` gives you a single state variable.  You can choose to make that variable a complex object containing all your state, or you can use lots of separate variable.  The react docs don't seem to indicate one approach is better than others, so seems up to the dev and what is more clear.  I'm sure we will end up heading in one direction or another as we get used to the patterns.

andycreeth (commented):
Seems like a solid foundation to me.  What should we do before merging this?  Do we need a call to sync up on how we feel and whether this is the right path fwd?

holiber:
I need to fix tests and we can merge. I think we should have a sync call about the migration roadmap

gettinToasty (approved):

PR #3242, opened 2021-01-08
Fix mac pathing for visual presets

gettinToasty:


app/services/source-filters.ts
```diff
@@ -66,18 +67,19 @@ export interface ISourceFilterIdentifier {
...
+      { title: $t('Grayscale'), value: `${prefix}grayscale.png` },
+      { title: $t('Sepiatone'), value: `${prefix}sepia.png` },
+      { title: $t('Dramatic'), value: `${prefix}gazing.png` },
+      { title: $t('Flashback'), value: `${prefix}muted.png` },
+      { title: $t('Inverted'), value: `${prefix}inverted.png` },
+      { title: $t('Action Movie'), value: `${prefix}cool_tone.png` },
+      { title: $t('Hearth'), value: `${prefix}warm_tone.png` },
+      { title: $t('Wintergreen'), value: `${prefix}green_tone.png` },
+      { title: $t('Heat Map'), value: `${prefix}heat_map.png` },
+      { title: $t('Cel Shade'), value: `${prefix}cel_shade.png` },
```
andycreeth, line 82: This works, but `path.join` https://nodejs.org/api/path.html#path_path_join_paths is the proper cross-platform way to handle this.

holiber (approved):
Path.resolve() would be a more correct way

andycreeth (approved):

PR #3253, opened 2021-01-13
Remove frontend references to facemasks

gettinToasty:


andycreeth:
We will need to make sure we have tested this from the perspective of someone who has facemasks enabled, and the perspective of someone who has the plugin in their scene collection.  We need to make sure the scene collection doesn't fail to load or display bad errors.

gettinToasty:
@andycreeth i think i had the plugin in my collection and but not sure if it was enabled since we hide the settings now 🤔 not sure how we can test this since it's no longer enable-able

andycreeth:
Will need to hack it enabled on your account.  Can do this by calling some dead code from dev tools probably.  Needs to be tested.

andycreeth (approved):

PR #3254, opened 2021-01-13
Electron 10 upgrade

andycreeth:
Upgrades us to electron 10.  The main reason for the large diff is upgrading spectron brought with it 2 major version upgrades of webdriverio, which contains an API overhaul with a large number of breaking changes.


holiber (approved):
Wow. Great job

PR #3258, opened 2021-01-14
Handle mixer login

holiber:


andycreeth (approved):
When the user is in this state, they will appear logged out, right?  Since `loginValidated` is false?

holiber:
The user should be logged out in the same state like if they launched SLOBS without a network

PR #3262, opened 2021-01-19
fix endless loading on overlay/widget import

trevorrl:
the button to import an overlay or widget file would load endlessly (until you closed Settings) if you hit "cancel" on the file dialog. this fixes that.


holiber (approved):

holiber:
Good catch

PR #3266, opened 2021-01-21
Use idomatic JS to resolve preset filter paths

gettinToasty:
Remembered I had this chore while looking at some file saving things lol


andycreeth (approved):
Thanks for going back and fixing this!

PR #3272, opened 2021-01-27
Fix crash in NameFolder.tsx

holiber:
Going to handle this better later


app/components-react/windows/NameFolder.tsx
```diff
@@ -36,7 +36,7 @@ export default function NameFolder() {
   });
 
   // define a submit method
-  function submit(e: Event) {
+  function submit(e: any) {
```
andycreeth, line 39: Surely there's a better type we can use here

andycreeth (approved):

PR #3275, opened 2021-01-28
add ability to change name/icon of custom layouts

trevorrl:
initial attempt, this will def need review as it's probably the most complicated thing I've done in the app so far lol


gettinToasty:
what's the status of this? afraid it could be getting pretty stale

trevorrl:
honestly very stale, if I have bandwidth this week I'll try and get it across the finish line otherwise I think we can abandon it. doesn't really add a whole lot of value IMO and was more of a learning experience for me I think. will close EoW if I haven't worked on it by then.

PR #3288, opened 2021-02-02
Update the electron dialog API when editing a Slideshow

gettinToasty:


andycreeth (approved):
Ouch this has probably been broken for a while.  Think this API changed in electron 6.

PR #3291, opened 2021-02-03
fix object could not be cloned error

andycreeth:


gettinToasty (approved):
is there a reason `0` is used here specifically as opposed to like a `null` or `undefined`?

andycreeth:
@gettinToasty No strong reason.  Anything that is serializable will work.  `0` is a single character and happens to be what people were using in the Github issue I was looking at.  😄

PR #3308, opened 2021-02-12
Port NewsBanner to React

gettinToasty:


app/components-react/root/NewsBanner.tsx
```diff
@@ -0,0 +1,83 @@
+import { shell } from 'electron';
+import { $t } from '../../services/i18n';
+import React, { useState } from 'react';
+import { Services } from '../service-provider';
+import cx from 'classnames';
+import styles from './NewsBanner.m.less';
+import { TAppPage } from '../../services/navigation';
+import { useVuex } from '../hooks';
+
+export default () => {
```
holiber, line 10: Use named functions `export default function MyComponent` for components. With the anonymous function, you will not see the component name in the call stack trace.

app/components-react/root/NewsBanner.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  // TODO: use hook after Alex's branch is merged
+  // const toggleAnimation = () => {
+  //   WindowsService.actions.updateStyleBlockers('main', true);
+  //   window.setTimeout(() => {
+  //     WindowsService.actions.updateStyleBlockers('main', false);
+  //   }, 500);
+  // };
+
+  const closeBanner = async (
```
holiber, line 28: Use simple named functions for the component methods. We don't need arrow functions here because we don't use `this` anymore. Also, named functions allow you to define them in any order

holiber (approved):

PR #3340, opened 2021-02-24
React shared components

holiber:
Is safe to merge now


app/components-react/shared/inputs/ListInput.tsx
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  function onSearchHandler(searchStr: string) {
+    p.onSearch && p.onSearch(searchStr);
+  }
+
+  return render();
+});
+
+export function renderOption(opt: IListOption, ind: number, inputProps: ICustomListProps) {
```
gettinToasty, line 79: is this an accidental export?

app/components/windows/go-live/GoLiveError.tsx
```diff
@@ -83,8 +83,6 @@ export default class GoLiveError extends TsxComponent<{}> {
         return this.renderRestreamError(error);
       case 'YOUTUBE_STREAMING_DISABLED':
         return this.renderYoutubeStreamingDisabled(error);
-      case 'YOUTUBE_PUBLISH_FAILED':
-        return this.renderYoutubePublishError(error);
```
gettinToasty, line 87: not sure why this is removed?

app/components-react/shared/inputs/ListInput.tsx
gettinToasty: disregard, i see where it's used

app/components-react/shared/inputs/ListInput.tsx
```diff
@@ -0,0 +1,149 @@
...
+// select what features from the antd lib we are going to use
+const ANT_SELECT_FEATURES = [
+  'showSearch',
+  'loading',
+  'placeholder',
+  'notFoundContent',
+  'onDropdownVisibleChange',
+  'onSearch',
+  'onSelect',
+] as const;
```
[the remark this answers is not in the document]
holiber:
```
const a = ['foo', 'bar']; // has type string[]
const b = ['foo', 'bar'] as const; // has type ['foo', 'bar']
```

app/components/windows/go-live/GoLiveError.tsx
holiber: This accidentaly came from my original big PR. YT should automatically publish the broadcast now, this error is not happening anymore

gettinToasty (approved):

PR #3351, opened 2021-03-03
Typescript 4.0.7

holiber:


app/components/GLVolmeters.tsx
```diff
@@ -108,6 +108,7 @@ export default class GLVolmeters extends TsxComponent<VolmetersProps> {
   private sourcesOrder: string[];
   private workerId: number;
   private requestedFrameId: number;
+  // @ts-ignore
```
holiber, line 111: Typescript thinks that `this.customizationService` is not defined here. We should use a new dependency-injector pattern to avoid this kind of error. This PR contains only changes for types to avoid the introduction of new bugs

app/components/widgets/SponsorBanner.vue.ts
```diff
@@ -36,6 +36,7 @@ export default class SponsorBanner extends WidgetSettings<
     return this.wData.settings.placement_options === 'double' ? ['1', '2'] : ['1'];
   }
 
+  // @ts-ignore
```
holiber, line 39: Typescript doesn't like overriding property with a getter.

app/components/GLVolmeters.tsx
```diff
@@ -108,7 +108,7 @@ export default class GLVolmeters extends TsxComponent<VolmetersProps> {
   private sourcesOrder: string[];
   private workerId: number;
   private requestedFrameId: number;
-  private bgMultiplier = this.customizationService.isDarkTheme ? 0.2 : 0.5;
+  private bgMultiplier = this!.customizationService.isDarkTheme ? 0.2 : 0.5;
```
gettinToasty, line 111: kind of wondering why we need to assert `this`, but only sometimes?

app/components/tsx-component.ts
```diff
@@ -42,7 +42,7 @@ export default abstract class TsxComponent<
   }> &
     Readonly<TProps>;
 
-  value: TProps['value'];
+  value: any;
```
gettinToasty, line 45: should this be unknown?

app/services/selection/global-selection.ts
```diff
@@ -26,10 +26,12 @@ export class GlobalSelection extends Selection {
   frozenState: ISelectionState;
   frozenSceneId: string;
 
+  // @ts-ignore
```
gettinToasty, line 29: why are we ignoring this one?

app/services/widgets/settings/event-list.ts
```diff
@@ -134,11 +134,7 @@ export class EventListService extends WidgetSettingsService<IEventListData> {
           ),
         },
       ],
-      mixer: [
-        { key: 'sparks_minimum', title: $t('Min. Sparks') },
-        { key: 'embers_minimum', title: $t('Min. Embers') },
-      ],
-    }[platform];
+    }[platform as 'twitch'];
```
gettinToasty, line 137: why are we even making this keyed object if there's only 1 key lol

app/components/tsx-component.ts
holiber: Unfortunately, I have an error in some components if I use `unknown` here

app/services/widgets/settings/event-list.ts
holiber: Because this object accepts reading only for the  `twitch` key, meanwhile `TPlatfom` type can be any of platforms.

app/services/selection/global-selection.ts
holiber: Fixed

app/components/GLVolmeters.tsx
holiber: Typescript thinks that `this!.customizationService` is not initialized here.

app/services/widgets/settings/event-list.ts
gettinToasty: yea i guess i'm asking why this is a keyed object in the first place if it only accepts one key

holiber: We had Mixer in that list before. I strived to not touch logic in this PR if I can to avoid occasional issues. So I just added a typecast here

app/components/tsx-component.ts
andycreeth: This should be removed.  `TsxComponent` should not be declaring a `value` attribute.

app/components/TransitionSettings.vue.ts
```diff
@@ -75,7 +75,7 @@ export default class SceneTransitions extends Vue {
     return this.transitionsService.getTransition(this.transitionId);
   }
 
-  properties = this.transitionsService.getPropertiesFormData(this.transitionId);
+  properties = this.transitionsService.getPropertiesFormData(this!.transitionId);
```
andycreeth, line 78: Move the `!` to the `transitionId` declaration

andycreeth (approved):

PR #3358, opened 2021-03-05
fix svg titlebar icon for prime themes

trevorrl:
before:
image

after:
image


gettinToasty (approved):

andycreeth:
Thanks for fixing!  Think there's still something else going on though.  Icon is supposed to be further from the left edge (see night/day theme)  Would be good to get it consistent across all themes.

trevorrl:
will take a look at that today

PR #3359, opened 2021-03-05
Add support for FB Gifted Supporters

gettinToasty:


app/services/widgets/settings/alert-box/alert-box-data.ts
```diff
@@ -23,22 +23,29 @@ export const API_NAME_MAP = {
...
   facebook_like: 'likes',
   facebook_stars: 'stars',
   facebook_share: 'shares',
   facebook_follow: 'fbfollows',
   loyalty_store_redemption: 'loyaltystore',
-  effect: 'effects',
-  sticker: 'stickers',
+};
+
+const determineTester = (key: string) => {
```
holiber, line 35: Can you add a comment about this method? I'm still not 100% sure why we have different testers here

holiber (approved):

app/services/widgets/settings/alert-box/alert-box-data.ts
gettinToasty: can do

PR #3363, opened 2021-03-08
implementing Keakr as a new top level platform within streamlabs-obs

soualid:
Hello,

You'll find below a PR that implements Keakr as a new top-level platform within Streamlabs OBS. 

image

Keakr is a social network dedicated to musicians available on iOS and Android as well as on keakr.com website, it includes a free livestream feature that is accessible through a few clicks within Streamlabs OBS thanks to this PR.

Actually, we disabled widgets because Keakr is not integrated as an authentication provider platform in streamlabs.com, so we cannot access a working token to show up the widget editor. For the same reason, all the streamlabs.com platform endpoint (`slobs/*`) ends with a 401 error code. 

However, we also developed a third-party application using Streamlabs API that makes the alerts and donations widgets fully operational once setup: ie. when an alert widget is setup in Streamlabs OBS, if another Keakr user donates to the streamer, the alert is properly triggered through the [trigger endpoint](https://dev.streamlabs.com/docs/triggering-alerts) and show up within the streamer's stream.

Please let us know if you would like to merge this new-top level platform inside Streamlabs OBS basecode or if we should distribute this integration as a dedicated fork to our end users. Obviously, if we do, we would disable all the `slobs` calls triggered by the primary authentication flow, in the other case, we'll need a way to authenticate keakr users against streamlabs API.

Thank you in advance for your feedbacks on this!

Simon for the Keakr app


andycreeth:
Hey, thanks for your submission. Adding a new platform integration is something we decide internally based on demand from the community and our strategy. If you would like to discuss an integration further can you please email bd (at) streamlabs.com? For now I am going to close this PR, but don't hesitate to reach out to discuss further.

PR #3365, opened 2021-03-08
Bring back Vue Chat

gettinToasty:
Don't want to hold up release by investigating the larger issues at play


app/components-react/root/Chat.tsx
```diff
@@ -58,6 +54,7 @@ export default function Chat(props: { restream: boolean }) {
   }
 
   function changeChat() {
+    console.log(props.restream);
```
andycreeth, line 57: Remove?

andycreeth (approved):

PR #3371, opened 2021-03-09
fix obs init failure flow

andycreeth:


app/app.ts
```diff
@@ -301,13 +301,13 @@ document.addEventListener('DOMContentLoaded', async () => {
       electron.remote.process.env.SLOBS_VERSION,
     );
 
-    if (apiResult !== obs.EVideoCodes.Success) {
+    if (apiResult !== obs.EVideoCodes.Success || true) {
```
gettinToasty, line 304: guessing you want to remove that `|| true` LOL

PR #3398, opened 2021-03-23
Port Loader to React

gettinToasty:
Also removes beaker dependency from it


app/components-react/pages/Loader.tsx
```diff
@@ -0,0 +1,132 @@
...
+      </div>
+    </div>
+  );
+}
+
+function Spinner() {
+  return (
+    <div className="s-spinner s-spinner__overlay">
+      <div className="s-bars">
+        <Svg src={spinnerSrc} className="s-spinner--large" />
```
holiber, line 59: Maybe change it to something like SvgContainer? `Svg` is already a reserved tag and IDE will not automatically suggest you importing it

holiber (approved):

PR #3400, opened 2021-03-24
filter localstorage on load for persistent stateful service

andycreeth:


gettinToasty (approved):

app/services/core/persistent-stateful-service.ts
```diff
@@ -22,7 +22,7 @@ export abstract class PersistentStatefulService<TState extends object> extends S
   static get initialState() {
     const persisted = JSON.parse(localStorage.getItem(this.localStorageKey) as string) || {};
 
-    return merge({}, this.defaultState, persisted);
+    return this.filter(merge({}, this.defaultState, persisted));
```
gettinToasty, line 25: does `this` need to be changed to `(this.constructor as typeof PersistentStatefulService)` here as well

gettinToasty (changes requested):

app/services/core/persistent-stateful-service.ts
andycreeth: Nope because `initialState` is static, so in this getter `this` already refers to the constructor.

gettinToasty: oh nice didn't catch that

gettinToasty (approved):

PR #3401, opened 2021-03-24
Fix watcher eagerly refreshing components

gettinToasty:


app/components/shared/ReactComponent.tsx
```diff
@@ -1,4 +1,5 @@
 import TsxComponent, { createProps } from '../tsx-component';
+import isEqual from 'lodash/isEqual';
```
holiber, line 2: You don't need this import

holiber (approved):

app/components/shared/ReactComponent.tsx
gettinToasty: wait im still using `isEqual` tho lol

PR #3403, opened 2021-03-25
fix empty checkboxes in Jar widget

trevorrl:
Issue: 
image

Two fixes: added support for Facebook Support Gifters value, and for merged accounts API was returning a `priority` value that we weren't filtering out (looked on the site and don't see it anywhere there either) so I just filtered it like we do `_id`. 

Marking as draft for now, going to double check with Eric to make sure that `priority` value isn't important.


trevorrl:
oh also, should probably merge Sean's PR first, I had to merge his into mine in order to test since widget settings window was broken on master lol.

gettinToasty (approved):

PR #3408, opened 2021-03-26
Port NavTools to React

gettinToasty:


trevorrl:
decided to get an early start on this, looks like it gets rid of the space between top and bottom nav icons
image

holiber (approved):

PR #3416, opened 2021-04-01
React GoLive window

holiber:


package.json
```diff
@@ -80,8 +80,11 @@
     "obs-studio-node": "https://s3-us-west-2.amazonaws.com/obsstudionodes3.streamlabs.com/osn-0.0.0-release.tar.gz",
     "overlayscrollbars-react": "^0.2.2",
     "raven": "^2.6.4",
+    "react-devtools-electron": "^4.7.0",
+    "react-transition-group": "^4.4.1",
     "request": "^2.88.0",
     "rimraf": "^2.6.1",
+    "save-dev": "^0.0.1-security",
```
andycreeth, line 87: Think this was probably a mistake.  Remove?

app/services/streaming/streaming-view.ts
```diff
@@ -97,39 +151,50 @@ export class StreamInfoView extends ViewHandler<IStreamingServiceState> {
     return getPlatformService(this.userView.auth.primaryPlatform).chatUrl;
   }
 
+  getTweetText(streamTitle: string) {
+    return `${streamTitle} ${Services.TwitterService.url}`;
```
andycreeth, line 155: One of the "rules" of views is that they shouldn't inject services.  The proper way to access other services from within a view handler is `this.getServiceViews()` which restricts your access to the views on another service, ensuring you don't accidentally do any IPC.

andycreeth (approved):

PR #3418, opened 2021-04-02
Change FB chat url and make description required

holiber:


app/services/platforms/facebook.ts
```diff
@@ -481,7 +496,7 @@ export class FacebookService extends BasePlatformService<IFacebookServiceState>
 
   get chatUrl(): string {
     if (this.state.settings.destinationType === 'page' && this.state.settings.game) {
-      return 'https://www.facebook.com/gaming/streamer/chat/';
+      return `https://www.facebook.com/live/producer/dashboard/${this.state.videoId}/COMMENTS/`;
```
andycreeth, line 499: I think we should be able to use this chat for more than just gaming creator pages now.  Can you please test the other cases and enable it for cases that it works?  This chat will be better than our read-only chat.

holiber: It works only for gaming pages unfortunately

gettinToasty (approved):

PR #3435, opened 2021-04-14
Add new entitlement

florin0x01:


andycreeth (changes requested):
I don't think we should merge this, especially without evidence that this will fix anything.  But even if this were the fix, I think this would warrant more discussion.

PR #3447, opened 2021-04-21
Catch type safety

andycreeth:
- Upgrades prettier for TypeScript 4 support
- Adds tslint rule `@typescript-eslint/no-implicit-any-catch` which forces us to set `unknown` type on all catch clauses, requiring us to do runtime checks before accessing the object in a way that requires it to be a particular type
- This PR strives to change no functionality, therefore fixes for existing code was done in a **Bad Way**.  All new code should not use `[]` for unsafe access like I did here, but should instead do runtime type checking and operate accordingly.

Note: This PR contains changes to updater code and yet is slated for bundle update - that's fine and expected.  The updater changes won't affect the compiled code so there's no risk here.  Just merging it to master to spread it through the codebase as fast as possible.


holiber (approved):
Looking good 👍

PR #3452, opened 2021-04-22
3 pixel margin to fix jank on logout/login button

trevorrl:
before: 
image

after:
image


gettinToasty:
i think a more consistent fix is changing line 177 to `align-items: center`

gettinToasty (approved):

PR #3459, opened 2021-04-23
fb chat fixes and go live window fixes

trevorrl:
- added the sleeping kevin offline image to FB chat like we have for YT
- removed GVC restriction from FB chat
- added ability to clear the game selector
- changed default tweet state to true

### DIFF
--- a/app/components-react/windows/go-live/GameSelector.tsx
+++ b/app/components-react/windows/go-live/GameSelector.tsx
@@ -74,6 +74,7 @@ export default function GameSelector(p: TProps) {
       imageSize={TwitchService.gameImageSize}
       loading={isSearching}
       notFoundContent={isSearching ? $t('Searching...') : $t('No matching game(s) found.')}
+      allowClear
     />
   );
 }
--- a/app/components/LiveDock.vue
+++ b/app/components/LiveDock.vue
@@ -77,7 +77,7 @@
         </div>
         <div
           class="live-dock-chat"
-          v-if="!hideStyleBlockers && (isTwitch || (isYoutube && isStreaming) || isFacebook)"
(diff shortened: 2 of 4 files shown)


app/components-react/windows/go-live/GameSelector.tsx
```diff
@@ -74,6 +74,7 @@ export default function GameSelector(p: TProps) {
       imageSize={TwitchService.gameImageSize}
       loading={isSearching}
       notFoundContent={isSearching ? $t('Searching...') : $t('No matching game(s) found.')}
+      allowClear
```
holiber, line 77: I we want to keep it the same for Twitch we should use `allowClear={!isTwitch}`

app/components/LiveDock.vue
```diff
@@ -77,7 +77,7 @@
         </div>
         <div
           class="live-dock-chat"
-          v-if="!hideStyleBlockers && (isTwitch || (isYoutube && isStreaming) || isFacebook)"
+          v-if="!hideStyleBlockers && (isTwitch || (isYoutube && isStreaming) || isFacebook && isStreaming)"
```
holiber, line 80: I suggest adding additional bracers for easier reading `(isTwitch || (isYoutube && isStreaming) || (isFacebook && isStreaming))`

holiber (approved):

app/components-react/windows/go-live/GameSelector.tsx
trevorrl: do you think we should? the field is required for Twitch and we have a warning if it's cleared saying the field is required. IMO this is a better experience (I personally prefer to clear the field before searching)

PR #3464, opened 2021-04-26
Make TwitchGameSelector required

holiber:


app/components-react/windows/go-live/useGoLiveSettings.ts
```diff
@@ -109,7 +109,7 @@ function initializeGoLiveSettings(
     renderPlatformSettings(
       commonFields: JSX.Element,
       requiredFields: JSX.Element,
-      optionalFields: JSX.Element,
+      optionalFields?: JSX.Element,
```
andycreeth, line 112: Don't understand how this change is related

andycreeth: Nvm I get it now

PR #3468, opened 2021-04-26
Integration

holiber:


app/services/platforms/twitch.ts
```diff
@@ -106,7 +106,7 @@ export class TwitchService
   };
 
   // Streamlabs Production Twitch OAuth Client ID
-  clientId = '8bmp6j83z5w4mepq0dn0q1a7g186azi';
+  clientId = '3eoucd9qwxqh7pu3l0e3rttomgrov2';
```
trevorrl, line 109: assuming this is the beta client ID so will want to change this back as well

app/components-react/shared/PlatformLogo.tsx
```diff
@@ -3,22 +3,65 @@ import { TPlatform } from '../../services/platforms';
...
+    alignItems: 'center',
+    justifyContent: 'center',
+  };
+
+  return (
+    <div style={wrapperStyles}>
+      <i
+        className="fab tiktok"
+        style={{ width: `${size / 2}px` }}
+        dangerouslySetInnerHTML={{ __html: getTikTokSvg(color) }}
```
gettinToasty, line 63: why are we using `i` with `dangerouslySetInnerHTML` here? you can just use an SVG like any other React or Vue component, see `KevinSVG.tsx` for an example

holiber: We don't ask to React to parse SVG nodes and track dom changes if we provide SVG as simple text. There are no benefits in this particular example but you might find performance issues for a bigger SVG markup,

holiber: Solution B always forks faster than solution A, and solution A has no advantages against solution B. I don't have any strong reasons to use solution A in this case.

holiber: >  but if there is no benefit...
There is a benefit but it's not noticeable. You will not feel the difference between `1ms` rendering and `10ms` rendering for one SVG. But technically it's `x10` difference. And when you're rendering SVG with React it really works as much slower than static rendering.

app/services/platforms/tiktok.ts
```diff
@@ -0,0 +1,83 @@
...
+  readonly displayName = 'TikTok';
+  readonly capabilities = new Set<TPlatformCapability>([]);
+
+  // support only donation widgets for now
+  readonly widgetsWhitelist = [
+    WidgetType.AlertBox,
+    WidgetType.DonationGoal,
+    WidgetType.DonationTicker,
+    WidgetType.TipJar,
+  ];
```
gettinToasty, line 30: you can use the `platforms` field of `WidgetDisplayData` in `widgets-data` to accomplish this :)

app/components/windows/go-live/CommonPlatformFields.tsx
```diff
@@ -176,7 +176,7 @@ export default class CommonPlatformFields extends TsxComponent<Props> {
             <div>
               {/*TITLE*/}
               <HFormGroup
-                value={fields.title}
+                value={fields['title']}
```
gettinToasty, line 179: not blocking but wondering why this change was made

app/services/platforms/tiktok.ts
```diff
@@ -0,0 +1,83 @@
...
+  /**
+   * prepopulate channel info and save it to the store
+   */
+  async prepopulateInfo(): Promise<void> {
+    this.SET_PREPOPULATED(true);
+  }
+
+  async fetchViewerCount(): Promise<number> {
+    // no API
+    return 0;
```
gettinToasty, line 68: wonder if we just want to hide view count for tik tok here if there isn't an api for it

app/services/platforms/tiktok.ts
holiber: I found the `platforms` field for events there but don't see any example for widgets

app/components/windows/go-live/CommonPlatformFields.tsx
holiber: There is a TS type issue that I'm not sure how to resolve with massive changes in type description. `fields.title` used to be a field for all platforms but it's not true now. However, we never show `CommonPlatformFields.tsx` for TikTok so should be no real errors here

app/services/platforms/tiktok.ts
holiber: We still show a `ViewerCount` for other platforms in multiplatform mode. But I agree it's forth to handle it for TikTok separately. I created an Asana tiket

app/services/platforms/tiktok.ts
gettinToasty: it currently only exists for platform-specific widgets, but can be extended to include all platforms except tiktok where needed
```
  [WidgetType.SupporterGoal]: {
    name: $t('Supporter Goal'),
    description: $t('Set a goal for your viewers to help you reach.'),
    demoVideo: false,
    demoFilename: 'source-follower-goal.png',
    supportList: [$t('Facebook Supporters')],
    platforms: new Set(['facebook']),
  },
```

andycreeth (approved):

PR #3473, opened 2021-04-27
make tags optional

andycreeth:


holiber (approved):
Yeah, That's the correct way. I mixed up the order of things in my original PR

PR #3493, opened 2021-04-29
React bugfixes and improvements

holiber:
## fixes
- fix an issue where an input component doesn't update if it has an array value (affected TagsInput)
- fix reactivity for the `Form.layout` property
- Fix the not-rendering icon in `Message.tsx`

## Improvements that minimize the chance of having staled state
 - don't re-create inputAttrs and wrapperAttrs for each input update
 - don't re-create the state object from useFormState for each component update

## other improvements
- add tooltips for all inputs, not only for checkboxes
- TagsInput shows a message when max selected tags reached
- create a `useBinding` hook for reactive input bindings in StateManager
- Allow SwitchInput to have Enabled/Disabled text
- Allow TextInput to have side buttons
- Remove redundant state dependency in TwitchEditSteamInfo.tx


app/components-react/shared/inputs/TagsInput.tsx
```diff
@@ -28,9 +31,17 @@ export type TTagsInputProps<TValue> = TSlobsInputProps<
...
-  const { inputAttrs, wrapperAttrs, forceUpdate } = useInput('tags', p);
+  const { inputAttrs, wrapperAttrs } = useInput('tags', p);
   const options = p.options;
   const tagsMap = useMemo(() => keyBy(options, 'value'), [options]);
+  const { s, updateState } = useFormState({ isAnimating: false });
+
+  async function animateWarning() {
+    if (s.isAnimating) return;
+    updateState({ isAnimating: true });
+    await Utils.sleep(1000);
```
gettinToasty, line 42: this animation takes a full second? that seems like way too long for UX

holiber: It takes less than a seccond. It just doesn't allow to start animation again within 1s

andycreeth (approved):

PR #3506, opened 2021-05-05
Show integration if primary platform

holiber:


app/components/windows/settings/StreamSettings.tsx
```diff
@@ -52,7 +52,10 @@ export default class StreamSettings extends TsxComponent {
   private editCustomDestMode: boolean | number = false;
 
   // TODO: fix TikTok merging on the Back End
-  private platforms = this.streamingView.allPlatforms.filter(platform => platform !== 'tiktok');
+  private platforms = this.streamingView.allPlatforms.filter(
+    // show tiktok only if it's a primary platform
+    (platform, ind) => ind === 0 || platform !== 'tiktok',
```
andycreeth, line 57: Seems kind of brittle to do this via index, would make more sense to compare it to `this.userService.views.platform`

holiber: That makes sense for a permanent solution

andycreeth (approved):

PR #3507, opened 2021-05-05
Scale Settings Logo

gettinToasty:


holiber (approved):

app/components-react/shared/PlatformLogo.tsx
```diff
@@ -44,7 +44,7 @@ function getTikTokSvg(color: string) {
 
 function TikTokLogo(p: IProps) {
   const color = 'black';
-  const size = 40;
+  const size = (p.size && (sizeMap[p.size] || p.size)) || 40;
```
andycreeth, line 47: When working with numbers doing null/undefined fallbacks it's better to use `??` instead of `||`.

andycreeth (approved):

app/components-react/shared/PlatformLogo.tsx
gettinToasty: was just copying over the existing code, but can make that change. in this case though i don't think size is ever expected to be 0

andycreeth: Sure makes sense.  Just mentioned because general good practice.

PR #3524, opened 2021-05-12
skip setting output res for tiktok users

trevorrl:
skip setting the output res with encoder optimization for tiktok users, since they stream in portrait

### DIFF
--- a/app/services/video-encoding-optimizations/video-encoding-optimizations.ts
+++ b/app/services/video-encoding-optimizations/video-encoding-optimizations.ts
@@ -165,14 +165,9 @@ export class VideoEncodingOptimizationService extends PersistentStatefulService<
       bitrate: currentSettings.streaming.bitrate,
     };
 
-    // change the resolution only if user didn't set a custom one
-    if (!currentSettings.streaming.hasCustomResolution) {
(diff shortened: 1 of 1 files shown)


app/services/video-encoding-optimizations/video-encoding-optimizations.ts
```diff
@@ -163,9 +165,14 @@ export class VideoEncodingOptimizationService extends PersistentStatefulService<
       bitrate: currentSettings.streaming.bitrate,
     };
 
+    // change the resolution only if user didn't set a custom one
     if (!currentSettings.streaming.hasCustomResolution) {
-      // change the resolution only if user didn't set a custom one
-      newStreamingSettings.outputResolution = encoderProfile.resolutionOut;
+      // skip this output res step for tiktok since they stream in portrait
+      if (this.userService.platformType === 'tiktok') {
```
gettinToasty, line 171: i feel like you can just add this condition to the first if statement

trevorrl: u rite, pushed a change

gettinToasty (approved):

PR #3539, opened 2021-05-24
fix showing variation conditions on default alerts

trevorrl:
default alerts are not just "default", they are "default-donations" or "default-bits", etc, so we were showing variation conditions on default alerts when we shouldn't be


holiber (approved):
LGTM but let @gettinToasty review too

PR #3551, opened 2021-06-01
Grow Tab

gettinToasty:


app/services/grow/grow-data.ts
```diff
@@ -0,0 +1,141 @@
+export const GROWTH_TIPS = () => [
+  {
+    title: 'Set Up Tips',
+    icon: 'icon-balance',
+    link: 'http://www.streamlabs.com',
+    cta: 'Set Up Tips',
+    description:
+      "Go to Donation Settings and enter your personal information to get started. If you're a new streamer, we recommend starting with PayPal.",
```
andycreeth, line 8: Not sure how often we will be changing these, but just bringing up in case we want to add these to i18n as well.

andycreeth, line 8: Not sure how often we will be changing these, but bringing up in case we want to add these to i18n as well.

andycreeth (approved):

app/services/grow/grow-data.ts
gettinToasty: yea i left them untranslated assuming the next iteration would be getting them from API, in which case we wouldn't be able to translate them locally via i18n regardless

PR #3555, opened 2021-06-02
Hide mixer if audio source isn't controlled by OBS

gettinToasty:


andycreeth (commented):
So I have a couple concerns with this solution:
- `getPropertiesFormData` is a fairly expensive call to be doing on re-renders
- This will be non-reactive when you change the setting in properties

Although it's more work, I wonder if we would be better served with a solution that keeps a flag in the vuex store and is updated whenever that property is changed.

holiber (approved):

PR #3556, opened 2021-06-02
Fix NameFolder Submit

gettinToasty:
It appears that onChange whilst focusing on an input swallows the initial MouseEvent that should be hitting the submit button


holiber:
I think there should be another way to fix that. BTW feel free to merge this solution for now

holiber (approved):

PR #3558, opened 2021-06-03
Fix twitter input

holiber:


trevorrl:
can confirm that this fixes the issue

gettinToasty (approved):

PR #3561, opened 2021-06-08
Add rough first draft of Highlighter

andycreeth:
Adds a rough first draft of the highlighter feature behind a feature flag.


app/components-react/shared/inputs/BoolButtonInput.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import cx from 'classnames';
+import { InputComponent, TSlobsInputProps, useInput } from './inputs';
+import styles from 'components/shared/inputs/BoolButtonInput.m.less';
+
+interface IBoolButtonInputCustomProps {
+  checkboxStyles: { [key: string]: any };
+  checkboxActiveStyles: { [key: string]: any };
```
holiber, line 8: should be 
`checkboxStyles: React.CSSProperties`

holiber (approved):
Looks great

app/components-react/highlighter/ExportModal.tsx
```diff
@@ -0,0 +1,289 @@
...
+    const url = `https://youtube.com/watch?v=${v.uploadInfo.videoId}`;
+
+    return (
+      <div>
+        TODO
+        <br />
+        <a onClick={() => electron.remote.shell.openExternal(url)}>{url}</a>
+      </div>
+    );
+  }
```
gettinToasty, line 174: so a cool thing that React lets us do is define what in Vue would have been render functions like this to be full functional components that we can access with a much more readable `<MyComponent />` than a `getMyComponent()` not blocking at all but something to keep in mind as we move forward with React patterning

andycreeth: Yeah for sure, good call out!  I've been really enjoying the ease of easily extracting sub-render-functions into separate components.  This one is a great candidate for splitting up as it's gotten pretty big.

PR #3563, opened 2021-06-08
Port AdvancedStatistics to React

gettinToasty:
Also removes the global sync status icon


app/components-react/hooks.ts
```diff
@@ -150,3 +150,16 @@ export function useForceUpdate() {
   const [, forceUpdate] = React.useReducer(x => x + 1, 0);
   return forceUpdate;
 }
+
+export function useInterval(callback: () => void, delay: number) {
```
andycreeth, line 154: I think maybe we should rename this hook.  It forces a re-render every time the interval runs.  This is what you want here, and you're explicitly using that side effect.  But the name of the hook does not make it clear that's the case.

andycreeth (changes requested):

app/components-react/hooks.ts
gettinToasty: maybe something like `useRenderInterval`?

andycreeth (approved):

PR #3567, opened 2021-06-10
Highlighter internal beta features

andycreeth:
- working render preview
- upload success window with copy link button
- new onboarding that walks through setup automatically
- fix a bug with video file caching in memory
- better flow without YT linked that will show video output and file location


app/components-react/highlighter/ExportModal.tsx
```diff
@@ -166,20 +177,32 @@ function YoutubeUpload(props: { defaultTitle: string; close: () => void }) {
 
     return (
       <div>
-        TODO
+        <p>
+          Your video was successfully uploaded! Click the link below to access your video. Please
+          note that YouTube will take some time to process your video.
+        </p>
```
gettinToasty, line 183: probably want to add these to i18n right?

gettinToasty (changes requested):

gettinToasty (approved):

PR #3569, opened 2021-06-11
Redux integration

holiber:
Introduces modular Flux architecture based on Redux as a replacement for the custom Flux implementation in `StateManager.ts`

Unfortunately, this PR also includes changes from the PR https://github.com/stream-labs/streamlabs-obs/pull/3493 that we reverted in https://github.com/stream-labs/streamlabs-obs/pull/3564.
Current PR was based on the reverted branch and heavily depends on it


app/components-react/windows/go-live/useGoLiveSettings.ts
```diff
@@ -1,294 +1,213 @@
...
+  get settings() {
+    return this.state;
+  }
 
-  // creates a StreamInfoView object based on provided state
-  function getView(state?: TState) {
-    const stateGetter = state ? () => state : getState;
-    return new StreamInfoView(StreamingService.state, stateGetter);
+  getSettings() {
+    return this.state;
```
gettinToasty, line 73: do we need 2 nearly identical functions here?

app/components-react/windows/go-live/useGoLiveSettings.ts
```diff
@@ -1,294 +1,213 @@
...
-        case 'multiplatformAdvanced':
-          return [essentialOptionalFields, requiredFields, optionalFields, commonFields];
-      }
-    },
-  };
+  /**
+   * Update top level settings
+   */
+  @mutation()
+  updateSettings(patch: Partial<IGoLiveSettingsState>) {
```
gettinToasty, line 101: are we ditching screaming snake  case for mutations with this switch? if so, what's the thinking?

app/services/app/app.ts
```diff
@@ -148,7 +148,8 @@ export class AppService extends StatefulService<IAppState> {
       this.shutdownHandler();
     });
 
-    this.performanceService.startMonitoringPerformance();
+    // TODO: enable
+    // this.performanceService.startMonitoringPerformance();
```
gettinToasty, line 152: we need to enable this, right?

app/services/streaming/streaming-view.ts
```diff
@@ -45,16 +37,16 @@ export class StreamInfoView extends ViewHandler<IStreamingServiceState> {
     return this.getServiceViews(TwitterService);
   }
 
-  private get incrementalRolloutView() {
-    return this.getServiceViews(IncrementalRolloutService);
+  private get streamingState() {
+    return Services.StreamingService.state;
```
gettinToasty, line 41: does the service provider work like this in views and other services?

gettinToasty (changes requested):

app/components-react/hooks/useModule.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+    return {
+      module,
+      selector,
+      select,
+    };
+  });
+
+  // unregister the component from the module onDestroy
+  useOnDestroy(() => {
```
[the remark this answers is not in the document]
holiber: Let discuss it in Slack

app/components-react/windows/go-live/GoLiveSettings.tsx
```diff
@@ -29,11 +29,11 @@ export default function GoLiveSettings() {
...
     error,
-    canAddDestinations,
     isLoading,
-  } = useGoLiveSettings(view => {
-    const linkedPlatforms = view.linkedPlatforms;
-    const customDestinations = view.customDestinations;
+    canAddDestinations,
+  } = useGoLiveSettings().selectExtra(module => {
+    const linkedPlatforms = module.linkedPlatforms;
+    const customDestinations = module.customDestinations;
```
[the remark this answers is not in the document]
holiber: I was unable to describe types for `select` in the way to support `selectExtra` syntax. Because of 2 tsconfigs I have weird issues here. I'll fix that when we switch to a single tsconfig

app/services/app/app.ts
holiber: Wow I didn't push this change

app/services/streaming/streaming-view.ts
holiber: Yes, it does. But I probably should rollback it to the previous style until we come with some convention about how we should inject services

gettinToasty: agreed, especially since views are explicitly only supposed to be able to access other views and state

package.json
```diff
@@ -82,6 +83,7 @@
     "overlayscrollbars-react": "^0.2.2",
     "raven": "^2.6.4",
     "react-devtools-electron": "^4.7.0",
+    "react-redux": "^7.2.4",
```
andycreeth, line 86: Should be a dev dependency because it gets compiled into our webpack bundle.

package.json
```diff
@@ -62,6 +62,7 @@
     "serial": true
   },
   "dependencies": {
+    "@reduxjs/toolkit": "^1.5.1",
```
andycreeth, line 65: Should also be a dev dependency

andycreeth (changes requested):

PR #3570, opened 2021-06-11
Improve JSON-RPC implementation

malliaridis:
_This PR is an OPTIONAL improvement._

What does it change:
- Reduces the type and argument checks made for request, event, error and response creation in the `JsonrpcService`
- Resolves the eslint-disable-rules
- Marks optional arguments as such (this caused further ordering changes)
- Fixes a wrong function call in `JsonrpcService.createEvent`

For further details see commit messages.

Note: Since there is no information for executing tests, multiple tests executed with `npm run test` fail

_Feel free to squash commits, cherry-pick only the bug fix or reject the whole PR_ :)

### DIFF
--- a/app/services/api/jsonrpc/jsonrpc.ts
+++ b/app/services/api/jsonrpc/jsonrpc.ts
@@ -107,6 +107,6 @@ export class JsonrpcService extends Service implements IJsonrpcServiceApi {
     data: any;
     isRejected?: boolean;
   }): IJsonRpcResponse<IJsonRpcEvent> {
-    return JsonrpcService.createResponse.apply(this, options);
+    return JsonrpcService.createEvent.apply(this, options);
   }
 }
--- a/app/services/api/jsonrpc/jsonrpc.ts
+++ b/app/services/api/jsonrpc/jsonrpc.ts
@@ -10,22 +10,14 @@ import uuid from 'uuid/v4';
 
 export class JsonrpcService extends Service implements IJsonrpcServiceApi {
   static createError(
-    requestOrRequestId: string | IJsonRpcRequest,
     options: { code: E_JSON_RPC_ERROR; message?: string },
+    request?: IJsonRpcRequest,
   ): IJsonRpcResponse<any> {
-    /* eslint-disable */
-    const id =
-      arguments[0] && typeof arguments[0] === 'object'
-        ? (arguments[0] as IJsonRpcRequest).id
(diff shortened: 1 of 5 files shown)


holiber:
A lot of our automated tests require access to a remote service to handle authorization. So you are not able to run the entire test suit.
However, your changes require only `API` tests to run. To run `API` tests compile the project and run 
`yarn test:file test-dist/test/regular/api/*`
^ this command will run only tests from the `api` folder

malliaridis:
Ok, I can confirm then that tests have been executed successfully too (except `afterEach.always hook for Streaming to Twitch via API` which probably requires that remote service).

holiber (approved):

andycreeth:
Heads up that I've reverted this PR.  It's causing an issue when calling `jsonrpcService.createRequest` where the arguments will get wrapped into an array, causing them not to work properly.

app/services/api/jsonrpc/jsonrpc.ts
```diff
@@ -78,23 +68,21 @@ export class JsonrpcService extends Service implements IJsonrpcServiceApi {
...
   ): IJsonRpcResponse<any> {
-    // eslint-disable-next-line
-    return JsonrpcService.createError.apply(this, arguments);
+    return JsonrpcService.createError.apply(this, [options, request]);
   }
 
   createRequest(resourceId: string, method: string, ...args: any[]): IJsonRpcRequest {
-    // eslint-disable-next-line
-    return JsonrpcService.createRequest.apply(this, arguments);
+    return JsonrpcService.createRequest.apply(this, [resourceId, method, args]);
```
malliaridis, line 85: @andycreeth I assume changes of this type caused this behavior. Probably it needed to be `...args` instead of `args` to unwrap the content of `args` in the apply function?

malliaridis: Let me know if I should make a patch for this. :)

andycreeth: Yup that was the issue.  I opted to revert the PR rather than debugging in the moment, but I agree that seems like the fix.  If you prepare a new PR we can go through a more extensive set of tests this time.

malliaridis: No problem and sorry for any troubles I might have caused. I will prepare these days a PR and double check and test the changes too.

PR #3586, opened 2021-06-22
Port RenameSource to React

gettinToasty:


gettinToasty:
based against staging since this branch originated trying to bugfix a child window rendering issue-- this doesn't solve the bug but it's work we had to do anyway

app/components-react/windows/RenameSource.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  const form = useForm();
+
+  const options = useRef(WindowsService.getChildWindowQueryParams());
+
+  useEffect(() => {
+    const source = SourcesService.views.getSource(options.current.sourceId);
+    if (!source) return;
+    setName(source.name);
+  }, []);
```
andycreeth, line 21: What's the point of using an effect to initialize the state?  Why not just:
`const [name, setName] = useState(SourcesService.views.getSource(options.current.sourceId));`

That entire operation shouldn't involve any IPC and will be serviced entirely from the local Vuex store.

andycreeth (changes requested):

app/components-react/windows/RenameSource.tsx
andycreeth: If the goal is to avoid making that call on every render (not a bad thing to think about), then the best way to handle this free of race conditions would be with `useMemo`:

```TypeScript
const currentName = useMemo(() => SourcesService.views.getSource(options.current.sourceId));
const [name, setName] = useState(currentName);
```

gettinToasty: yea i was actually thinking about that myself re: initializing with it or not, but yea i think `useMemo` or `useRef` is probably a better fit

gettinToasty: just read up on it-- apparently the initialState argument of `useState` is only calculated on the first render so should be fine to directly initialize in there

andycreeth (approved):

PR #3589, opened 2021-06-23
Some highlighter fixes/improvements

andycreeth:
- Fix initialization bug
- Add clip trim button
- Add clip removal
- Add header bar
- Add ability to skip tutorial

### DIFF
--- a/app/components/SideNav.tsx
+++ b/app/components/SideNav.tsx
@@ -16,6 +16,7 @@ import styles from './SideNav.m.less';
 import { LayoutService } from 'services/layout';
 import { getPlatformService } from '../services/platforms';
 import { getOS, OS } from 'util/operating-systems';
+import Utils from 'services/utils';
 
 interface IPageData {
   target: TAppPage;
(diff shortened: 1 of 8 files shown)


app/components/SideNav.tsx
```diff
@@ -130,7 +131,9 @@ export default class SideNav extends Vue {
     if (
       getOS() === OS.Windows &&
       this.userService.isLoggedIn &&
-      this.incrementalRolloutService.views.featureIsEnabled(EAvailableFeatures.highlighter)
+      this.incrementalRolloutService.views.featureIsEnabled(EAvailableFeatures.highlighter) &&
+      // TODO: Remove via bundle when v1 is complete and ready to start rolling out
+      (Utils.isPreview() || Utils.isDevMode())
```
holiber, line 136: the feature is always enabled in DevMode

holiber (approved):

app/components/SideNav.tsx
andycreeth: Yeah the purpose of this check is to ensure it's always disabled in production (for now).

andycreeth: We don't have a `Utils.isProductionButNotPreview()` check so that's what this is effectively.

PR #3593, opened 2021-06-24
Port NotificationsArea to React

gettinToasty:
Blocked by #3563 which introduces useInterval


holiber (approved):

holiber:
We probably should use `antd.message()` here

PR #3600, opened 2021-07-01
don't rely on videos directory for initial store state

andycreeth:


app/services/highlighter/index.ts
```diff
@@ -360,6 +360,17 @@ export class HighlighterService extends StatefulService<IHighligherState> {
   }
 
   init() {
+    try {
+      // On some very very small number of systems, we won't be able to fetch
+      // the videos path from the system.
+      // TODO: Add a fallback directory?
+      this.SET_EXPORT_INFO({
+        file: path.join(electron.remote.app.getPath('videos'), 'Output.mp4'),
```
holiber, line 368: What is Output.mp4 ?

PR #3605, opened 2021-07-03
Improve JSON-RPC implementation (2)

malliaridis:
This PR fixes major bugs added by e3b235b and improves further the JSON-RPC implementation.


malliaridis:
@andycreeth is there a specific reason why the `apply` function is used? Cause I read a little about it and found out, that it is an untyped function that is mainly used in JavaScript. But since we write in TypeScript it might be less error prone if we replace the function with direct calls to static functions.

If the `apply` function is not necessary, I would start replacing that too and simplify the arguments "array" notation too. That way TypeScript will warn if something is mismatching.

holiber:
This project originally started as a Javascript project, Even when it was migrated to Typescript it was a pretty old typescript version.

andycreeth:
Hi, thanks again for your interest in contributing to Streamlabs Desktop.  After discussing with the team, we aren't comfortable merging this PR at this time.  If you feel these changes are still warranted, we would really appreciate it if you could describe the benefits of these changes to API consumers.  Thank you again, and sorry for the delay in responding to this PR.

PR #3608, opened 2021-07-06
Add option to use Twitch URL in tweet body

gettinToasty:
This is a really nice example of the `useEffect` dependency array working to full effect here


andycreeth (approved):
Looks good 👍

PR #3613, opened 2021-07-07
Prevent multiple simultaneous theme installs

gettinToasty:


app/components/pages/BrowseOverlays.vue.ts
```diff
@@ -86,9 +93,11 @@ export default class BrowseOverlays extends Vue {
         overlayName: name,
       });
     } else {
+      this.downloadInProgress = true;
       const sub = this.sceneCollectionsService.downloadProgress.subscribe(progressCallback);
       await this.sceneCollectionsService.installOverlay(url, name);
```
andycreeth, line 98: To be 100% safe it might be a good idea to wrap the rest of the code in a `try/catch` that also sets the flag to false.  Make sure to re-raise the error so it still gets surfaced up to the themes library.

andycreeth (changes requested):

andycreeth (approved):

PR #3622, opened 2021-07-08
Fix selective recording projector in fullscreen

gettinToasty:


app/util/menus/EditMenu.ts
```diff
@@ -335,11 +336,13 @@ export class EditMenu extends Menu {
     this.append({
       label: $t('Create Stream Output Projector'),
       click: () => this.projectorService.createProjector(ERenderingMode.OBS_STREAMING_RENDERING),
+      enabled: this.streamingService.state.selectiveRecording || Utils.isDevMode(),
```
andycreeth, line 339: Dev mode behavior is for Trevor to test, or?

andycreeth (approved):

app/util/menus/EditMenu.ts
gettinToasty: yep exactly

PR #3643, opened 2021-07-20
Load Browser Extensions on Session Creation from Local Directory

sirsparrow:
I was looking for a way to add in support for a Twitch Chat Pronouns extension (https://pronouns.alejo.io/) and decided to generically load extensions that were placed in an `extensions` directory.

This way updates to extensions can be performed manually by users and nothing needs to be distributed with the install.

If there are PR submission / contribution guidelines that I missed somewhere, please let me know.  I decided to keep this small and not bother offering a UI to manage extensions.  I also am not certain if there's a better directory location that should be used so I just placed it directly in the application path.


sirsparrow:
Just checking back in to see if there is an appetite for this tweak, or if there is anything that I'm meant to be doing to push this along (or have it denied as not wanted).  Thanks!

andycreeth:
@sirsparrow Thank you for your submission, and I want to apologize for taking so long to get back to you about this.  We can see how this would be a useful change for some users, but unfortunately we're not comfortable with the security implications of loading arbitrary extensions off disk currently.  We are going to look into options for integrating the Twitch Chat Pronouns extension natively, but for now we are unable to merge this PR.  Thanks again for taking the time to work on this and for you interest in Streamlabs OBS.

PR #3646, opened 2021-07-21
Fix jank in Poll image and notif area

gettinToasty:


andycreeth (approved):

app/components/widgets/Poll.tsx
```diff
@@ -13,6 +13,15 @@ import { $t } from 'services/i18n';
...
 export default class Poll extends WidgetSettings<IPollData, PollService> {
   UserService = Services.UserService;
+  TwitchService = Services.TwitchService;
+  OnboardingService = Services.OnboardingService;
+  WindowsService = Services.WindowsService;
+
+  hasPollScopes = false;
+
+  async mounted() {
+    this.hasPollScopes = await this.TwitchService.hasScope('channel:manage:polls');
```
andycreeth, line 23: I wonder if we should memo-ize this somewhere so that we don't re-fetch it every time the properties open?

PR #3648, opened 2021-07-23
Tests refactoring: part 1

holiber:
- Fix tests not working with Antd modals
- Create the form util that works with React inputs (as a replacement for FormMonkey)
- Add tests for React inputs (TextInput, NumberInput, TextareaInput, ListInput, TagsInput, SwitchInput, CheckboxInput)
- Add streaming tests for React version of the GoLiveWindow
- Reorganize helpers code by modules (core, user, streaming, onboarding )
- Speed-up some tests via`setTimeout({ implicit: 0 })
- Update the Antd library 4.10.3 -> 4.16.7


app/components-react/hooks.ts
```diff
@@ -89,10 +89,17 @@ export function useFormState<T extends object>(initializer: T | (() => T)): TUse
   // create a reference to the last actual state
   const stateRef = useRef(s);
 
+  // use isDestroyed flag to prevent updating state on destroyed components
```
andycreeth, line 92: I this mainly a performance optimization?

andycreeth (approved):
Looks good.  We should probably do some extensive testing on this one because it updates antd.

app/components-react/hooks.ts
holiber: This is a quick fix for the case when a finished xhr request executes a callback that changes the component's state but the component is destroyed. React write errors in the console in that case that cause the failure in automated tests.

For example, in FB tests we click `Confrm & GoLive` before the list of FB groups has been loaded.

PR #3661, opened 2021-07-31
Stream scheduler

holiber:


app/components-react/pages/stream-scheduler/StreamScheduler.tsx
```diff
@@ -0,0 +1,242 @@
...
+        {/* DELETE BUTTON */}
+        {shouldShowRemove && (
+          <Button danger onClick={onDeleteClick}>
+            {$t('Delete')}
+          </Button>
+        )}
+      </Col>
+      <Col flex={'50%'}>
+        {/*/!* GO LIVE BUTTON *!/*/}
+        {/*{shouldShowGoLive && <Button type="primary">{$t('Go Live')}</Button>}*/}
```
andycreeth, line 224: Should this be added back, or intentionally commented out?

andycreeth (approved):

app/components-react/pages/stream-scheduler/StreamScheduler.tsx
holiber: I think this should be done as a separate PR. It requires some additional changes in our services

PR #3699, opened 2021-08-17
Add analytics for the StreamScheduller

holiber:
Should be merged after https://github.com/stream-labs/streamlabs-obs/pull/3686


app/components-react/pages/stream-scheduler/useStreamScheduler.tsx
```diff
@@ -207,6 +207,12 @@ class StreamSchedulerModule {
 
   getPlatformDisplayName = this.streamingView.getPlatformDisplayName;
 
+  private recordFeatureUsage(
+    featureName: 'StreamSchedulerView' | 'StreamSchedulerEdit' | 'StreamSchedulerGoLive',
+  ) {
+    this.recordFeatureUsage('StreamSchedulerView');
```
andycreeth, line 213: This looks like infinite recursion?

andycreeth (changes requested):

app/components-react/pages/stream-scheduler/useStreamScheduler.tsx
holiber: good catch

andycreeth (approved):

andycreeth (approved):

PR #3705, opened 2021-08-18
Store game id instead of name for in the TwitchService

holiber:


app/components-react/windows/go-live/useGoLiveSettings.ts
```diff
@@ -46,6 +47,11 @@ export class GoLiveSettingsModule extends StreamInfoView<IGoLiveSettingsState> {
    */
   async prepopulate() {
     await Services.StreamingService.actions.return.prepopulateInfo();
+
+    // TODO: investigate why mutations from `StreamingService.prepopulateInfo` are not applied at this moment
```
andycreeth, line 51: I think we discovered this issue before - I had a mutation refactor that was pretty ambitious and it caused some issues.  We never revisited it.

app/components-react/windows/go-live/useGoLiveSettings.ts
```diff
@@ -46,6 +47,11 @@ export class GoLiveSettingsModule extends StreamInfoView<IGoLiveSettingsState> {
...
+    // use `sleep` to wait for mutations as a temporary fix
+    await Utils.sleep(100);
```
andycreeth, line 53: This makes me a bit nervous.
A) it's bad for perceived perf, and
B) 100ms may not be enough for everyone

andycreeth (changes requested):
IMO we should find a way to avoid adding the `sleep`.  This refactor is not worth adding the `sleep`, especially on such an important function in such an important window.  I'm not comfortable merging this.

app/components-react/windows/go-live/useGoLiveSettings.ts
holiber: I tried to understand what is going on here. Yeah, it seems we have an issue with mutations order when we have a deal with Promises. That requires some time to investigate the issue and it's going to be a very dangerous change. So I prefer to revisit it at the moment when I'm going to migrate all our services to Redux.  So I decided to stay with this dirty `sleep` fix here

app/components-react/windows/go-live/useGoLiveSettings.ts
holiber: A) It takes about 1s to prepopulate platforms first time for me. Additional 100ms doesn't make a big difference
B) I took 100ms as a huge reserve. Actually, we need only time for executing one IPC request here which is usually 2 - 10ms
I agree that looks like a hack. We should fix an original issue with incorrect mutations order here

app/components-react/windows/go-live/useGoLiveSettings.ts
andycreeth: Let's hold off on this PR until we fix it.  I can look into it, since I already tried fixing it before.  And yeah it was a dangerous change that we ended up having to revert 😬

andycreeth: https://github.com/stream-labs/streamlabs-obs/pull/2757 I actually made this fix last year which was smaller and didn't get reverted.  It was meant to fix this issue, but maybe it didn't catch all the cases.

gettinToasty:
where are we with this?

holiber:
We consider these changes unacceptable. We need to investigate the issues with mutations first

PR #3706, opened 2021-08-19
Bugfix for Scene.getNestedItems()

enzeart:
Scene.getNestedItems uses the current scene instance to look up items that are returned by a helper method. The helper method may return scene items that belong to nested scenes. When it uses the current scene instance to look up the scene items that belong to nested scenes, a null value is returned.

This pull request is to fix this so the expected scene items are always returned.

### DIFF
--- a/app/services/api/external-api/scenes/scene.ts
+++ b/app/services/api/external-api/scenes/scene.ts
@@ -142,7 +142,9 @@ export class Scene implements ISceneModel, ISerializable {
    * @returns A list of all scene items of this scene and all nested scenes
    */
   getNestedItems(): SceneItem[] {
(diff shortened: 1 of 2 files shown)


enzeart:
Am I supposed to open a pull request for the master branch or the staging branch?

andycreeth:
Hi @enzeart - since this PR only contains changes to typescript files in the `/app` directory, it can go straight into `master`.  Can you rebase this PR onto `master` and then we can get tests running and get it merged.  Thank you!

enzeart:
Hey @andycreeth - I've rebased my changes and updated the target branch of the PR. Please let me know if you need anything else.

PR #3730, opened 2021-09-07
update logo in onboarding splash image

andycreeth:


app/services/utils.ts
```diff
@@ -264,6 +264,9 @@ export function keys<T>(target: T) {
  */
 export function $i(path: string) {
   try {
+    // Useful for testing media fetches properly from the CDN
+    if (process.env.SLOBS_USE_CDN_MEDIA) throw new Error('Using CDN');
```
holiber, line 268: In my last performance refactoring, I found that accessing `process.env` in the renderer process is a slow operation. I cached env variables in the `Utils.env`

andycreeth: Good call out, ty!

holiber (approved):

PR #3762, opened 2021-09-23
Support compilation on case-sensitive filesystems

Owersun:
For case sensitive file system there are two folders named differently (case wise) than there reference in the indexes.
Fix indexes to reference folders by their real names, to make compilation on the case sensitive system possible.


andycreeth (approved):
Thanks!

PR #3775, opened 2021-09-27
Update osn to v0.18.5

EddyGharbi:
EddyGharbi	Update libobs to v27.2.5 (#1009)


EddyGharbi:
Prob the build started while mac binaries were still being uploaded. Re
running this job.

EddyGharbi:
Actually no, it indeed failed on Mac.

PR #3881, opened 2021-11-09
Update README.md

loftwah:
I updated the version of Node.js last time and noticed it was out again. I have changed the wording so it doesn't need to be updated each time the LTS version changes.


andycreeth:
Makes sense, thank you.

PR #3884, opened 2021-11-12
working source filters window (needs cleanup)

andycreeth:


app/services/source-filters.ts
```diff
@@ -82,6 +94,15 @@ class SourceFiltersViews extends ViewHandler<{}> {
...
   }
 
+  get presetFilterOptionsReact() {
+    return this.presetFilterOptions.map(opt => ({
+      label: opt.title,
+      // Empty string doesn't play nicely with our react list input
+      // due to it being a falsey value in JS.
+      value: opt.value === '' ? 'none' : opt.value,
+    }));
+  }
```
gettinToasty, line 104: is there a reason to make this an additional mapping function vs just changing `presetFilterOptions`? are we still using that in vue somewhere?

app/services/source-filters.ts
```diff
@@ -91,25 +112,84 @@ class SourceFiltersViews extends ViewHandler<{}> {
...
   filterUpdated = new Subject<ISourceFilterIdentifier>();
   filtersReordered = new Subject<void>();
 
+  init() {
+    this.sourcesService.sourceAdded.subscribe(s => {
+      if (!this.state.filters[s.sourceId]) this.SET_FILTERS(s.sourceId, []);
+    });
+
+    this.sourcesService.sourceRemoved.subscribe(s => {
+      this.SET_FILTERS(s.sourceId, null);
```
gettinToasty, line 178: why do we set null here but empty array above?

app/services/scene-collections/nodes/sources.ts
```diff
@@ -261,6 +262,18 @@ export class SourcesNode extends Node<ISchema, {}> {
...
+        sourceInfo.id,
+        sourceInfo.filters.items.map(f => {
+          return {
+            name: f.name,
+            type: f.type,
+            visible: f.enabled,
+            settings: f.settings,
+          };
+        }),
+      );
```
gettinToasty, line 276: is this the fix for media backup filters?

gettinToasty (approved):

app/services/source-filters.ts
andycreeth: Yeah, the hardware onboarding is still Vue, and uses the preset values.

app/services/source-filters.ts
andycreeth: It's a little weird.  It's a dictionary of active sources.  So `null` or `undefined` means the source doesn't exist.  Whereas `[]` is a source that exists but doesn't have any filters.  I'm actually doing this a little wrong through, because I should be deleting the source id from the dictionary entirely, instead of just setting it to `null`.

app/services/scene-collections/nodes/sources.ts
andycreeth: This is actually just due to the fact that filters are loaded in a bulk function.  Without this, the store would be out of date, because the sources are being created in bulk by the backend, and the frontend doesn't know about them.  We didn't need this before because we weren't trying to keep an accurate picture of filters in the frontend store before.

app/services/source-filters.ts
andycreeth: Pushed a commit to address this.

PR #3888, opened 2021-11-15
fix obs form constraints

andycreeth:


holiber:
Which bug should it fix?

PR #3892, opened 2021-11-16
Redesign SourceShowcase Window

gettinToasty:


trevorrl:
Feedback: 
- the scroll bar overlaps the buttons which looks kinda jank: 
image

- selecting an Essential Source also highlights the same source in the General Sources section. not a huge deal but looks weird imo
image

- in the existing Add Source window, if the user has no apps installed then we don't show it, but in this redesign it shows regardless. IMO it should be hidden if no App Sources exist
image

holiber (approved):

andycreeth (approved):
Looks awesome 🎉 💯

andycreeth (approved):

PR #3893, opened 2021-11-16
Remove `emptyVal` from ListInputs

holiber:
This is refactoring for selecting a default value in React ListInputs in the SettingsWindow. Technically, it should not change the UI


andycreeth:
Looks like you have a conflict on this branch

PR #3914, opened 2021-11-30
Fix flaky FB tests

holiber:


app/components-react/pages/stream-scheduler/useStreamScheduler.tsx
```diff
@@ -217,9 +217,9 @@ class StreamSchedulerModule {
    * Shows a modal for creating a new event
    */
   @mutation()
-  showNewEventModal(platform: TPlatform, selectedTime?: Moment) {
+  showNewEventModal(platform: TPlatform, selectedTime?: number) {
```
andycreeth, line 220: Ah yes all args of mutations need to be serializable otherwise there will be issues.

andycreeth (approved):

gettinToasty (approved):

app/components-react/pages/stream-scheduler/useStreamScheduler.tsx
andycreeth: Ah wait this is a redux module not a vuex module.  They look so similar :)

PR #3915, opened 2021-11-30
Port image_source to React

holiber:


app/components-react/windows/SourceProperties.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const { WindowsService, SourcesService } = Services;
+
+  // get source
+  const source = useMemo(() => {
+    const { sourceId } = WindowsService.getChildWindowQueryParams();
+    return SourcesService.views.getSource(sourceId);
+  }, []);
+
+  // define reactive variables
+  const [properties, setProperties] = useState(source ? source.getPropertiesFormData() : []);
```
gettinToasty, line 20: as i understand this will be running `getPropertiesFormData` every render, which we might want to avoid as i think it's a somewhat expensive operation

gettinToasty (changes requested):

app/components-react/windows/SourceProperties.tsx
andycreeth: Good callout.  This will happen on every render, and it is definitely an expensive operation.  The solution here is to pass a function as the argument to `useState` which returns the initial state.

app/components-react/windows/SourceProperties.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const hideStyleBlockers = useSelector(() => WindowsService.state.child.hideStyleBlockers);
+
+  function saveProperties(formData: TObsFormData) {
+    assertIsDefined(source);
+    // save source settings in OBS source
+    source.setPropertiesFormData(formData);
```
andycreeth, line 26: This needs to go through the EditorCommandsService as source properties change is an undoable action.

andycreeth: `EditSourcePropertiesCommand` is the name of the command

app/components-react/windows/SourceProperties.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  return (
+    <ModalLayout
+      scrollable
+      fixedChild={source && !hideStyleBlockers && <Display sourceId={source.sourceId} />}
+    >
+      <ObsForm value={properties} onChange={saveProperties} layout="horizontal" />
+    </ModalLayout>
+  );
+}
```
andycreeth, line 40: This window is missing some functionality that the original window had:
- The window will not automatically close when the underlying source is removed
- The window will not automatically update when the properties are changed programmatically, i.e. when doing an undo/redo operation

andycreeth (approved):

PR #3917, opened 2021-11-30
Port browser_source to React

holiber:


trevorrl:
not sure if I like how the page reloads with every keystroke in the URL field now instead of when you click out of the field

andycreeth:
@trevorrl Yes - we need to fix that before this goes live.  Loading intermediate pages as typing is a potential security risk - as a partially typed URL could in theory resolve to a valid but malicious website.

andycreeth (approved):

PR #3921, opened 2021-12-01
Port dshow_input to React

holiber:


trevorrl:
image
these options (resolution/fps/etc) should be disabled if "Device Default" is chosen for Resolution/FPS Type instead of custom.

andycreeth (approved):

PR #3935, opened 2021-12-08
Port Stream Settings to React

holiber:


trevorrl:
@holiber I am unable to save additional custom multistream destinations. Clicking Add will show the fields, but clicking Save does nothing. 

`TypeError: Cannot assign to read only property 'url' of object '#<Object>'`

holiber:
@trevorrl  should be fixed now

app/components-react/windows/settings/Stream.tsx
```diff
@@ -0,0 +1,435 @@
...
+
+/**
+ * Renders a Platform placeholder
+ */
+function Platform(p: { platform: TPlatform }) {
+  const platform = p.platform;
+  const { UserService, StreamingService } = Services;
+  const { canEditSettings, platformMerge, platformUnlink } = useStreamSettings();
+  const isMerged = StreamingService.views.checkPlatformLinked(platform);
+  const username = UserService.state.auth!.platforms[platform]?.username;
```
andycreeth, line 284: You're accessing `isMerged` and `username` in a non-reactive way, i.e. not using `useVuex`.  Was that intentional?

andycreeth (approved):

app/components-react/windows/settings/Stream.tsx
holiber: They don't need to be reactive here

PR #3938, opened 2021-12-09
Port RemoteControl settings to React

holiber:


trevorrl:
Some notes on this one: 

- previously, we'd show a blurred QR code as a placeholder whereas now it's a pure white square. not a huge deal but a worse UX imo and looks kind of ugly
- another minor thing: after revealing and clicking show details, in the Vue version clicking Show Details again would hide the details, that does not work anymore. 

neither of these are blockers/deal breakers but lead to a slightly worse experience I think, otherwise no functional issues.

app/components-react/windows/settings/Advanced.tsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
 import { ObsGenericSettingsForm } from './ObsSettings';
 
-export function AudioSettings() {
+export function AdvancedSettings() {
```
trevorrl, line 4: this breaks Advanced Audio settings

andycreeth (approved):

PR #3942, opened 2021-12-13
electron 14.2.2

andycreeth:


holiber:
You need to sync with master to make tests passed

PR #3964, opened 2021-12-22
change to show stream audio track even in simple output

trevorrl:
This fixes an edge case we missed when shipping the advanced audio window rework, reported by Thibaud here: https://app.asana.com/0/734380881425048/1201545277902728/f

tl;dr - we don't show audio tracks in this window if you're in Simple Output and don't have Twitch VOD track enabled. because of this, it's possible to get into a state where no audio is recorded/broadcast because you disabled Track 1 while it was showing, and then switched to Simple / disabled VOD track before re-enabling it. When in this state, you can only fix by switching back to Advanced or re-enabling Twitch VOD and checking the Track 1 box again, which means the workaround can't be done while live. 

My change makes it so that Track 1 shows on this window regardless of output mode, other modes are unchanged and only Track 1 is shown if in Simple / VOD track is disabled.


gettinToasty (approved):

gettinToasty:
i think this is somewhat confusing UX that may lead to more issues than it fixes-- lets discuss it before merging

PR #3972, opened 2022-01-05
Port LegacyEvents to React

gettinToasty:


app/components-react/editor/elements/LegacyEvents.tsx
```diff
@@ -0,0 +1,76 @@
...
+import useBaseElement from './hooks';
+import { Services } from 'components-react/service-provider';
+import BrowserView from 'components-react/shared/BrowserView';
+import styles from './RecentEvents.m.less';
+
+export default function LegacyEvents(p: { onPopout: () => void }) {
+  const { UserService, RecentEventsService, MagicLinkService } = Services;
+
+  const containerRef = useRef<HTMLDivElement>(null);
+  const [magicLinkDisabled, setMagicLinkDisabled] = useState(false);
```
andycreeth, line 14: Would recommend using a ref for this rather than state, for the following reasons:
- It's not used for rendering, it's just an internal logic flag.  Setting state will trigger an unnecessary re-render of the component.
- State semantics can lead to weird bugs when its used for something like this.  The reason is that when you set state it doesn't take effect immediately.  It will trigger a new render, and the new state will be reflected on that next render.  These are good semantics for something used in the rendering output, but can be weird semantics when used for internal logic within an event handler like this.  In this case I don't foresee it leading to any bugs, but just something to keep in mind.

Super minor thing - this will work just fine.  Just trying to make sure we are always weighing the tradeoffs of which tools is slightly better for each job as we do these ports.

andycreeth (approved):

PR #3973, opened 2022-01-06
fix for safe mode api issue

trevorrl:
not sure if/how/when this broke but when we fetch a user's Safe Mode settings via API, it doesn't return all required fields, so if the user tries to activate Safe Mode without changing any settings, it fails because we're not sending all required fields. 

my proposed fix will set each setting to whatever is in state if it isn't defined via the API fetch, that way if/when the API is fixed to actually return all required values, it won't break anything in our app


andycreeth (approved):

trevorrl:
closing this PR as the issue was resolved on the API side so no change needed on our end.

PR #3978, opened 2022-01-07
Code cleaning and function description

elmarceloc:


app/services/chat.ts
```diff
@@ -17,6 +17,12 @@ import { assertIsDefined } from 'util/properties-type-guards';
 import * as remote from '@electron/remote';
 import { SourcesService } from 'app-services';
 
+/**
+ * Enables the twitch's Better TTV emotes
```
andycreeth, line 21: Technically this function doesn't actually enable BTTV, it just returns the script that can be injected to enable it.  I think this comment could lead to confusion unless we update it to reflect that.

andycreeth (changes requested):
Hi, thank you for your interest in contributing to Streamlabs Desktop.  I've done a quick review of your submission, and found a minor issue.  Thank you.

app/services/chat.ts
elmarceloc: Thanks for the feedback

andycreeth (approved):

PR #3981, opened 2022-01-10
Port LayoutEditor to React

gettinToasty:


andycreeth (approved):
I think this PR probably needs extra attention from QA before we merge / release.

app/components-react/shared/inputs/ImagePickerInput.tsx
```diff
@@ -0,0 +1,31 @@
...
+          })}
+          onClick={() => p.onInput(option.value)}
+        >
+          {!p.isIcons && <img src={option.label} />}
+          {p.isIcons && <i className={option.label} />}
+        </div>
+      ))}
+    </div>
+  );
+});
```
holiber, line 31: We actually already have a `CardInput` component for that purpose.  You can see the example of usage inside `LayoutInput` component that we use in Alertbox settings

app/components-react/pages/layout-editor/LayoutEditor.tsx
```diff
@@ -0,0 +1,105 @@
...
+export default function LayoutEditor() {
+  const { LayoutService } = Services;
+
+  const { slottedElements, setBrowserUrl, currentLayout, showModal } = useLayoutEditor();
+
+  useEffect(() => {
+    if (slottedElements[ELayoutElement.Browser]) {
+      setBrowserUrl(slottedElements[ELayoutElement.Browser]?.src || '');
+    }
+  }, []);
```
holiber, line 24: That doesn't look like the right place for calling `setBrowserUrl`.
If you change the component's state after the initial mount that most cases an antipattern because it forces the component to render twice.

holiber (approved):

trevorrl:
image

Feel like spacing between these components should be a bit more normalized, there's way more space between the "Add Tab" button and the dropdown than between the dropdown and the trashcan.

trevorrl:
image

also the "Add Tab" modal is v jank.

trevorrl:
If you delete a custom layout, it's not reverting the shown layout to whatever layout is next in the list. 

Behavior in Live Build: 
46syN1GwwC

Behavior on this branch: 
2V3aOkV9fG

delankallen:
Webiste_element_jank

Some jank with the target URL input
The website element is also not loading sites

delankallen:
https://user-images.githubusercontent.com/27327746/150043821-fbe806a3-5425-4dda-a7f5-79461e456015.mp4

Tab name input not updating as user types

PR #3982, opened 2022-01-11
Port Projector to React

gettinToasty:


app/components-react/windows/Projector.tsx
```diff
@@ -0,0 +1,86 @@
...
+import Util from 'services/utils';
+import Scrollable from 'components-react/shared/Scrollable';
+import { Services } from 'components-react/service-provider';
+import { useVuex } from 'components-react/hooks';
+import styles from './Projector.m.less';
+
+export default function Projector() {
+  const { WindowsService, SourcesService } = Services;
+
+  const [oldBounds, setOldBounds] = useState<electron.Rectangle | null>(null);
```
andycreeth, line 15: Given that this isn't used for rendering HTML, but instead for remembering a value that's used by event handlers (`enterFullscreen` and `exitFullscreen`) semantically I would probably use a ref here instead of state.  The advantage of a ref here is it doesn't cause a re-render and you don't need to worry about stale state.

In this component I'm sure it doesn't make a huge difference, but just wanted to mention it for semantics reasons as in other components it can make a big difference I think.

app/components-react/windows/Projector.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  useEffect(() => {
+    const sourcesSub = SourcesService.sourceRemoved.subscribe(source => {
+      if (source.sourceId === sourceId) {
+        remote.getCurrentWindow().close();
+      }
+    });
+
+    return sourcesSub.unsubscribe;
+  }, []);
```
andycreeth, line 35: Could use `useSubscription` here.  One thing we should probably get better at is using higher level general hooks when possible.  Improves readability I think.

andycreeth (approved):
Have a couple nitpicking comments but otherwise looks great.

holiber (approved):

PR #3983, opened 2022-01-12
Enforce Form component for Antd Inputs

holiber:


gettinToasty (commented):
I found that we use some other Input components without wrapping them into form, and I don't want to introduce too many changes with this PR. Also, it looks like that only ListInput has issues when you're using it without a form.

gettinToasty (approved):

PR #3991, opened 2022-01-18
Fix test formatting due to react hooks

gettinToasty:


holiber:
Maybe it's better to just create an `eslint` exception for the files in the `test` folder?

andycreeth:
Yeah - can we just make it so this rule only applies to files in the `components-react` folder?  Assuming that every single function that starts with `use` is a React hook feels like a bit of a sledgehammer when >50% of the code in the app is not component code.

PR #3994, opened 2022-01-19
protect against unsupported transition types

trevorrl:
while investigating a user report, discovered that during obs import we don't handle unsupported transition types gracefully. added a protection so that if a user is using an unsupported transition type (like, added via plugin), we skip transition import to avoid breaking the user's scene collection

### DIFF
--- a/app/services/obs-importer.ts
+++ b/app/services/obs-importer.ts
@@ -423,7 +423,10 @@ export class ObsImporterService extends StatefulService<{ progress: number; tota
       configJSON.transitions &&
       configJSON.transitions.length > 0 &&
       // only import if it's a supported transition type
-      this.transitionsService.views.getTypes().map(t => t.value).includes(configJSON.transitions[0].id as ETransitionType)
+      this.transitionsService.views
+        .getTypes()
+        .map(t => t.value)
+        .includes(configJSON.transitions[0].id as ETransitionType)
     ) {
       this.transitionsService.deleteAllTransitions();
       this.transitionsService.createTransition(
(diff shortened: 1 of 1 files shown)


app/services/obs-importer.ts
```diff
@@ -419,7 +419,12 @@ export class ObsImporterService extends StatefulService<{ progress: number; tota
   importTransitions(configJSON: IOBSConfigJSON) {
     // Only import a single transition from OBS for now.
     // Eventually we should import all transitions
-    if (configJSON.transitions && configJSON.transitions.length > 0) {
+    if (
+      configJSON.transitions &&
+      configJSON.transitions.length > 0 &&
+      // only import if it's a supported transition type
+      Object.values<string>(ETransitionType).includes(configJSON.transitions[0].id)
```
andycreeth, line 426: The better way to do this would be to use `this.transitionsService.views.getTypes().map(t => t.value)`.  The reason is that available transitions varies by OS, so this will only return the transitions available on the currently running OS.

andycreeth (changes requested):

app/services/obs-importer.ts
trevorrl: wouldn't we want to avoid importing transitions that aren't supported on the current OS? afaik the only example at the moment is the "Motion" transition, which isn't supported on Mac, but ideally we avoid importing any transition that isn't valid on the currently running OS right?

andycreeth: Correct - sorry if it wasn't clear.  Your code isn't restricted to currently available transitions.  The code I provided is restricted to what's available.

trevorrl: ahhh, yep, I misunderstood. will update, ty!

andycreeth (approved):

PR #3997, opened 2022-01-20
add "spin wheel" button to minifeed

trevorrl:


gettinToasty (changes requested):
i think we probably only want to show this if they have a spin wheel widget

gettinToasty:
the spin wheel api call should also probably be in the widget service instead of recent events

gettinToasty (approved):

PR #4000, opened 2022-01-25
Add Superchat Goal

gettinToasty:


andycreeth (approved):
Thanks for taking care of this.  I believe this isn't deployed on the web side yet, right?

PR #4019, opened 2022-02-08
Fix color copy-paste

holiber:
Copy-paste was broken in RGBA color inputs in source filters


app/components-react/shared/inputs/ColorInput.tsx
```diff
@@ -68,7 +77,7 @@ export function ColorInput(p: TSlobsInputProps<{}, string | IRGBAColor>) {
             inputAttrs.onChange(`#${data.hex}`);
           } else {
             const rgb = hexToRGB(`#${data.hex}`);
-            inputAttrs.onChange({ ...rgb, a: inputAttrs.value.a });
+            inputAttrs.onChange({ ...rgb, a: 1 });
```
andycreeth, line 80: Curious why this change.  Wouldn't it make more sense to keep the alpha slider in place when using the eyedropper rather than resetting it to `1`?

andycreeth (approved):

PR #4028, opened 2022-02-10
Add GameWidget

gettinToasty:


app/components-react/shared/inputs/SliderInput.tsx
```diff
@@ -36,7 +36,7 @@ export const SliderInput = InputComponent((partialProps: TSliderInputProps) => {
     <InputWrapper {...wrapperAttrs}>
       <Row>
         <Col flex="auto" {...dataAttrs} data-role="input" data-value={inputAttrs.value}>
-          <Slider {...inputAttrs} />
+          <Slider {...inputAttrs} onChange={onChangeHandler} />
```
holiber, line 39: what is this change for?

app/components-react/widgets/common/useWidget.tsx
```diff
@@ -206,7 +206,9 @@ export class WidgetModule<TWidgetState extends IWidgetState = IWidgetState> {
    * Update settings and save on the server
    */
   public async updateSettings(formValues: any) {
-    const newSettings = { ...this.settings, ...formValues };
+    console.log(formValues);
+    const newSettings = merge(cloneDeep(this.settings), formValues);
+    console.log(newSettings);
```
holiber, line 211: remove logs?

app/services/widgets/widgets-config.ts
```diff
@@ -1,16 +1,12 @@
 import { AnchorPoint } from '../../util/ScalableRectangle';
 import { $t } from '../i18n';
 import { TAlertType } from './alerts-config';
+import { WidgetType } from './widgets-data';
 
-export type TWidgetType = 'AlertBox' | 'ViewerCount';
+export type TWidgetType = WidgetType.AlertBox | WidgetType.ViewerCount | WidgetType.GameWidget;
```
holiber, line 6: What are benefits of using enums instead strings here? I think since Typescript introduced strict string types we don't need enums in most cases. Strings are easier for debugging

app/services/widgets/widgets-data.ts
```diff
@@ -682,4 +668,13 @@ export const WidgetDisplayData = (platform?: string): { [x: number]: IWidgetDisp
...
   },
+  [WidgetType.GameWidget]: {
+    name: $t('Game Widget'),
+    description: $t('Let your viewers play a game in chat'),
+    demoVideo: false,
+    demoFilename: 'game-widget.png',
+    supportList: [],
+    platforms: new Set(['twitch']),
+    icon: 'icon-face-masks',
+  },
```
holiber, line 679: I think we should eventually migrate all widget configuration to the single file. I created `widget-config.ts` for that but it's not connected in the SourcesShowcase yet

app/components-react/widgets/GameWidget.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+function GameOptions(p: { game: string }) {
+  const { settings, updateSettings } = useGameWidget();
+  const game = p.game;
+
+  function updateGameOption(key: string) {
+    return (value: any) => {
+      updateSettings({ game_options: { [game]: { [key]: value } } });
+    };
+  }
```
holiber, line 86: You lose a type checking in this function. It accepts any string here

You may create a type `type TGameOptionName = keyof  %your_interface%`

Or you can use an `createBinding` util that generates types-safe `onChange` handlers for you

LMK if you need help here

app/components-react/widgets/GameWidget.tsx
```diff
@@ -0,0 +1,171 @@
...
+          cannot_play_here: string;
+        };
+      };
+    };
+  };
+}
+
+function msToS(ms: number) {
+  return `${Math.floor(ms / 1000)}s`;
+}
```
holiber, line 40: We use the same formatter in many components. Lets use `metadata.seconds` to avoid code duplication.
The example is here https://github.com/stream-labs/desktop/pull/4030

gettinToasty: since this will be merged before #4030, i'll refactor this in a separate PR :)

app/services/widgets/widgets-config.ts
gettinToasty: it's not so much a benefit, but rather a necessity of using `widgets-config` in place of things like `WidgetDefinitions` where we already have a ton of logic built around our enumerable types. i think eventually when we port all of our widget to React this could be worth exploring well but i don't think it's feasible whilst we have both React and Vue widgets.

app/components-react/shared/inputs/SliderInput.tsx
gettinToasty: leftover from my divisor prop, will remove

app/services/widgets/widgets-data.ts
gettinToasty: i think there is a decent amount of separation of function between `WidgetsConfig` and `WidgetsDisplayData`. WidgetsDisplayData is only used in the SourceSelector and SourceShowcase, and cares about different things (like primary platform) which is unnecessary for WidgetsConfig, which appears to mostly handle how the widget actually functions and cares about things like the host and user token. it's technically two sources of truth, but if you look at it closely, there is zero overlap in function between them.

app/components-react/widgets/GameWidget.tsx
holiber: The `metadata.seconds` helper is already in the master

app/services/widgets/widgets-data.ts
holiber: Just look at the way how we add new widgets. The more files you should modify for a new widget the more complexity it brings. There is no action require for this PR here. But I'm going to switch widget to a single config in the future.

app/services/widgets/widgets-config.ts
holiber: I think I already have some of the code that relies on the string types. 
Should it help if you just add an `id` field here?

```
GameWidget: { 
 id: WidgetType.GameWidget
 ....
```

So we can keep a new simplified style for the new React widgets and have the backward capability with old Vue code.
Or did I miss something?

gettinToasty: yea, what i mean is there are several places in the sources/widgets services that don't have access to the string type of the widget but do have access to the enum

app/components-react/widgets/GameWidget.tsx
```diff
@@ -0,0 +1,208 @@
...
+        <Menu.Item key="game">{$t('Game Settings')}</Menu.Item>
+      </Menu>
+      <Form>
+        {!isLoading && activeTab === 'general' && (
+          <>
+            <SliderInput
+              label={$t('Chat Decision Time')}
+              tooltip={{
+                title: $t(
+                  "The duration in seconds to collect chat's responses before processing passing them to the game.",
```
andycreeth, line 64: Seems like copy here isn't quite right.  Maybe `processing` was meant to be removed?

app/services/widgets/widgets-data.ts
```diff
@@ -682,4 +668,13 @@ export const WidgetDisplayData = (platform?: string): { [x: number]: IWidgetDisp
     platforms: new Set(['twitch']),
     icon: 'icon-community',
   },
+  [WidgetType.GameWidget]: {
+    name: $t('Game Widget'),
+    description: $t('Let your viewers play a game in chat'),
+    demoVideo: false,
+    demoFilename: 'game-widget.png',
```
andycreeth, line 675: Are these hooked up to the CDN fallback / are these up on the CDN yet?

andycreeth (changes requested):

app/components-react/widgets/GameWidget.tsx
gettinToasty: yea this was just copy-pasted from the site, can change

app/services/widgets/widgets-data.ts
gettinToasty: yep!

andycreeth (approved):

holiber (approved):

PR #4036, opened 2022-02-16
Appearance settings changes and announcements opt-out

andycreeth:


app/components-react/windows/settings/Appearance.tsx
```diff
@@ -56,25 +56,14 @@ export function AppearanceSettings() {
...
-          {...bind.folderSelection}
-          label={$t('Scene item selection mode')}
-          options={[
-            { value: true, label: $t('Single click selects group. Double click selects item') },
-            {
-              value: false,
-              label: $t('Double click selects group. Single click selects item'),
-            },
-          ]}
-        />
```
gettinToasty, line 70: why are we getting rid of this setting?

andycreeth: It's not gone, I just moved it to the bottom.  The goal here was to create a single section with all the chat settings in a group together.  This was the only one in that section that didn't belong.

gettinToasty (approved):

PR #4039, opened 2022-02-22
Fix audio playback in MediaGallery

gettinToasty:


app/components-react/windows/MediaGallery.tsx
```diff
@@ -62,9 +62,9 @@ export default function MediaGallery() {
     fetchGalleryInfo();
 
     return () => {
-      if (audio) {
-        audio.pause();
-        setAudio(null);
+      if (audio.current) {
+        audio.current.pause();
+        audio.current = null;
```
andycreeth, line 67: Is there an advantage to setting the ref to `null` here?  I notice you're initializing it with a value, so it seems like this code could be a lot simpler if you just removed `null` from the ref type, and could remove all your null checks.

According to docs here: https://developer.mozilla.org/en-US/docs/Web/API/HTMLAudioElement/Audio#memory_usage_and_management simply calling `pause` (and losing all references) is sufficient for the object to get garbage collected.

andycreeth (approved):

PR #4058, opened 2022-03-08
input jank in event filter window

trevorrl:
before: 
image

after: 
image


andycreeth (approved):
Looks good, thanks for fixing!

PR #4075, opened 2022-03-22
refactor: replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/app/components/shared/inputs/FormInput.vue.ts
+++ b/app/components/shared/inputs/FormInput.vue.ts
@@ -33,7 +33,7 @@ export default class FormInput extends BaseInput<any, IInputMetadata> {
     const type = this.options.type;
 
     // tslint:disable-next-line:prefer-template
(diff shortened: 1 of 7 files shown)


holiber:
Thanks for contributing

andycreeth (approved):
Thanks for your contribution.  We can get this merged in once tests pass and our Contributor License Agreement is signed.  The CLA-bot seems a little slow today so it might take a little time to load.  Thanks again for taking the time to contribute.

PR #4098, opened 2022-04-11
Port DonationTicker to React

holiber:


app/components-react/widgets/DonationTicker.tsx
```diff
@@ -0,0 +1,152 @@
...
+            {...bind.font_color3}
+            tooltip={tips.amountColorTooltip}
+          />
+        </>
+      )}
+    </WidgetLayout>
+  );
+}
+
+export class DonationTickerModule extends WidgetModule<IDonationTickerState> {
```
gettinToasty, line 96: i don't think we need to export this right?

gettinToasty (approved):

PR #4127, opened 2022-04-22
Url protocol fixes

karljakober:
Twitch auth flow wasnt respecting proper protocol and local environment auth was not functioning.
There are still other areas this could be improved, and no other platforms were fixed.

Note: Twitch client id still needs to be manually overwritten.


andycreeth (approved):
Thanks for doing this!

PR #4137, opened 2022-04-27
Fix unreactive mute button in AdvancedAudio

gettinToasty:


app/components-react/windows/advanced-audio/SourceSettings.tsx
```diff
@@ -49,19 +49,26 @@ export default function AdvancedAudio() {
...
+    vodTrack,
+    muted,
+  } = useVuex(() => ({
+    isAdvancedOutput: SettingsService.views.isAdvancedOutput,
+    streamTrack: SettingsService.views.streamTrack,
+    recordingTracks: SettingsService.views.recordingTracks,
+    vodTrackEnabled: SettingsService.views.vodTrackEnabled,
+    vodTrack: SettingsService.views.vodTrack,
+    // Hack to make muted property reactive
+    muted: AudioService.views.getSource(p.source.sourceId).muted,
```
andycreeth, line 68: Would it be possible to just try `muted: p.source.muted`.  I don't believe re-creating the source is necessary here.  Just accessing the property within `useVuex` should be enough.

andycreeth (changes requested):

andycreeth (approved):

PR #4140, opened 2022-04-27
Port Alertbox Library to React

gettinToasty:


app/components-react/shared/BrowserView.tsx
```diff
@@ -151,5 +152,5 @@ export default function BrowserView(p: BrowserViewProps) {
     );
   }
 
-  return <div style={{ height: '100%' }} ref={sizeContainer} className={p.className} />;
+  return <div style={{ height: '100%', ...p.style }} ref={sizeContainer} className={p.className} />;
```
andycreeth, line 155: Side note - I really miss how Vue let you set a `style` attribute on custom components and it would automatically apply to the root DOM node of that component.

PR #4153, opened 2022-05-03
Port PlatformApp Pages to React

gettinToasty:


app/components-react/shared/PlatformAppPageView.tsx
```diff
@@ -0,0 +1,90 @@
...
+      }
+    });
+
+    return () => {
+      subscription.unsubscribe();
+      unmountContainer();
+    };
+  }, []);
+
+  useRenderInterval(checkResize, 100);
```
andycreeth, line 41: This seems like the wrong thing - this will cause a full re-render every 100ms.

andycreeth (changes requested):

andycreeth (approved):

PR #4176, opened 2022-05-27
Update tips in the Loader

gettinToasty:


andycreeth (commented):
Now that these are proper help tips, we should probably add them to our translations system right?

andycreeth (approved):

PR #4188, opened 2022-06-17
Port Manage Scene Collections window to React

gettinToasty:


app/components-react/windows/ManageSceneCollections.tsx
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  function duplicate() {
+    setDuplicating(true);
+
+    setTimeout(() => {
+      SceneCollectionsService.actions.return
+        .duplicate(p.collection.name, p.collection.id)
+        .then(() => setDuplicating(false))
+        .catch(() => setDuplicating(false));
```
andycreeth, line 139: FYI promises these days have a function called `finally()` that lets you schedule a callback for when the promise is settled, regardless or whether it resolves or rejects.  It's useful for behavior that you want to ensure always happens when the operation finishes, regardless of result.

andycreeth (approved):

PR #4189, opened 2022-06-18
namespace service helper mutations to their parent service

andycreeth:
All service helpers are now required to claim which service they are helpers for.

This is used to namespace their mutations.  For example, the mutation `Scene.SET_NODES_ORDER` would now be `ScenesService.Scene.SET_NODES_ORDER`.

The purpose for this change is to make it clear which mutations modify which service's state.  Since service helpers have mutations, but don't actually own their own state, this creates a runtime association between service helper mutations and the underlying state they are mutating.

This should fix a reactivity issue with the new `useModule` code.  Specifically, the `ReactRoot` component looks at mutation names to track which components have a dependency on which state.  This should resolve issues with reactivity related to mutations on service helpers.

### DIFF
--- a/app/services/api/external-api/audio/audio-source.ts
+++ b/app/services/api/external-api/audio/audio-source.ts
@@ -36,7 +36,7 @@ export interface IAudioSourceModel {
  * API for audio source management. Provides operations related to a single
  * audio source like set deflection and mute / unmute audio source.
  */
-@ServiceHelper()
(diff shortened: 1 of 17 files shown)


gettinToasty (approved):

andycreeth:
Closing this PR because it's going to be merged as part of the source selector PR.

PR #4190, opened 2022-06-21
performance: cache proxied services in non-worker windows

andycreeth:
Creating proxies is somewhat expensive.  There's no need to be re-creating the proxy every time we access a service from a non-worker window.

### DIFF
--- a/app/services/core/service.ts
+++ b/app/services/core/service.ts
@@ -7,6 +7,7 @@ import { Subject } from 'rxjs';
 const singleton = Symbol('singleton');
 const singletonEnforcer = Symbol('singletonEnforcer');
 const instances: Service[] = [];
+const proxies: Service[] = [];
 
 /**
  * Makes all functions return a Promise and sets other types to never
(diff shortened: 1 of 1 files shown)


gettinToasty (approved):

app/services/core/service.ts
```diff
@@ -90,7 +91,14 @@ export abstract class Service {
 
   static get instance() {
     const instance = !this.hasInstance ? Service.createInstance(this) : this[singleton];
-    return this.proxyFn ? this.proxyFn(instance) : instance;
+
+    if (this.proxyFn) {
+      if (!proxies[this.name]) proxies[this.name] = this.proxyFn(instance);
```
gettinToasty, line 96: i just noticed that `proxies` is an array but here we key into it like an object, i presume that's not intentional?

gettinToasty (changes requested):

gettinToasty (approved):

PR #4197, opened 2022-06-29
Fix NotificationsArea issues

gettinToasty:


app/components-react/root/NotificationsArea.tsx
```diff
@@ -1,58 +1,116 @@
...
+          [styles.hasAction]: notif.action,
+        }),
+      })
+      .then(() => this.playNext());
+  }
+
+  // Play next Notification in the queue, otherwise clear the current Notification
+  playNext() {
+    if (this.currentNotif) {
+      message.destroy();
```
andycreeth, line 52: It seems weird to me that this is a global function rather than something we call on a specific message.

andycreeth (approved):
Seems good to me as long as we can fix these couple of issues discussed in slack.

app/components-react/root/NotificationsArea.tsx
gettinToasty: i ran into some issues trying to destroy a specific message, but can try messing around with it again

PR #4200, opened 2022-07-05
Added support for virtual camera toggle hot key

ryanfalzon:


ryanfalzon:
Any particular reason why the build validation failed? Can't seem to see the error.

ryanfalzon:
@andycreeth I have committed again with a fix for why the pipelines failed

andycreeth:
Hey @ryanfalzon thanks for your submission!  I had our QA team test this branch, and one thing they noted is that this toggle doesn't work if you assign both start and stop to the same hotkey (therefore acting as a toggle).  This should start working if you define `isActive` on these hotkey actions.  These functions should just return `true`/`false` based on whether that action is already active and shouldn't be applied.  So for example for `TOGGLE_VIRTUAL_CAMERA_ON`, `isActive` would return `true` if the virtual camera is already on.  And inversely, for `TOGGLE_VIRTUAL_CAMERA_OFF`, `isActive` would return `true` if the virtual camera is already off.

Once you add this, the hotkey should start working as a toggle if you binding both actions to the same key.  Let me know if that doesn't make sense.

Thanks again for your contribution and we would be happy to merge this once the above change is made.

ryanfalzon:
@andycreeth thanks for the feedback and sorry took so long to get back to you but have been busy with other projects. I have implemented the requested update and have pushed the changes. Let me know if there are any further issues with my PR.

andycreeth (approved):
Thanks for pushing that!  Our QA team is taking another pass at this but I think it should be good.  Thanks again for your contribution, we really appreciate it.

ryanfalzon:
@andycreeth just checking... was this PR merged or do I need to do anything else?

ryanfalzon:
@andycreeth any update on when this is going to be closed?

ryanfalzon:
@andycreeth any update on when this is going to be closed?

PR #4224, opened 2022-07-14
Fix file dragging onto source tree

gettinToasty:


holiber (approved):
Looks good, but let's do additional QA check

PR #4225, opened 2022-07-18
Hide action icons for empty folders

gettinToasty:


holiber (approved):
We need to test performance before releasing that

PR #4232, opened 2022-07-20
fixed a typo which prevents the child window to close correctly when the OBS initialization fails

eugen15:
In the current production code, if `obs.NodeObs.OBS_API_initAPI` fails, the child window process does not close. Some processes still stay in memory. This does not allow you to start **Streamlabs Desktop** again. The typo correction fixes the issue. 

It is interesting why the same issue does not happen when the application is closed by a user via the **close** button. In this case the uninitialization is different. The `shutdownHandler` method is called. It does `this.windowsService.shutdown()` which closes the windows.

```
private shutdownHandler() {
  ...
  window.setTimeout(async () => {
    ...
    this.windowsService.shutdown();
```

So, if the typo correction is not acceptable because `app.quit` was never actually called before, and it may break the other steps of the uninitialization, then you can add something like this `appService.windowsService.closeOneOffWindow('child')` to the following code before `obs.IPC.disconnect()`:

```
if (apiResult !== obs.EVideoCodes.Success) {
  const message = apiInitErrorResultToMessage(apiResult);
  showDialog(message);

  ipcRenderer.send('unregister-in-crash-handler', { pid: process.pid });

  obs.NodeObs.InitShutdownSequence();
  obs.IPC.disconnect();

  electron.ipcRenderer.send('shutdownComplete');
  return;
}
```


holiber (approved):
Looks good for me

PR #4236, opened 2022-07-22
Add per-item controls to scenes and sources

gettinToasty:
https://app.asana.com/0/527249672602050/1202158132289892


andycreeth (approved):
Nice work - this is a big upgrade.  Let's wait until QA gets a chance to do some performance testing on this before merging, but looking at the changes I don't anticipate any regressions.

PR #4241, opened 2022-07-27
Unify sources with more descriptive names

gettinToasty:


app/i18n/en-US/source-props.json
```diff
@@ -1,8 +1,8 @@
 {
  "Image": "Image",
- "Color Source": "Color Source",
- "Browser Source": "Browser Source",
- "Media Source": "Media Source",
+ "Color Block": "Color Block",
+ "Browser Window": "Browser Window",
```
andycreeth, line 4: Agree with the others but I wonder if this one warrants more discussion with the team.  "Browser Source" is kind of an ubiquitous industry term at this point and I'm wondering if "Browser Window" might introduce more confusion.

gettinToasty: if you want to bring it up with the greater team feel free to do so :) eddy, trevor, and myself discussed this iteration with the goal of removing the redundant "source" from source names

andycreeth (approved):

PR #4264, opened 2022-08-17
Collab Cam - Implement multiple guests

andycreeth:
Full support for N guests (limited by server) and UI for assigning guests to sources and swapping between them.


app/services/guest-cam/index.ts
```diff
@@ -24,6 +24,14 @@ import { ENotificationType } from 'services/notifications';
 import { $t } from 'services/i18n';
 import { JsonrpcService } from 'services/api/jsonrpc';
 import { EStreamingState } from 'services/streaming';
+import Vue from 'vue';
```
gettinToasty, line 27: bit confused about why vue is imported here?

app/services/guest-cam/index.ts
```diff
@@ -639,24 +659,26 @@ export class GuestCamService extends StatefulService<IGuestCamServiceState> {
...
-    this.SET_GUEST({ name: event.data.name, streamId: event.data.streamId });
+    // If we have a vacant source, assign the guest now
+    const vacantSources = this.views.vacantSources;
+    const sourceId = vacantSources.length ? vacantSources[0].sourceId : undefined;
 
-    // Clean up any existing consumer before connecting the new one
-    if (this.consumer) {
-      this.consumer.destroy();
-    }
+    this.ADD_GUEST({ remoteProducer: event.data, sourceId, showOnStream: false });
```
gettinToasty, line 666: if i'm understanding this part right the first guest that joins will fill the first available slot? does that mean it's impossible to have sources tied to guests? could be a bit inconvenient for regular collaborators who have certain areas of the stream blocked for specific people

gettinToasty (approved):

app/services/guest-cam/index.ts
andycreeth: This was specifically for a `Vue.set` call I did inside a mutation.  `Vue.set` ensures Vuex reactivity in cases where you are doing something Vuex wouldn't normally recognize as reactive.  i.e. initializing a property on an object that wasn't there when the object was originally created.

app/services/guest-cam/index.ts
andycreeth: That's correct.  By default we assign people to slots in the order they come in.  However, guests are hidden until you show them, and included in this PR is the ability to swap guests between sources.  So I'm imagining that if you have a specific slot set up for a specific person, you would just swap the source they were assigned to before showing them on your stream.

Long term maybe we could figure out a way to "remember" this for regular guests, but I'm not entirely sure what this would like.

app/services/guest-cam/index.ts
gettinToasty: ah yep! just missed that set call :)

app/services/guest-cam/index.ts
gettinToasty: gotcha! sounds good, didn't realize there was a swapping function so i think that should be good enough as long as it's obvious

PR #4279, opened 2022-08-23
Notification scroll, preview opens notifications window, refactor to show news and notifications in pop out

michelinewu:


app/components-react/root/NotificationsArea.tsx
```diff
@@ -88,7 +88,7 @@ class NotificationsModule {
   clickNotif() {
     if (!this.currentNotif) return;
     Services.NotificationsService.actions.applyAction(this.currentNotif.id);
-    Services.NotificationsService.actions.markAsRead(this.currentNotif.id);
+    Services.NotificationsService.actions.showNotifications();
```
andycreeth, line 91: I think this will open the notifications window on all notifications.  I think we only want to do this for notifications that don't have any actions associated with them.

So for example:
Notification with action - 1) apply action, 2) mark as read
Notification without action - 1) show notifications window 2) mark as read

The reason for this behavior is that some notifications actually have an action that opens up a child window themselves.  This is mutually exclusive with opening up the notifications window.

app/components-react/root/StartStreamingButton.tsx
```diff
@@ -114,6 +114,14 @@ export default function StartStreamingButton(p: { disabled?: boolean }) {
 
     if (!primaryPlatform) return false;
 
+    if (
+      !!UserService.state.auth?.platforms &&
+      StreamingService.views.isMultiplatformMode &&
+      Object.keys(UserService.state.auth?.platforms).length > 1
+    ) {
+      return true;
+    }
```
andycreeth, line 123: Seems like these changes aren't meant to be part of this PR?

app/components-react/windows/NotificationsAndNews.tsx
```diff
@@ -0,0 +1,51 @@
...
+  useEffect(() => {
+    if (!!notificationGroups && notificationGroups.unread.length > 0) {
+      setSelectedTab(tabNames.notifications);
+    }
+  }, [notificationGroups]);
+
+  return (
+    <ModalLayout hideFooter>
+      <div className={styles.notificationsMenu}>
+        <Menu
```
andycreeth, line 36: I think we should switch to the `<Tabs>` component.  It's much simpler than the `<Menu>` component for implementing tabs and will reduce the complexity here a bit.  It keeps track of the currently active tab and has a simple prop for the default tab.

app/components-react/windows/NotificationsAndNews.tsx
```diff
@@ -0,0 +1,51 @@
...
+    notificationGroups: {
+      unread: NotificationsService.views.getUnread(),
+    },
+  }));
+
+  useEffect(() => {
+    if (!!notificationGroups && notificationGroups.unread.length > 0) {
+      setSelectedTab(tabNames.notifications);
+    }
+  }, [notificationGroups]);
```
andycreeth, line 31: This will probably no longer be an issue when you switch to using the `<Tabs>` component, but in general I don't think this is a good use of an effect.  Unless I'm missing something, the main purpose of this effect is just to set the tab we appear on when the window first opens.  That's much better done when initializing the state rather than afterwards with an effect.  Since there's async operations to get this information, it can easily be computed during initial render.

andycreeth (changes requested):
Looks good but had a few thoughts on this one.

app/components-react/root/NotificationsArea.tsx
michelinewu: Thanks for the clarification! I'll revert.

app/components-react/root/StartStreamingButton.tsx
michelinewu: Oh weird, I think I forgot to branch off of master since the fix is in https://github.com/stream-labs/desktop/commit/a3adc49cf966b00db74c4dd959e72f007706e94b already, which has been merged. I can remove this commit from the history?

gettinToasty: I think you can just merge master onto this branch and it should clear it up

app/components-react/windows/NotificationsAndNews.tsx
michelinewu: Done! And yes, definitely more elegant.

app/components-react/windows/NotificationsAndNews.tsx
michelinewu: Oh, ok thanks! I think it's something I picked up working with Redux, but also very good to know going forward. Thanks for the clarification!

app/components-react/root/StartStreamingButton.tsx
michelinewu: Ok, thanks @gettinToasty!

andycreeth: Looks like this still isn't merged up to master.  LMK if you need help getting rid of this.

andycreeth (approved):
Looks good to me!

PR #4285, opened 2022-08-26
Fix to multistream when going live from calendar.

michelinewu:


andycreeth (approved):
Looks good.  Let's make sure that QA plays with this.  It's strange that this could would have been added in the first place if it's no longer needed.

PR #4321, opened 2022-09-28
Merge obs28

EddyGharbi:


app/services/selection/selection.ts
```diff
@@ -158,6 +161,31 @@ export class Selection {
     return rootNodes.length === 1 && rootNodes[0].sceneNodeType === 'folder';
   }
 
+  isScaleFilterSelected(filter: EScaleType): boolean {
+    const items = this.getItems().filter(item => item.scaleFilter === filter);
+    return items.length > 0;
```
andycreeth, line 166: Not worth changing but JFYI there's a cool array function in JS called `some` that simplifies this a bit.  It returns true if any of the items in the array pass the test.  So this would become:
```TypeScript
return this.getItems().some(item => item.scaleFilter === filter);
```

andycreeth (approved):

PR #4323, opened 2022-10-04
Add ultra icons

michelinewu:


gettinToasty (approved):

app/components-react/shared/UltraIcon.tsx
```diff
@@ -0,0 +1,201 @@
...
+        >
+          <stop stopColor="#2DE8B0" />
+          <stop offset="0.488027" stopColor="#CBE953" />
+          <stop offset="0.758621" stopColor="#FFAB48" />
+          <stop offset="1" stopColor="#FF5151" />
+        </linearGradient>
+      </defs>
+    </svg>
+  );
+}
```
gettinToasty, line 201: it looks like `simple` and the "default" icon type are two different svgs? is this intentional?

app/services/recent-events.ts
```diff
@@ -171,12 +171,12 @@ const filterName = (key: string): string => {
     filter_subscription_9_months: $t('9 Months'),
     filter_subscription_12_months: $t('12 Months'),
     filter_subscription_minimum_enabled: $t('Min. Months'),
-    primesub: $t('Prime'),
+    primesub: $t('Ultra'),
```
gettinToasty, line 174: all of the replacements in this file actually refer to Twitch Prime subs, so they should remain as Prime

gettinToasty (changes requested):
Couple of small things, including something I didn't catch in my first review

(Also thanks for fixing the tests for me :))

app/services/recent-events.ts
michelinewu: Done.

app/services/recent-events.ts
```diff
@@ -375,7 +375,7 @@ class RecentEventsViews extends ViewHandler<IRecentEventsState> {
       });
     }
     if (event.sub_type === 'primepaidupgrade') {
-      return $t('has converted from a Prime Gaming sub to a %{tier} sub', {
+      return $t('has converted from an Prime Gaming sub to a %{tier} sub', {
```
gettinToasty, line 378: grammar check kekw

gettinToasty (changes requested):

app/services/recent-events.ts
michelinewu: Fixed - thanks!

gettinToasty (approved):

PR #4327, opened 2022-10-11
Port Video Settings to new API

gettinToasty:


app/components-react/shared/inputs/FormFactory.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+export default function FormFactory(p: {
+  metadata: IFormMetadata;
+  onChange: (key: string) => (value: TInputValue) => void;
+  values: Dictionary<TInputValue>;
+  formOptions?: FormProps;
+}) {
+  const form = useForm();
+
+  useEffect(() => form.setFieldsValue(p.values), []);
```
andycreeth, line 33: Not entirely sure what this is doing, but is an effect definitely what you want?  Since this is an effect, it will running after the first render and the component has been mounted in the DOM.

If you're just looking to only do something once, you use `useMemo` with an empty deps array to accomplish this synchronously.  Anyway, just wasn't sure what the semantics of this were, but in general I think we should try to reserve `useEffect` for things that need to happen after first render.

app/components-react/shared/inputs/FormFactory.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { FormProps } from 'antd/lib/form';
+import debounce from 'lodash/debounce';
+import * as inputs from './inputList';
+import { TSlobsInputProps } from './inputs';
+import Form, { useForm } from './Form';
+import { TInputMetadata } from './metadata';
+
+type TInputValue = string | number | boolean;
+
+const componentTable: Dictionary<React.FunctionComponent<TSlobsInputProps<{}, TInputValue>>> = {
```
andycreeth, line 11: For a slightly better type, you could specify that only `TInputType` is a valid key for the object.

```TypeScript
const componentTable: { [k in TInputType]: React.FunctionComponent<TSlobsInputProps<{}, TInputValue>> } = ...
```

app/services/settings-v2/video.ts
```diff
@@ -0,0 +1,206 @@
...
+//   baseWidth: number;
+//   baseHeight: number;
+//   outputWidth: number;
+//   outputHeight: number;
+//   outputFormat: EVideoFormat;
+//   colorspace: EColorSpace;
+//   range: ERangeType;
+//   scaleType: EScaleType;
+//   fpsType: EFPSType;
+// }
```
andycreeth, line 21: I take it this is commented out since you're using the interface exposed by OSN instead?  If so, should we delete this?

andycreeth (approved):
Looks good, just a few small comments.

app/services/settings-v2/video.ts
gettinToasty: This was copied over from the OSN file just for ease of my reference :) will remove

app/components-react/shared/inputs/FormFactory.tsx
gettinToasty: antd was having some issues with named forms/inputs populating values on initial render for some reason, so this is basically a way to ensure that all of the proper input values are displaying correctly, it might work with useMemo, i'll give it a shot

PR #4336, opened 2022-10-19
Custom widget for customizing code, etc.

michelinewu:


app/components-react/widgets/CustomWidget.tsx
```diff
@@ -0,0 +1,50 @@
...
+  patchBeforeSend(settings: ICustomWidgetState['data']['settings']): any {
+    return {
+      ...settings,
+      custom_enabled: settings.custom_enabled,
+      custom_html: settings.custom_html,
+      custom_css: settings.custom_css,
+      custom_js: settings.custom_js,
+      custom_json: settings.custom_json,
+    };
+  }
```
gettinToasty, line 49: I don't think we need these `patchAfter` and `patchBefore` functions here:

We use these functions when the api sends us data in a format that's un-standardized or less useful to transform the data into a "nicer" shape for desktop. Notably things like converting ms to s, or making nested data flatter. Here it doesn't look like we're preforming any transforming, so we can just omit these functions :)

gettinToasty (changes requested):

app/components-react/widgets/CustomWidget.tsx
michelinewu: Sounds good - thanks!

gettinToasty (approved):

PR #4339, opened 2022-10-21
Sidebar UI Updates

michelinewu:


app/components-react/sidebar/SideNav.m.less
```diff
@@ -1,147 +1,217 @@
...
+body:last-child {
+  :global(div.ant-tooltip) {
+    display: none;
+  }
+}
+
+body > div > div {
+  :global(div.ant-menu-submenu.ant-menu-submenu-popup.ant-menu) {
+    display: none;
+  }
```
gettinToasty, line 14: would really hope we can use some less generic selectors here

app/components-react/sidebar/SideNav.m.less
```diff
@@ -1,147 +1,217 @@
...
+}
 
-  .cell {
-    height: 35px;
-    font-size: 14px;
+.ant-menu-inline-collapsed-tooltip {
+  display: none;
+}
+
+.side-nav.ant-slider.ant-slider-handle.ant-tooltip-open {
```
gettinToasty, line 204: i imagine these are either old/irrelevant or need to be made `:global`?

app/components-react/sidebar/NavTools.tsx
```diff
@@ -91,64 +94,159 @@ export default function SideNav() {
...
   }
 
+  const handleAuth = () => {
+    if (isLoggedIn) {
+      remote.dialog
+        .showMessageBox({
+          title: $t('Confirm'),
+          message: $t('Are you sure you want to log out?'),
+          buttons: [$t('Yes'), $t('No')],
+        })
```
gettinToasty, line 104: i think ideally we want to use antd modals here if possible

app/components-react/sidebar/SideNav.tsx
```diff
@@ -1,255 +1,412 @@
...
+                      icon={<i className={tab.icon} />}
+                      onClick={() => navigateToStudioTab(tab.target, tab.trackingTarget, tab.key)}
+                    >
+                      {tab.title}
+                    </Menu.Item>
+                  ))
+                ) : (
+                  <Menu.SubMenu
+                    key={menuItem.key}
+                    title={$t(menuItem.title)}
```
gettinToasty, line 236: ah i see how the title translation is happening now... i guess that works but probably not a pattern we want to adopt widely 🤔

app/components-react/sidebar/NavTools.tsx
```diff
@@ -91,64 +94,159 @@ export default function SideNav() {
...
+            return (
+              <Menu.Item
+                key={menuItem.key}
+                title={$t(menuItem.title)}
+                icon={
+                  <div>
+                    <Badge count={<i className={cx('icon-pop-out-3', styles.linkBadge)} />}>
+                      <i className={menuItem.icon} />
+                    </Badge>
+                  </div>
```
gettinToasty, line 147: unnecessary div?

app/components-react/sidebar/NavTools.tsx
```diff
@@ -91,64 +94,159 @@ export default function SideNav() {
...
+                      {platform && (
+                        <PlatformLogo
+                          platform={platform?.type!}
+                          className={cx(
+                            styles.platformLogo,
+                            styles[`platform-logo-${platform?.type ?? 'default'}`],
+                          )}
+                        />
+                      )}
+                      <Typography.Text className={styles.username}>
```
gettinToasty, line 238: what's the purpose of this element?

app/components-react/sidebar/AppsNav.tsx
```diff
@@ -132,42 +140,64 @@ export default function AppsNav() {
...
+//     <div className={cx(styles.arrow, styles.up)} onClick={scrollUp}>
+//       <i className="icon-down" />
+//     </div>
+//   )}
+//   {downArrowVisible && (
+//     <div className={cx(styles.arrow, styles.down)} onClick={scrollDown}>
+//       <i className="icon-down" />
+//     </div>
+//   )}
+// </div>
```
gettinToasty, line 203: leftover?

app/components-react/sidebar/AppsNav.m.less
```diff
@@ -1,117 +1,124 @@
...
+//   top: 0;
+
+//   i {
+//     transform: translate(-50%, -50%) rotate(180deg);
+//   }
+// }
+
+// .arrow.down {
+//   bottom: 0;
+// }
```
gettinToasty, line 124: leftover?

app/components-react/pages/BrowseOverlays.tsx
```diff
@@ -12,7 +12,8 @@ import { Services } from 'components-react/service-provider';
 import { useVuex } from 'components-react/hooks';
 
 export default function BrowseOverlays(p: {
-  params: { type?: 'overlay' | 'widget-theme'; id?: string };
+  //TODO add 'site-theme' to type
+  params: { type?: 'overlay' | 'widget-themes' | 'site-themes'; id?: string };
```
gettinToasty, line 16: TODO completed?

gettinToasty (changes requested):
first of all this is a massive project so absolutely awesome that you were able to get all this done yourself!

most of it looks alright but there were a few instances where i noticed you implement index-matters code patterns, which always makes me a bit nervous when largely iterating over JS objects, and i think are at least partially responsible for some of the bugs below, which i noticed while playing around with the build:

- Did not start with legacy nav bar, even though the custom nav bar was disabled in the settings
- Enable button doesnt seem to do anything, and clicking it makes my displayed apps disappear, and seems to reset the customized settings
- hiding app store in this state makes installed apps re-appear, but only while the app store is hidden
- Theme and App stores, Dashboard don't have hover titles in compact mode
- red outline around displayed apps is ugly/maybe unintentional?
- inconsistent sizing and spacing of app cells due to word wrapping, even in compact mode
- app swapping is buggy/inconsistent

PS: when interpreting my comments i read this from bottom to top, as looking at the service layer first helps contextualize the rest of the changes for me, so if some of my comments don't make sense, that's probably why haha

app/services/side-nav/menu-data.ts
```diff
@@ -0,0 +1,357 @@
...
+  [MenuItem in EMenuItem]: IMenuItem | IParentMenuItem;
+};
+
+// the key for primary menu items must be the title
+// so that the menu can apply open menu items on startup
+export const SideNavMenuItems = (): TMenuItems => ({
+  [EMenuItem.Editor]: {
+    key: EMenuItemKey.Editor,
+    target: 'Studio',
+    title: EMenuItem.Editor,
```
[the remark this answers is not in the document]
michelinewu: I didn't know that! Will change

app/services/customization.ts
```diff
@@ -152,6 +153,7 @@ export class CustomizationService extends PersistentStatefulService<ICustomizati
     },
     enableCrashDumps: true,
     enableAnnouncements: true,
+    testProp: true,
```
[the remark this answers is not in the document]
michelinewu: Removed

app/services/layout/index.ts
```diff
@@ -206,6 +206,8 @@ export class LayoutService extends PersistentStatefulService<ILayoutServiceState
     this.REMOVE_TAB(this.state.currentTab);
   }
 
+  // @mutation()
```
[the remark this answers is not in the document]
michelinewu: Removed

app/services/layout/layout-data.ts
```diff
@@ -34,6 +34,7 @@ type ILayoutData = {
   };
 };
 
+// TODO: explore refactoring to this kind of export
```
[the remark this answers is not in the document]
michelinewu: Removed

app/components/windows/Main.vue.ts
```diff
@@ -27,6 +27,8 @@ import fs from 'fs';
 import * as remote from '@electron/remote';
 Vue.use(VueResize);
 
+// THIS
```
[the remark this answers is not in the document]
michelinewu: Removed

app/components/windows/Main.vue
```diff
@@ -84,6 +88,19 @@
...
+  display: flex;
+  flex-grow: 1;
+}
+
+.sidenav > * {
+  min-height: 100%;
+  height: 100%;
+  display: flex;
+  flex-grow: 1;
+}
```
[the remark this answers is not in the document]
michelinewu: There was a point in the work where I wasn't able to show it full height without shifting the footer but it looks like I resolved that along the way so I've removed this.

app/components-react/sidebar/AppsNav.m.less
michelinewu: Removed

app/components-react/sidebar/NavTools.tsx
michelinewu: This is so I can target the text to `flex-grow` it so that the login arrow shows at the right side of the container. I opted to use an antd element but could use just a div.

app/components-react/sidebar/NavTools.tsx
michelinewu: Without this div, the icon property on an antd menu item won't apply the transformation to translate the popout icon over the menu item icon in the collapsed menu. In fact, if won't apply an initial transformation of any kind on the menu item icon in the collapsed menu without this wrapping div (I experimented with scaling).
badge-div

app/components-react/sidebar/NavTools.tsx
michelinewu: logout-modal-middle
logout-modal-low
If I use an antd modal and I center it, we run into the issue where we can't display components over the editor. But if positioning it below the editor works for you, we can do that.

app/services/side-nav/menu.ts
```diff
@@ -0,0 +1,416 @@
...
+    [ENavName.TopNav]: SideBarTopNavData(),
+    [ENavName.BottomNav]: SideBarBottomNavData(),
+  };
+
+  init() {
+    super.init();
+
+    const loggedIn = this.userService.views.isLoggedIn;
+
+    // TODO: set Date to specific date
```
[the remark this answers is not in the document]
michelinewu: Sounds good!

app/services/side-nav/menu-data.ts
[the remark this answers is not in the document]
gettinToasty: usually when we have a situation like this i'll make a name mapping table something to the effect of
```
const menuTitles(item: EMenuItem) {
  return {
    [EMenuItem.Editor]: $t('Editor'),
    ...
  }[EMenuItem]
}
```

app/components-react/sidebar/NavTools.tsx
gettinToasty: so because our layout is modular we can't reliable assume that the editor is going to be in a specific position, but what we *can* do is use `updateStyleBlockers` like you did with the animation to ensure that while the modal is active we don't show the display

app/components-react/sidebar/NavTools.tsx
gettinToasty: gotcha-- yea i think if the purpose is just to hold a css class it's probably better to keep things simple and just use a `span` or `div`, since we don't necessarily know what antd components are doing under the hood and could lead to unexpected or unperformant behavior

i think a good rule of thumb is if you're using an antd component you should have a present and clear reason to do so, like the `Menu` component for example

app/services/windows.ts
```diff
@@ -147,6 +148,7 @@ export function getComponents() {
     WidgetWindow,
     CustomCodeWindow,
     SourceShowcase,
+    NewBadge,
```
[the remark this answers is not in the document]
michelinewu: Ah, yes that is incorrect. It's from when I was trying to figure out how to import the react component into the vue component. Removed

app/components-react/sidebar/NavTools.tsx
michelinewu: Thanks! Will refactor

app/services/side-nav/menu-data.ts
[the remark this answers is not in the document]
michelinewu: Awesome, thanks!

app/components-react/sidebar/SideNav.tsx
michelinewu: Closing because I refactored to use the menu title map

app/components-react/sidebar/SideNav.m.less
michelinewu: Refactored

app/components-react/sidebar/SideNav.m.less
michelinewu: Refactored

app/components-react/sidebar/SideNav.tsx
```diff
@@ -1,255 +1,412 @@
...
-      icon: 'fas fa-exclamation-triangle',
-      title: 'Theme Audit',
-      trackingTarget: 'themeaudit',
+  const resizeObserver = new ResizeObserver((entries: ResizeObserverEntry[]) => {
+    entries.forEach((entry: ResizeObserverEntry) => {
+      const width = Math.floor(entry.contentRect.width);
+      if (width === 50 || width === 200) {
+        updateStyleBlockers('main', false);
+      } else if (!hideStyleBlockers && (width > 50 || width < 200)) {
+        updateStyleBlockers('main', true);
```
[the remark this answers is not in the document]
michelinewu: Thanks! I didn't realize this. Logic added to only set once when animation starts and once when it ends.

app/components-react/sidebar/SideNav.tsx
```diff
@@ -1,255 +1,412 @@
...
-  const [showTabDropdown, setShowTabDropdown] = useState(false);
-  const { LayoutService } = Services;
-  const { currentTab, tabs } = useVuex(() => ({
-    currentTab: LayoutService.state.currentTab,
-    tabs: LayoutService.state.tabs,
-  }));
+  const menuItems = menu.menuItems;
 
   const studioTabs = Object.keys(tabs).map((tab, i) => ({
+    key: i === 0 ? EMenuItem.Editor : `editor-layout-${i}`,
```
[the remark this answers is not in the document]
michelinewu: Thanks! Refactored.

app/components-react/sidebar/NavTools.tsx
michelinewu: Thanks! A doozy but done!

app/components-react/sidebar/SideNav.m.less
gettinToasty: is there a reason we need to use `body` here? makes me a little nervous instead of just using the child's selector itself, since the structure of the page might change in the future

app/components-react/sidebar/SideNav.tsx
```diff
@@ -1,255 +1,465 @@
...
+                // otherwise, show a menu item or a menu item with a submenu
+                return menuItem.hasOwnProperty('subMenuItems') ? (
+                  <SubMenu
+                    key={menuItem.key}
+                    title={menuTitles(menuItem.title)}
+                    icon={
+                      menuItem?.icon && (
+                        // <div>
+                        <i className={menuItem.icon} />
+                        // </div>
```
gettinToasty, line 330: i assume this is just leftovers

app/services/side-nav/menu-data.ts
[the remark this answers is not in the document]
gettinToasty: it looks like these titles need to be changed to `menuTitles(EMenuItem.Editor)`, i'm also a little confused about why there is both `EMenuItem` and `EMenuItemKey`, when they seem to have somewhat identical functions as enums

app/services/side-nav/menu.ts
```diff
@@ -0,0 +1,318 @@
...
+      if (i === index) return newApp;
+
+      if (!app || app?.id === newApp.id) {
+        // if the new app is already in the array, remove it
+        return null;
+      }
+
+      return app;
+    });
+    this.state.apps = updatedApps;
```
gettinToasty, line 295: this doesn't seem to handle the case of an existing app at index `i` being moved to where the target app used to be, unless that's intended

app/components-react/sidebar/SideNav.tsx
michelinewu: Yes, it is. Removed.

app/services/side-nav/menu.ts
michelinewu: Yes, after speaking further we Ashley about it, we decided swapping the two apps may be confusing. So the intended behavior is if the app is already slotted in a different index, we remove it from that index and add it in the new index.

andycreeth (commented):
Nice work! 🎉

app/services/side-nav/menu.ts
michelinewu: Done.

app/components-react/sidebar/SideNav.tsx
```diff
@@ -1,255 +1,465 @@
...
     if (!UserService.views.isLoggedIn && page !== 'Studio') return;
 
     if (trackingTarget) {
-      UsageStatisticsService.actions.recordClick('SideNav', trackingTarget);
+      UsageStatisticsService.actions.recordClick('SideNav2', trackingTarget);
+    }
+
+    if (type) {
+      // eslint-disable-next-line object-shorthand
+      NavigationService.actions.navigate(page, { type: type });
```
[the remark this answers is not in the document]
michelinewu: Ah, that's left over from when I was debugging navigating within the theme tabs. Removed.

app/components-react/pages/PlatformAppStore.tsx
```diff
@@ -5,7 +5,9 @@ import { GuestApiHandler } from 'util/guest-api-handler';
 import * as remote from '@electron/remote';
 import { Services } from 'components-react/service-provider';
 
-export default function PlatformAppStore(p: { params: { appId?: string } }) {
+export default function PlatformAppStore(p: {
+  params: { appId?: string | undefined; type?: string | undefined };
+}) {
```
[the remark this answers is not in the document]
michelinewu: Removed.

app/services/side-nav/menu-data.ts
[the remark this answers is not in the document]
gettinToasty: i'm still wondering why we have both `EMenuItem` and `EMenuItemKey`

app/components-react/sidebar/NavTools.tsx
```diff
@@ -91,64 +102,179 @@ export default function SideNav() {
...
+            );
+          }
+        })}
+      </Menu>
+      <LogoutModal
+        showModal={showModal}
+        handleAuth={handleAuth}
+        handleShowModal={handleShowModal}
+      />
+    </>
```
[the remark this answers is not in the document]
gettinToasty: agree with Andy here, this is going to be tough to update/debug in it's current state

gettinToasty (changes requested):

app/services/user/index.ts
```diff
@@ -893,6 +907,16 @@ export class UserService extends PersistentStatefulService<IUserServiceState> {
...
+    if (!this.sideNavService.views.hasLegacyMenu && !this.views.auth.hasRelogged) {
+      this.sideNavService.setNewUserLogin();
+    } else if (
+      this.sideNavService.views.hasLegacyMenu &&
+      this.dismissablesService.views.shouldShow(EDismissable.NewSideNav)
+    ) {
+      this.sideNavService.setLegacyView();
+    }
+
+    this.dismissablesService.dismiss(EDismissable.LoginPrompt);
```
gettinToasty, line 919: just caught this but i think this functionality should live in the SidenavService and subscribe to the login `Subject`. we don't really want side-effects living in this function or else it'll get really clogged up

michelinewu: Ok sounds good

gettinToasty (approved):

PR #4344, opened 2022-10-28
Fix to center error notifications in bar.

michelinewu:


app/components-react/root/NotificationsArea.m.less
```diff
@@ -32,12 +41,10 @@
   white-space: nowrap;
   overflow: hidden;
   text-overflow: ellipsis;
-  position: absolute;
 
   :global(.ant-message-notice-content) {
     padding: 0;
-    box-shadow: none;
-    background: transparent;
```
gettinToasty, line 40: is there a reason you got rid of these? makes things look kind of wonky

app/components-react/root/NotificationsArea.m.less
```diff
@@ -69,6 +75,9 @@
   cursor: pointer;
   white-space: nowrap;
   margin-right: 10px;
+  height: 100%;
+  display: flex;
+  align-items: center;
```
gettinToasty, line 80: this looks kinda weird will small notifications (see screenshot)
image

gettinToasty (changes requested):

gettinToasty (approved):

gettinToasty (approved):

PR #4345, opened 2022-10-28
Center main page transition loader.

michelinewu:


gettinToasty (changes requested):
the spinner looks extremely small when navigating to a BrowserView with these changes for some reason

michelinewu:
Should it match the size of the loader when the app is starting?
large-loader

michelinewu:
Preview: large-loader

gettinToasty (approved):

PR #4363, opened 2022-11-10
Multistream settings

michelinewu:


gettinToasty (approved):

app/components-react/windows/settings/Multistreaming.tsx
```diff
@@ -42,7 +42,17 @@ export function MultistreamingSettings() {
...
+                  message={
+                    'Step 2: Ensure the \"Confirm stream title and game before going live\" option is checked in the <general>General</general> settings tab."'
+                  }
+                >
+                  <u slot="general" />
+                </Translate>
+              </li>
+              <li>
+                {/* eslint-disable-next-line no-useless-escape */}
+                {$t('Step 3: Select which platforms you are streaming to when you hit \"Go Live\".')}
```
gettinToasty, line 55: i don't think you need to escape this even if the translation is escaped, not a blocker though

michelinewu: Ok, good to know for the future. I think my linter is a little hyperactive so I'll check my settings.

PR #4371, opened 2022-11-21
Add ultra icons

michelinewu:


app/components-react/shared/UltraIcon.tsx
```diff
@@ -15,6 +16,10 @@ interface IUltraIcon {
  * when displaying the icon with multicolor styling.
  */
 export default function UltraIcon({ type, className, style }: IUltraIcon) {
+  const { isDarkTheme } = useVuex(() => ({
+    isDarkTheme: Services.CustomizationService.isDarkTheme,
```
gettinToasty, line 20: i think what you'll want here is `CustomizationService.views.isDarkTheme` to ensure reactivity, otherwise seems good though makes sense for the component to do this parsing itself

gettinToasty (changes requested):

gettinToasty (approved):

PR #4416, opened 2022-12-21
Update css that prohibited dragging on title bar area during load

dec0y:
Ticket details 
https://app.asana.com/0/734380881425048/1202684183731459/f


gettinToasty (approved):
fix looks good, we might want to stagger that css so we don't need to use the `!important` tag though, something like:
```
.main-loading.initial-loading {
  top: 0px;
}
```

going to try rerunning the failed test since it doesn't look like it's tied to this change

PR #4419, opened 2023-01-03
Resolve spin wheel button not showing up on mini feed when widget is added to scene

dec0y:


dec0y:
https://app.asana.com/0/734380881425048/1202485348976624/f

michelinewu (approved):

gettinToasty (changes requested):
this fix is good if it works, but just to comply with modern code patterns i think it'd be nice if we can move it to `WidgetServiceViews`: generally speaking that's where we want to put things that read our state going forward

dec0y:
I can totally make that change and get that moved over to the new service. Totally makes sense if we're using that moving forward.

gettinToasty (approved):

PR #4420, opened 2023-01-04
Add missing YT test events on the widget tester

dec0y:
https://app.asana.com/0/734380881425048/1202134938984128/f

### DIFF
--- a/app/app.g.less
+++ b/app/app.g.less
@@ -1,17 +1,17 @@
-@import './styles/foundation';
-@import '../node_modules/overlayscrollbars/css/OverlayScrollbars.css';
-@import '../node_modules/vue-multiselect/dist/vue-multiselect.min.css';
-@import './styles/index';
-@import './styles/checkboxes';
-@import './styles/tables';
-@import './styles/buttons';
-@import './styles/classes';
-@import './styles/custom-icons';
-@import './styles/inputs';
-@import './styles/sl-vue-tree';
-@import './styles/tooltips';
-@import './styles/animations';
-@import './styles/loader.less';
+@import "./styles/foundation";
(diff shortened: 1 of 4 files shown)


app/app.g.less
```diff
@@ -240,7 +240,7 @@ a,
   right: calc(~'100% + 20px');
   align-self: center;
   display: flex;
-  z-index: 10;
+  z-index: 1011;
```
gettinToasty, line 243: hmmm i don't know if we should be doing this on a global class like this, if there isn't a locally scoped class for that element we can just add it to the style itself. also looks like your linter is disagreeing with the team codestyles here for some reason 🤔

gettinToasty (changes requested):

app/components/windows/WidgetEditor.vue
```diff
@@ -17,7 +17,7 @@
           </div>
         </div>
         <div v-if="apiSettings.testers" class="button button--action test-button">
-          <test-widgets :componentProps="{ testers: apiSettings.testers }" />
+          <test-widgets :componentProps="{ testers: null }" />
```
gettinToasty, line 20: i think we probably just want to remove this prop or make it optional rather than passing in null right?

app/app.g.less
```diff
@@ -477,9 +478,15 @@ select {
...
-.checkboxdone-enter-to {width: 16px;}
+.checkboxdone-enter-active {
+  transition: width 0.3s;
+}
+.checkboxdone-enter {
+  width: 0;
+}
+.checkboxdone-enter-to {
+  width: 16px;
+}
```
gettinToasty, line 489: can you just undo the diffs in this file? afaik there isn't anything actually materially changing in it anymore

gettinToasty (changes requested):

PR #4422, opened 2023-01-05
use vcam new path on mac

EddyGharbi:


gettinToasty (changes requested):
this will probably break virtual cam on windows right? do we not want to handle both cases?

andycreeth:
This should be OK as plists are exclusively a mac thing.

gettinToasty (approved):

PR #4449, opened 2023-01-30
Port ChatBox Widget to React

gettinToasty:


test/regular/widgets/chat-box.ts
```diff
@@ -1,21 +1,20 @@
...
+import { assertFormContains, fillForm } from '../../helpers/modules/forms';
 import { waitForWidgetSettingsSync } from '../../helpers/widget-helpers';
+import { clickButton, focusChild, focusMain, waitForDisplayed } from '../../helpers/modules/core';
+import { sleep } from '../../helpers/sleep';
 
 useSpectron();
 
-test('Chatbox Visual Settings', async t => {
+// TODO: Fix test for react
+test.skip('Chatbox Settings', async t => {
```
michelinewu, line 12: Is there a particular reason we're skipping this test for now? Is it the API integration changes?

gettinToasty: basically the test framework for react forms doesn't appear to be completely built out yet and only seems to handle a few specific cases, but that was all work Alex was doing

michelinewu: Ah ok, thanks for the clarification

michelinewu (approved):

PR #4451, opened 2023-02-02
Sidenav fixes

michelinewu:


app/services/protocol-links.ts
```diff
@@ -108,9 +116,11 @@ export class ProtocolLinksService extends Service {
     const appId = info.path.replace('/', '');
 
     if (this.platformAppsService.views.getApp(appId)) {
-      this.navigationService.navigateApp(appId);
+      this.navigationService.navigate('PlatformAppMainPage', { appId });
+      this.sideNavService.setCurrentMenuItem(appId);
```
gettinToasty, line 120: i might be misunderstanding but i thought `navigateApp` did these two things?

gettinToasty (approved):
non-blocking question but otherwise LGTM

app/services/protocol-links.ts
michelinewu: Ah yes, it does. I was thinking about that optional key when setting the menu item at the time :sweat_smile:

PR #4462, opened 2023-02-13
Check if user is logged in to access the theme library.

michelinewu:


app/components-react/windows/ManageSceneCollections.tsx
```diff
@@ -57,8 +68,19 @@ export default function ManageSceneCollections() {
...
-    NavigationService.actions.navigate('BrowseOverlays');
-    WindowsService.actions.closeChildWindow();
+    if (isLoggedIn) {
+      NavigationService.actions.navigate('BrowseOverlays');
+      WindowsService.actions.closeChildWindow();
+    }
+  }
+
+  function upgradeToPrime() {
+    UsageStatisticsService.actions.recordClick('SideNav2', 'prime');
```
gettinToasty, line 78: i think we want to change this from `SideNav2` to something else here right?

gettinToasty (changes requested):

gettinToasty (approved):

PR #4478, opened 2023-02-27
Refactor consumer layers to be based on source height

gettinToasty:


app/services/guest-cam/guest-track.ts
```diff
@@ -58,18 +58,19 @@ export class GuestTrack extends MediasoupEntity {
     if (this.opts.kind !== 'video') return;
     if (!this.consumerId) return;
 
-    const totalStreams = this.guestCamService.consumer.guests.length;
+    const height = this.guestCamService.views.getSourceForGuest(this.opts.streamId).height;
```
andycreeth, line 61: This may not be what you actually want.  This is the size of the actual underlying video feed, not the size of the actual scene item in the scene.

andycreeth (changes requested):

app/services/guest-cam/guest-track.ts
gettinToasty: oh got it, is there a way to access that?

andycreeth: It will be kind of a complex operation.  You'll need to find all the scene items referencing a particular source, find the one with the largest scale, and then multiply that scale by the original heigh of the source.

PR #4497, opened 2023-03-10
Streaming refactor for backend changes

michelinewu:


app/services/settings/streaming/stream-settings.ts
```diff
@@ -143,12 +147,15 @@ export class StreamSettingsService extends PersistentStatefulService<IStreamSett
     let streamFormData = cloneDeep(this.views.obsStreamSettings);
 
     streamFormData.forEach(subCategory => {
+      console.log('subCategory ', subCategory);
```
gettinToasty, line 150: console log

app/services/streaming/streaming-view.ts
```diff
@@ -156,6 +156,22 @@ export class StreamInfoView<T extends Object> extends ViewHandler<T> {
     return this.streamingState.streamingStatus !== 'offline';
   }
 
+  get isGreen(): boolean {
+    return Services.VideoSettingsService.isGreen;
```
gettinToasty, line 160: this isn't reactive

app/services/streaming/streaming.ts
```diff
@@ -372,6 +421,21 @@ export class StreamingService
...
 
+  private handleSetupPlatformError(e: unknown, platform: TPlatform) {
+    console.error('Error running beforeGoLive for platform', e);
+    // cast all PLATFORM_REQUEST_FAILED errors to SETTINGS_UPDATE_FAILED
+    if (e instanceof StreamError) {
+      e.type =
+        (e.type as TStreamErrorType) === 'PLATFORM_REQUEST_FAILED'
+          ? 'SETTINGS_UPDATE_FAILED'
+          : e.type || 'UNKNOWN_ERROR';
+      this.setError(e, platform);
```
gettinToasty, line 432: why are we doing this?

app/services/streaming/streaming.ts
```diff
@@ -614,7 +678,25 @@ export class StreamingService
...
 
-    obs.NodeObs.OBS_service_startStreaming();
+    if (this.views.isGreen) {
+      if (this.views.enabledPlatforms.length > 1) {
+        this.views.contextsToStream.forEach(async (contextName: string) => {
+          const context = this.videoSettingsService.contexts[contextName];
+          obs.NodeObs.OBS_service_setVideoInfo(context, contextName);
+          obs.NodeObs.OBS_service_startStreaming(contextName);
+
+          await new Promise(resolve => setTimeout(resolve, 1000));
```
gettinToasty, line 688: what's the purpose of this?

app/services/streaming/streaming.ts
```diff
@@ -685,7 +767,23 @@ export class StreamingService
         remote.powerSaveBlocker.stop(this.powerSaveId);
       }
 
-      obs.NodeObs.OBS_service_stopStreaming(false);
+      // obs.NodeObs.OBS_service_stopStreaming(false);
```
gettinToasty, line 770: commented out code

app/services/streaming/streaming.ts
```diff
@@ -707,7 +805,23 @@ export class StreamingService
...
+        } else {
+          const platform = this.views.enabledPlatforms[0];
+          const display = this.views.getPlatformDisplay(platform);
+          obs.NodeObs.OBS_service_stopStreaming(true, display);
+        }
+      } else {
+        obs.NodeObs.OBS_service_stopStreaming(true);
+      }
+
+      // obs.NodeObs.OBS_service_stopStreaming(true);
```
gettinToasty, line 824: commented out code

app/services/streaming/streaming.ts
```diff
@@ -707,7 +805,23 @@ export class StreamingService
     }
 
     if (this.state.streamingStatus === EStreamingState.Ending) {
-      obs.NodeObs.OBS_service_stopStreaming(true);
+      if (this.views.isGreen) {
+        if (this.views.enabledPlatforms.length > 1) {
+          this.views.contextsToStream.forEach(async (contextName: string) => {
```
gettinToasty, line 810: shouldnt this be the same regardless of if `enabledPlatforms` is one or more? seems like a bit of a superfluous if step here (and also in the previous instance of this)

app/services/green/green-data.ts
```diff
@@ -0,0 +1,62 @@
+import { $t } from 'services/i18n';
+import { TPlatform } from 'services/platforms';
+
+export enum EGreenPlatform {
+  Twitch = 'twitch',
+  Facebook = 'facebook',
+  YouTube = 'youtube',
+  Trovo = 'trovo',
+}
```
gettinToasty, line 9: i'm not sure this is 100% needed since we have a type `TPlatform`, but either way if we do want an enum for platforms it should probably go there

app/services/green/green.ts
```diff
@@ -0,0 +1,441 @@
...
+    //   if (this.videoSettingsService.contexts.green) {
+    //     this.assignSceneNodes(scene.id);
+    //   }
+    // });
+
+    // this.scenesService.sceneSwitched.subscribe((scene: IScene) => {
+    //   if (this.scenesService.views.getSceneItemsBySceneId(scene.id).length > 0) {
+    //     this.confirmOrCreateGreenNodes(scene.id);
+    //   }
+    // });
```
gettinToasty, line 211: i assume this is all meant to be commented for now?

gettinToasty (changes requested):

app/services/green/green-data.ts
michelinewu: Moved. I'd like to use the enum so that I don't have to do any trickery trying to get the strings from `TPlatform`

app/services/streaming/streaming.ts
michelinewu: This is to guarantee that the Green code is skipped for now since it will always return false. The logic inside this if statement what will actually handle the Green functionality. I can also just remove this block.

app/services/streaming/streaming.ts
michelinewu: Removed

app/services/streaming/streaming.ts
michelinewu: Removed

app/services/streaming/streaming.ts
michelinewu: This is not necessary, just a failsafe to throttle the calls and prevent possible errors.

app/services/streaming/streaming.ts
michelinewu: This is a copy/paste from the previous code, which I separated out into its own function to make it dry when I call it in the Green code. The error types determine which error message is shown to the user.

app/services/streaming/streaming-view.ts
michelinewu: This is intentional for now so that it always returns false and skips over the Green code.

app/services/settings/streaming/stream-settings.ts
michelinewu: Removed

app/services/settings-v2/video.ts
```diff
@@ -5,78 +5,230 @@ import { mutation, StatefulService } from '../core/stateful-service';
...
 
-    this.SET_VIDEO_CONTEXT();
+    this.state[contextName] = obs.VideoFactory.create();
+    this.state[display] = {} as obs.IVideoInfo;
+    this.migrateSettings(display);
+    this.state[contextName].video = this.state[display];
 
-    this.migrateSettings();
-    obs.VideoFactory.videoContext = this.state.videoContext;
+    return !!this.state[contextName];
```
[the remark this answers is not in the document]
michelinewu: For both `establishVideoContext` and `destroyVideoContext`, this was to check if the contexts were in fact created/destroyed to catch errors. This may not be the ideal way to do this so please let me know if you have any thoughts about implementing error handling for this.

app/services/settings-v2/video.ts
```diff
@@ -5,78 +5,230 @@ import { mutation, StatefulService } from '../core/stateful-service';
 import * as obs from '../../../obs-api';
 import { SettingsManagerService } from 'services/settings-manager';
 
+/**
+ * Display Types
+ *
+ * Add display type options by adding the display name to the displays array
+ * and the context name to the context name map.
+ */
+const displays = ['default', 'horizontal', 'green'] as const;
```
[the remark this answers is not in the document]
michelinewu: I didn't want to remove the default property just yet in production to prevent anything possible referencing it and breaking. But I can go ahead and do that.

app/services/settings-v2/video.ts
```diff
@@ -5,78 +5,230 @@ import { mutation, StatefulService } from '../core/stateful-service';
...
+    return this.settingsManagerService.views.videoSettings;
   }
 
-  migrateSettings() {
-    Object.keys(this.videoSettings).forEach((key: keyof obs.IVideo) => {
-      this.SET_VIDEO_SETTING(key, this.videoSettings[key]);
-    });
+  getBaseResolution(display: TDisplayType = 'horizontal') {
+    return `${this.state[display].baseWidth}x${this.state[display].baseHeight}`;
+  }
```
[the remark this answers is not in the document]
michelinewu: Leftover from experimenting with moving the function. Removed

app/services/settings-v2/video.ts
```diff
@@ -5,78 +5,230 @@ import { mutation, StatefulService } from '../core/stateful-service';
...
+      this.establishVideoContext('green');
+    }
+  }
+
+  get contexts() {
+    return {
+      default: this.state.horizontalContext,
+      horizontal: this.state.horizontalContext,
+      green: this.state.greenContext,
+    };
```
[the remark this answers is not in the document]
michelinewu: Let's connect about the ideal way to do this

app/components-react/shared/Display.tsx
```diff
@@ -28,7 +29,7 @@ export default function Display(props: DisplayProps) {
   const v = useVuex(
     () => ({
       paddingColor: CustomizationService.views.displayBackground,
-      baseResolution: VideoService.baseResolution,
+      baseResolution: VideoSettingsService.actions.getBaseResolution(p.type),
```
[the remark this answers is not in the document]
michelinewu: Yes, thanks!

app/components/windows/settings/Settings.vue.ts
```diff
@@ -181,7 +181,13 @@ export default class Settings extends Vue {
   }
 
   get categoryNames() {
-    return this.settingsService.getCategories();
+    return this.settingsService
+      .getCategories()
+      .filter(
+        category =>
+          !category.toLowerCase().startsWith('stream', 0) ||
+          (category.toLowerCase().startsWith('stream', 0) && category.length < 7),
```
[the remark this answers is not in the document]
michelinewu: Done

app/services/settings-manager.ts
```diff
@@ -1,16 +1,136 @@
...
+
+    this.state.videoSettings.defaultDisplay = display;
+  }
+
+  @mutation()
+  private SET_VIDEO_SETTING(setting: Partial<IVideoSetting>, display: TDisplayType = 'green') {
+    this.state.videoSettings.activeDisplays = {
+      ...this.state.videoSettings.activeDisplays,
+      [display]: setting,
+    };
```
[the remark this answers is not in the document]
michelinewu: Moved

app/services/settings-v2/video.ts
```diff
@@ -5,78 +5,230 @@ import { mutation, StatefulService } from '../core/stateful-service';
...
+  default: obs.IVideoInfo;
+  horizontal: obs.IVideoInfo;
+  green: obs.IVideoInfo;
+  // outputIdMap: { [key: string]: number } // @@@ TODO: map outputId to context when created. Key is context name
+}
+
 @InitAfter('UserService')
-export class VideoSettingsService extends StatefulService<{ videoContext: obs.IVideo }> {
+@InitAfter('SettingsManagerService')
+@InitAfter('SceneCollectionsService')
```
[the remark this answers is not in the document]
michelinewu: Let's connect on this to clarify and go over best practice

app/services/settings-v2/video.ts
```diff
@@ -3,80 +3,201 @@ import { Inject } from 'services/core/injector';
...
+      baseRes: `${context.video.baseWidth}x${context.video.baseHeight}`,
+      outputRes: `${context.video.outputWidth}x${context.video.outputHeight}`,
+      scaleType: context.video.scaleType,
+      fpsType: context.video.fpsType,
+      fpsCom: `${context.video.fpsNum}-${context.video.fpsDen}`,
+      fpsNum: context.video.fpsNum,
+      fpsDen: context.video.fpsDen,
+      fpsInt: context.video.fpsNum,
+    };
+  }
```
gettinToasty, line 62: i think for this one and `get baseResolution` we actually do want to reference state since these values are displayed to the user

app/services/settings-v2/video.ts
```diff
@@ -3,80 +3,201 @@ import { Inject } from 'services/core/injector';
...
+  migrateSettings(display?: TDisplayType) {
+    if (display === 'default' || display === 'horizontal') {
+      this.state.horizontal = this.contexts.horizontal.video;
+      Object.keys(this.contexts.horizontal.legacySettings).forEach(
+        (key: keyof obs.IAdvancedStreaming | keyof obs.ISimpleStreaming) => {
+          this.SET_VIDEO_SETTING(key, this.contexts.horizontal.legacySettings[key]);
+        },
+      );
+      Object.keys(this.contexts.horizontal.video).forEach((key: keyof obs.IVideo) => {
+        this.SET_VIDEO_SETTING(key, this.contexts.horizontal.video[key]);
```
gettinToasty, line 89: i'm a little confused about why we're doing two loops here-- aren't the keys for `legacySettings` and `video` identical?

app/services/settings-v2/video.ts
```diff
@@ -3,80 +3,201 @@ import { Inject } from 'services/core/injector';
...
+          this.createDefaultFps();
+        }
+      });
+    } else {
+      const data = this.videoSettings.green;
 
-    if (invalidFps(this.state.videoContext.fpsNum, this.state.videoContext.fpsDen)) {
-      this.setVideoSetting('fpsNum', 30);
-      this.setVideoSetting('fpsDen', 1);
+      this.state[display] = this.contexts[display].video;
```
gettinToasty, line 99: we shouldn't set state outside of mutations

app/services/settings-v2/video.ts
```diff
@@ -3,80 +3,201 @@ import { Inject } from 'services/core/injector';
...
 
-  establishVideoContext() {
-    if (this.state.videoContext) return;
+  establishVideoContext(display: TDisplayType = 'horizontal') {
+    if (this.contexts[display]) return;
+
+    this.contexts[display] = obs.VideoFactory.create();
+    this.state[display] = {} as obs.IVideoInfo;
+    this.migrateSettings(display);
+    this.contexts[display].video = this.state[display];
```
gettinToasty, line 119: i don't think this is necessary, since curing the migration we already set these values to mirror eachother

app/services/settings-v2/video.ts
```diff
@@ -3,80 +3,201 @@ import { Inject } from 'services/core/injector';
...
+      }
+    });
   }
 
-  setVideoSetting(key: string, value: unknown) {
-    this.SET_VIDEO_SETTING(key, value);
-    this.updateObsSettings();
+  @mutation()
+  DESTROY_VIDEO_CONTEXT(display: TDisplayType = 'horizontal') {
+    this.contexts[display] = null as obs.IVideo;
```
gettinToasty, line 187: usually we don't want side effects in out mutations, so this should probably just exist in `destroyVideoContext`

app/services/settings-v2/video.ts
```diff
@@ -3,80 +3,201 @@ import { Inject } from 'services/core/injector';
...
+    this.state[display] = null as obs.IVideoInfo;
   }
 
   @mutation()
   SET_VIDEO_CONTEXT() {
-    this.state.videoContext = {} as obs.IVideo;
+    this.contexts.horizontal = {} as obs.IVideo;
```
gettinToasty, line 193: we no longer need this mutation

app/services/green/green.ts
```diff
@@ -0,0 +1,544 @@
...
+
+  getPlatformContextName(platform: TPlatform): TOutputOrientation {
+    return this.getPlatformDisplay(platform) === 'horizontal' ? 'landscape' : 'portrait';
+  }
+}
+
+@InitAfter('UserService')
+@InitAfter('ScenesService')
+@InitAfter('SceneCollectionsService')
+@InitAfter('VideoSettingsService')
```
gettinToasty, line 190: multiple InitAfter's again

app/services/settings-manager.ts
```diff
@@ -1,16 +1,18 @@
...
+// from the backend.
+// */
 
 class SettingsManagerViews extends ViewHandler<{}> {}
 
 export class SettingsManagerService extends Service {
+  @Inject() private greenService: GreenService;
   get videoSettings() {
-    return obs.VideoFactory.legacySettings;
+    return this.greenService.views.videoSettings;
```
gettinToasty, line 16: as far as i can understand, this will reset all of our user's settings to default on application load-- is that not the case?

gettinToasty (changes requested):

app/services/green/green.ts
michelinewu: Removed

app/services/settings-manager.ts
michelinewu: Ah yes. This actually isn't currently being reference anywhere, so I've removed it.

app/services/settings-v2/video.ts
michelinewu: Done

app/services/settings-v2/video.ts
michelinewu: Done

app/services/settings-v2/video.ts
michelinewu: Agreed. Removed

app/services/settings-v2/video.ts
michelinewu: Done

app/services/settings-v2/video.ts
michelinewu: Done

gettinToasty (approved):

app/services/green/green.ts
```diff
@@ -0,0 +1,514 @@
...
+    return this.state.platformSettings;
+  }
+
+  get platformSettingsList(): IGreenPlatformSetting[] {
+    return Object.values(this.state.platformSettings);
+  }
+
+  get hasGreenScenes() {
+    if (!this.state.nodeMaps) return false;
+    for (const display in this.state.nodeMaps) {
```
andycreeth, line 60: This can be written pretty cleanly using `Array.some()`:

```TypeScript
return this.state.nodeMaps.some(display => {
  return this.state.nodeMaps.hasOwnProperty(display);
});
```

app/services/green/green.ts
```diff
@@ -0,0 +1,514 @@
...
+  }
+
+  get greenNodeIds(): string[] {
+    const activeSceneId = this.scenesService.views.activeSceneId;
+
+    if (!this.hasGreenNodes) return;
+
+    return Object.entries(this.state.sceneNodeMaps[activeSceneId]).map(
+      ([key, value]: [string, string]) => value,
+    );
```
andycreeth, line 83: Seems like this could just be `Object.values(this.state.sceneNodeMaps[activeSceneId])`?

app/services/green/green.ts
```diff
@@ -0,0 +1,514 @@
...
+      this.state.nodeMaps['green'][sceneItemId],
+    ];
+  }
+
+  getPlatformContextName(platform: TPlatform): TOutputOrientation {
+    return this.getPlatformDisplay(platform) === 'horizontal' ? 'landscape' : 'portrait';
+  }
+}
+
+export class GreenService extends PersistentStatefulService<IGreenServiceState> {
```
andycreeth, line 187: Would be good to chat about what is being stored in persistent state here.  If it's storing stuff like IDs to scene items, then we should be persisting it in the Scene Collections system rather than with `PersistentStatefulService`

app/components-react/shared/Display.tsx
```diff
@@ -28,7 +29,9 @@ export default function Display(props: DisplayProps) {
   const v = useVuex(
     () => ({
       paddingColor: CustomizationService.views.displayBackground,
-      baseResolution: VideoService.baseResolution,
+      baseResolution: `${VideoSettingsService.contexts[p.type ?? 'horizontal']?.video.baseWidth}x${
+        VideoSettingsService.contexts[p.type ?? 'horizontal']?.video.baseHeight
+      }`,
```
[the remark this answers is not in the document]
michelinewu: Done

app/components/windows/settings/Settings.vue.ts
```diff
@@ -181,7 +181,13 @@ export default class Settings extends Vue {
   }
 
   get categoryNames() {
-    return this.settingsService.getCategories();
+    return this.settingsService
+      .getCategories()
+      .filter(
+        category =>
+          !category.toLowerCase().startsWith('stream') ||
+          (category.toLowerCase().startsWith('stream') && category.length < 7),
```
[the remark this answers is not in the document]
michelinewu: The stream settings for the second context need to be filtered out of this list. To make it less brittle, I opted to filter out any additional stream settings returned rather than the string `StreamSecond`.

app/services/green/green.ts
michelinewu: Sounds good

app/components/windows/settings/Settings.vue.ts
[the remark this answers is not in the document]
andycreeth: Yeah, that makes sense.  I think my only point was that if you're checking for a string of length less than 7 which starts with `stream`, the only thing it could be is `stream`, so you might as well make the check simpler.

app/services/green/green.ts
andycreeth: Oh of course right

app/components/windows/settings/Settings.vue.ts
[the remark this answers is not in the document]
michelinewu: Ohhh gotcha. Changed

app/services/green/green.ts
gettinToasty: could maybe use Object.values(this.state.nodeMaps).some()? but not sure if that'd be cleaner

PR #4518, opened 2023-04-06
fix(youtube): category set as default when scheduling event

blackxored:
Category for events was being set to the default of "Entertainment"
since it was only updated by the "before go-live" logic.

This moves the update to `createBroadcast` and re-orders some
logic to avoid duplicate requests.

Ideally, we would refactor `updateBroadcast` to behave in the
same way with some diffing and remove all `updateCategory` calls
from the codebase.

ref: https://app.asana.com/0/911512908891105/1202062997993099


xoredg:
This is a blind change for now as YouTube still has me rate-limited.

PR #4519, opened 2023-04-06
fix(facebook): event doesn't appear on list after created until reload

xoredg:
Newly-created `LiveVideo`'s do not seem to have `broadcast_start_time`,
which causes one of the many conversion logics around them to fail and
results in an invalid date, preventing the newly-created event from
showing on the list until the view is reloaded and data is fetched and
converted again.

NOTE: there's still a bug on the FB scheduler, clicking on any event does not
show the Edit modal, it fails to fetch the proper video/event
information (they're a mix at this point), revolving around
`fetchVideo` and `fetchStartStreamOptionsForVideo`.

### DIFF
--- a/app/services/platforms/facebook.ts
+++ b/app/services/platforms/facebook.ts
@@ -44,11 +44,14 @@ export interface IFacebookLiveVideo {
   id: string;
   stream_url: string;
   title: string;
+  /** @deprecated: doesn't exist on their API docs  */
   game: string;
   description: string;
   permalink_url: string;
   video: { id: string };
   broadcast_start_time: string;
+  planned_start_time?: string;
+  /** custom fields we've created */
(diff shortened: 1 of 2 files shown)


app/services/platforms/facebook.ts
```diff
@@ -44,11 +44,14 @@ export interface IFacebookLiveVideo {
...
   stream_url: string;
   title: string;
+  /** @deprecated: doesn't exist on their API docs  */
   game: string;
   description: string;
   permalink_url: string;
   video: { id: string };
   broadcast_start_time: string;
+  planned_start_time?: string;
+  /** custom fields we've created */
```
gettinToasty, line 54: non-blocking comment but `event_params` isn't actually a custom field, it's what is required to send to the new events api: https://developers.facebook.com/docs/live-video-api/guides/scheduling/

PR #4532, opened 2023-04-17
refactor(settings): port Hotkey settings to react

xoredg:


app/components-react/windows/settings/Hotkey.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+/**
+ * Represents a binding that has a unique key for CSS animations
+ */
+interface IKeyedBinding {
+  binding: IBinding;
+  key: string;
+}
+
+class HotkeyProps {
```
gettinToasty, line 20: this can probably just be an interface, the class version of props was just for Vue tsx components

app/components-react/windows/settings/Hotkey.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+  keys.push(key);
+
+  return keys.join('+');
+}
+
+export default function Hotkey(props: HotkeyProps) {
+  const { MarkersService } = Services;
+  const { hotkey } = props;
+  const description = hotkey.description!;
```
gettinToasty, line 107: we want to avoid bangs where possible, if a description is required in the component i would probably do something like `const description = hotkey.description || ''`

app/components-react/windows/settings/Hotkey.tsx
```diff
@@ -0,0 +1,216 @@
...
+  useEffect(() => {
+    if (bindings.length === 0) {
+      setBindings([createBindingWithKey(getBlankBinding())]);
+    }
+  }, []);
+
+  // TODO: avoid mutation
+  useEffect(() => {
+    hotkey.bindings = bindings.map(b => b.binding);
+  }, [bindings]);
```
gettinToasty, line 161: i'm not 100% sure this is going to work as expected, as i don't think React will call this when there is a deep change in the bindings array

app/components-react/windows/settings/Hotkey.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+  function HotkeyDescription() {
+    if (hotkey.isMarker) {
+      return (
+        <TextInput
+          className={styles.HotkeyDescription}
```
gettinToasty, line 167: not super important but we usually use camelCase for style names

app/components-react/windows/settings/Hotkey.m.less
```diff
@@ -0,0 +1,27 @@
...
+  width: 200px;
+  padding-right: 10px;
+}
+
+.Hotkey-input {
+  display: inline-block;
+  width: 150px;
+}
+
+.Hotkey-control {
```
gettinToasty, line 19: i assume these are capitalized because the old vue ones were but its not really a pattern we use these days

app/components-react/windows/settings/Hotkey.tsx
```diff
@@ -0,0 +1,216 @@
...
+      setBindings([createBindingWithKey(getBlankBinding())]);
+    }
+  }, []);
+
+  // TODO: avoid mutation
+  useEffect(() => {
+    hotkey.bindings = bindings.map(b => b.binding);
+  }, [bindings]);
+
+  function HotkeyDescription() {
```
gettinToasty, line 163: maybe i'm missing something but i dont think i see this function being called anywhere?

app/components-react/windows/settings/HotkeyGroup.tsx
```diff
@@ -0,0 +1,65 @@
...
+class CommonProps {
+  title: string | null = null;
+}
+
+class HotkeyGroupProps extends CommonProps {
+  hotkeys: IHotkey[] = [];
+  isSearch: boolean = false;
+}
+
+class HeaderProps extends CommonProps {
```
gettinToasty, line 16: these classes can be made into interfaces for the same reason as above :)

app/components-react/windows/settings/HotkeyGroup.tsx
```diff
@@ -0,0 +1,65 @@
...
+    setCollapsed(collapsed => !collapsed);
+  };
+
+  const headerProps = { title, isCollapsible, collapsed, handleClick: toggleCollapsed };
+
+  return (
+    <div className="section">
+      <Header {...headerProps} />
+      {/* TODO: not sure whether this matches Vue's expand */}
+      <Animate transitionName="slidedown">
```
gettinToasty, line 52: i think instead of the Animate component here we can probably just use Antd's collapse component: https://ant.design/components/collapse

gettinToasty: ^ this would probably let us take out a lot of the logic in HotkeyGroup as well

app/components-react/windows/settings/Hotkeys.tsx
```diff
@@ -0,0 +1,153 @@
...
+            {Object.keys(filteredHotkeySet.scenes).map(renderScenesHotkeyGroup)}
+          </>
+        )}
+        {hasSourceHotkeys && (
+          <>
+            <h2>{$t('Sources')}</h2>
+            {Object.keys(filteredHotkeySet.sources).map(renderSourcesHotkeyGroup)}
+          </>
+        )}
+        {/* TODO: 1-1 conversion, not sure if we're supposed to render this title if there aren't any markers */}
```
gettinToasty, line 145: we should put this in the condition as well i think, i overlooked it originally

app/components/windows/settings/Settings.vue
```diff
@@ -82,7 +82,7 @@
           @scanCompleted="onScanCompletedHandler"
           v-slot:default="{ page, scanning }"
         >
-          <hotkeys
+          <Hotkeys
             v-if="page === 'Hotkeys'"
             :globalSearchStr="scanning ? '' : searchStr"
             :highlightSearch="highlightSearch"
```
gettinToasty, line 88: these props should be wrapped in a componentProps object like so:
```
:componentProps="{ globalSearchStr: "scanning ? '' : searchStr", highlightSearch... etc }"
```

gettinToasty (changes requested):

app/components-react/windows/settings/Hotkey.tsx
xoredg: Will change when I get to the style fixes, it was set this way on the Vue version and I thought tests depended on it. But I can't find what I saw anymore, so I'm guessing if tests won't care we can use any class names and scoped too.

app/components-react/windows/settings/Hotkey.tsx
xoredg: There's a few functions that aren't used anymore since I found `HotkeyBinding` existed, will remove.

app/services/hotkeys.ts
```diff
@@ -548,6 +548,7 @@ export class HotkeysService extends StatefulService<IHotkeysServiceState> {
   }
 
   getHotkeysSet(): IHotkeysSet {
+    console.log('getHotkeysSet');
```
gettinToasty, line 551: errant console log

gettinToasty (approved):

PR #4535, opened 2023-04-18
Mw fix preview 8

michelinewu:


app/services/transitions.ts
```diff
@@ -151,6 +153,11 @@ export class TransitionsService extends StatefulService<ITransitionsState> {
     this.sceneCollectionsService.collectionWillSwitch.subscribe(() => {
       this.disableStudioMode();
     });
+
+    const establishedContext = this.videoSettingsService.establishedContext.subscribe(() => {
+      if (!this.studioModeTransition) this.createStudioModeTransition();
+      establishedContext.unsubscribe();
+    });
```
gettinToasty, line 160: i think we could use an explanatory comment here since it's not necessarily immediately understood behavior

gettinToasty (changes requested):

app/services/transitions.ts
michelinewu: Sounds good

app/services/transitions.ts
```diff
@@ -241,9 +248,11 @@ export class TransitionsService extends StatefulService<ITransitionsState> {
     }
   }
 
-  getStudioTransitionName() {
+  get studioTransitionName(): string | undefined {
     if (this.studioModeTransition) {
-      return this.studioModeTransition.name;
+      return this.studioModeTransition?.name;
+    } else {
+      return undefined;
```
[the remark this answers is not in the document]
gettinToasty: got it

PR #4589, opened 2023-06-07
Migrate GLVolmeters to React

holiber:


app/components-react/root/GLVolmeters.tsx
```diff
@@ -0,0 +1,438 @@
...
+  get audioSources() {
+    return this.audioService.views.sourcesForCurrentScene.filter(source => {
+      return !source.mixerHidden && source.isControlledViaObs;
+    });
+  }
+
+  /**
+   * add or remove subscription for volmeters depending on current scene
+   */
+  // @Watch('audioSources')
```
gettinToasty, line 147: is this meant to be commented out?

app/components-react/root/GLVolmeters.tsx
```diff
@@ -0,0 +1,438 @@
...
+    // Make sure all state is cleared out
+    this.gl = null!;
+    this.program = null!;
+    this.positionLocation = null!;
+    this.resolutionLocation = null!;
+    this.translationLocation = null!;
+    this.scaleLocation = null!;
+    this.volumeLocation = null!;
+    this.peakHoldLocation = null!;
+    this.bgMultiplierLocation = null!;
```
gettinToasty, line 231: not sure why all these nulls are being asserted?

app/components-react/root/GLVolmeters.tsx
```diff
@@ -0,0 +1,438 @@
...
+  }
+
+  private setColorUniform(uniform: string, color: number[]) {
+    const location = this.gl.getUniformLocation(this.program, uniform);
+    // eslint-disable-next-line
+    this.gl.uniform3fv(location, color.map(c => c / 255));
+  }
+
+  private setCanvasSize() {
+    const $parent = getDefined(this.$refs.canvas?.parentElement);
```
gettinToasty, line 325: not sure if i understand the interaction of `getDefined` and safe navigation here

Hayderhammami (approved):

app/components-react/root/GLVolmeters.tsx
andycreeth: Looks like a Vue `@Watch` decorator - most likely a copy paste from the Vue class.  Should remove.

app/components-react/root/GLVolmeters.tsx
andycreeth: In theory, using `getDefined` here produces a better stack trace, and, if we see it in sentry it should be prioritized because it means we made an incorrect assumption about whether something could be null.

In practice - I'm not sure we're very diligent about fixing these when we make incorrect assumptions.

From a runtime perspective it doesn't matter - it's runtime exception either way.

andycreeth (commented):
Seems like the old component isn't removed as part of this PR?

Hayderhammami (approved):

gettinToasty (approved):

PR #4599, opened 2023-06-15
Fix Auto-optimizer running with HDR settings

gettinToasty:


app/services/settings/settings.ts
```diff
@@ -169,6 +169,12 @@ class SettingsViews extends ViewHandler<ISettingsServiceState> {
   get advancedAudioSettings() {
     return this.state.Advanced.formData.find(data => data.nameSubCategory === 'Audio');
   }
+
+  get hasHDRSettings() {
+    const advVideo = this.state.Advanced.formData.find(data => data.nameSubCategory === 'Video');
+    const colorSetting = advVideo.parameters.find(data => data.name === 'ColorFormat');
```
andycreeth, line 175: I think you should be able to use the `values` view to make this a lot easier.  `this.views.values.Video.ColorFormat` or something like that.

andycreeth (approved):

xoredg (approved):

app/services/settings/settings.ts
gettinToasty: after checking this actually doesn't work, since the subcategory Video is a nested category and our `values` view doesn't have access to those currently. definitely a potential refactor there but i think it's out of scope for the PR

PR #4606, opened 2023-06-26
fix(twitch): shows VOD track when using different service in Advanced Mode

xoredg:
VOD track controls remain available with broken functionality if the
user is not in Protected Mode and has chosen a different service
from the Advanced Mode. This is present in the Global Audio Mixer
settings at the very least.

ref: https://app.asana.com/0/home/911512908736502/1204770688336817

### DIFF
--- a/app/services/settings/settings.ts
+++ b/app/services/settings/settings.ts
@@ -156,6 +156,10 @@ class SettingsViews extends ViewHandler<ISettingsServiceState> {
     return Utils.numberToBinnaryArray(this.values.Output.RecTracks, 6).reverse();
   }
 
+  get streamService() {
+    return this.values.Stream.service;
+  }
+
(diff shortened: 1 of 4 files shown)


app/services/settings/settings.ts
```diff
@@ -156,6 +156,10 @@ class SettingsViews extends ViewHandler<ISettingsServiceState> {
     return Utils.numberToBinnaryArray(this.values.Output.RecTracks, 6).reverse();
   }
 
+  get streamService() {
+    return this.values.Stream.service;
+  }
```
gettinToasty, line 161: i think we should maybe consider a different name here to avoid confusion since we have so many things named `xService`, i think `streamPlatform` would make sense here even if the property is called `service`

gettinToasty (approved):

PR #4619, opened 2023-07-11
Toggle dual output when logging in via dual output login prompt.

michelinewu:


app/components-react/windows/settings/Video.tsx
```diff
@@ -390,6 +390,10 @@ class VideoSettingsModule {
   handleAuth() {
     Services.WindowsService.actions.closeChildWindow();
     this.userService.actions.showLogin();
+    Services.OnboardingService.onboardingCompleted.subscribe(() => {
+      Services.DualOutputService.actions.setdualOutputMode();
+      Services.SettingsService.actions.showSettings('Video');
+    });
```
gettinToasty, line 396: i think you'll probably want to unsubscribe from this subscription while in the callback

gettinToasty (changes requested):

app/components-react/windows/settings/Video.tsx
michelinewu: Done

gettinToasty (approved):

PR #4633, opened 2023-07-21
Persist scene node map in cloud

michelinewu:


app/services/scene-collections/nodes/node-map.ts
```diff
@@ -0,0 +1,28 @@
...
+    const activeCollection = this.sceneCollectionsService.activeCollection;
+    if (activeCollection?.sceneNodeMaps) {
+      this.data = {
+        sceneNodeMaps: this.sceneCollectionsService?.sceneNodeMaps,
+      };
+    }
+  }
+
+  async load() {
+    if (this.data?.sceneNodeMaps && !this.sceneCollectionsService.hasOwnProperty('sceneNodeMaps')) {
```
gettinToasty, line 24: i'm a little worried about the way this conditional is set, i think it could lead to desync issues if a local scene collection has an outdated node map

app/services/scene-collections/nodes/root.ts
```diff
@@ -50,21 +53,24 @@ interface ISchema {
  * This is the root node of the config file
  */
 export class RootNode extends Node<ISchema, {}> {
-  schemaVersion = 4;
+  schemaVersion = 5;
```
gettinToasty, line 56: we don't usually need to update the schema if we're not depending on it's version or don't require a migration

app/services/scene-collections/nodes/scene-items.ts
```diff
@@ -139,18 +138,41 @@ export class SceneItemsNode extends Node<ISchema, {}> {
...
-    });
 
-    if (
-      this.dualOutputService.views.hasNodeMap(context.scene.id) &&
-      !this.videoSettingsService.contexts.vertical
-    ) {
-      this.videoSettingsService.establishVideoContext('vertical');
+      const nodeMap = this.dualOutputService.views.sceneNodeMaps[context.scene.id];
+
+      const verticalNodeIds = Object.entries(nodeMap).map(([key, value]: [string, string]) => {
```
gettinToasty, line 151: i think you can just use `Object.values(nodeMap)` here

app/services/scenes/scene.ts
```diff
@@ -465,8 +466,6 @@ export class Scene {
       if (nodeModel.sceneNodeType !== 'folder') return;
       this.getSelection(nodeModel.childrenIds).moveTo(this.id, nodeModel.id);
     });
-
-    this.scenesService.sourcesAdded.next(this.id);
```
gettinToasty, line 469: not sure why this was removed? disregard if it's a subscription that was only used by one thing that's not necessary anymore

gettinToasty (changes requested):
overall looking great! this seems like a much cleaner setup, just a couple of things that caught my eye

app/services/scene-collections/nodes/node-map.ts
michelinewu: Yeah, that's a great point. Pushed up a suggested fix.

app/services/scene-collections/nodes/root.ts
michelinewu: Reverted

app/services/scene-collections/nodes/scene-items.ts
michelinewu: Sure can lol

app/services/scenes/scene.ts
michelinewu: Since we now assign the displays when when loading the `scene-items` node, and assign the contexts when loading the item attributes in the `scene-item` service, this subscription was no longer needed and was only used by the dual output service to assign contexts after loading the scene items.

gettinToasty (approved):

PR #4645, opened 2023-07-31
Handle dual output scene items when working with scene items from app…

michelinewu:
…s. Remove unuser imports in apps sources service.


gettinToasty (approved):
looks like a good fix, i think we'll probably want to enable some dual output functions in the app API as well, but i think that can probably wait until after initial release

PR #4650, opened 2023-08-03
Feat/url protocol enhancement

avacreeth:


xoredg (approved):
LGTM, perhaps would've preferred the protocol to be implemented in the `host` service, requiring potentially less changes overall, but hard to tell whether we would be able to track all usages breaking because of the added protocol to the host.

PR #4652, opened 2023-08-03
Use optimizer with dual output scenes

michelinewu:


gettinToasty:
i think Dual Output is at a point of stability that we can consider refreshing/rebuilding this PR now :)

PR #4658, opened 2023-08-07
Selective recording fixes for dual output.

michelinewu:


app/services/editor.ts
```diff
@@ -569,7 +574,11 @@ export class EditorService extends StatefulService<IEditorServiceState> {
    * @returns Boolean representing if the mouse is over the source
    */
   isOverSource(event: IMouseEvent, source: SceneItem) {
-    if (event.display !== source.display) return false;
+    console.log('source ', source);
+    console.log('source.display ', source.display);
```
gettinToasty, line 578: there's a few console logs and commented imports to clean up here but otherwise looks good!

gettinToasty (changes requested):

app/services/editor.ts
michelinewu: Tidied

gettinToasty (approved):

PR #4690, opened 2023-08-24
Integrate Zustand + Immer state manager

holiber:
This PR introduces the integration of Zustand with the immer middleware, serving as an replacement to `useModule` for local state management. While the updated approach is more verbose, it relies on fewer custom code implementations. The SourceShowcase module has been successfully migrated to Zustand within this PR.

Additionally, "Modules" have been rebranded as "Controllers" in alignment with the classic MVC architecture.


package.json
```diff
@@ -94,7 +94,8 @@
     "socket.io-client": "2.3.1",
     "tasklist": "^3.1.1",
     "uuid": "^3.0.1",
-    "vuejs-datepicker": "^1.6.2"
+    "vuejs-datepicker": "^1.6.2",
+    "zustand": "^4.4.1"
```
gettinToasty, line 98: looks good so far! is there a way we can move this to devDependencies or does it need to be here? also, no issues from not being on TS5?

regardless, we should hold off on merging until after dual output release just to make sure the app is in a stable state if we need to add some dual output fixes in a bundle

gettinToasty (approved):

package.json
holiber: There were some issues with types, but I found a workaround. Yeah, Zustand should be in dev deps.

andycreeth:
@holiber this looks really great.  I think it's a testament to your original design of the `useModule` system that it can be so simply rewritten onto Zustand without major changes.  Excited to get this rolling out.

PR #4692, opened 2023-08-25
Zustand: migrate highlighter ExportModal

holiber:


app/components-react/highlighter/ExportModal.tsx
```diff
@@ -62,8 +63,19 @@ class ExportModule {
   }
 }
 
-export default function ExportModal(p: { close: () => void }) {
-  const { exportInfo, dismissError } = useModule(ExportModule);
+export const ExportModalCtx = React.createContext<ExportController | null>(null);
+
+export default function ExportModalProvider(p: { close: () => void }) {
```
andycreeth, line 68: Would love to hear your thoughts on how we should be managing these contexts.  Generally speaking I think this makes sense, and should just work if you need to add child components that need to utilize the same controller.

However, I could see this getting a little confusing if you needed to add a parent component, and then you need to move the provider/context further up into the tree.  Maybe that's all by design, just curious on your thoughts on this.

andycreeth (approved):

app/components-react/highlighter/ExportModal.tsx
```diff
@@ -74,7 +86,7 @@ export default function ExportModal(p: { close: () => void }) {
 }
 
 function ExportProgress() {
-  const { exportInfo, cancelExport } = useModule(ExportModule);
+  const { exportInfo, cancelExport } = useModule(ExportController);
```
andycreeth, line 89: Is it intentional that this wasn't replaced with `useController` and passing in the context instead?

holiber: Good catch

app/components-react/highlighter/ExportModal.tsx
holiber: I also don't like that we need to add an additional wrapper. I did my best to avoid this in `useModule` hook, however, it's a known pitfall in React
image

Zustand allows to connect the store without specifying the Context. In that case, the root context will be used, but the controller will not be destroyed on component unmount. Witch is different logic than we have before for that component

PR #4702, opened 2023-08-25
Apply dual output restrictions to hot keys.

michelinewu:


app/services/transitions.ts
```diff
@@ -163,6 +167,14 @@ export class TransitionsService extends StatefulService<ITransitionsState> {
 
   enableStudioMode() {
     if (this.state.studioMode) return;
+    if (this.dualOutputService.views.dualOutputMode) {
+      this.notificationsService.actions.push({
+        message: $t('Cannot toggle Studio Mode in Dual Output Mode.'),
```
gettinToasty, line 172: i think you might have missed a translation here, unless this sentence is already in our system in which case pls disregard!

gettinToasty (changes requested):

app/services/transitions.ts
michelinewu: I did indeed. Added

gettinToasty (approved):

PR #4716, opened 2023-09-01
Fix undo/redo errors with drag and drop and undo after conversion to dual output collection.

michelinewu:


app/services/editor-commands/commands/add-files.ts
```diff
@@ -61,5 +71,11 @@ export class AddFilesCommand extends Command {
 
   async rollback() {
     await this.removeNodesSubCommand.execute();
+
+    // try {
+    //   await this.removeNodesSubCommand.execute();
+    // } catch (error: unknown) {
+    //   console.error('Error rolling back add files command. ', error);
+    // }
```
andycreeth, line 79: Not sure if this is intentionally being left here or just some debugging code?

app/services/editor-commands/commands/copy-nodes.ts
```diff
@@ -93,7 +93,7 @@ export class CopyNodesCommand extends Command {
      */
     if (isDualOutputMode && !this.hasNodeMap) {
       // Create all nodes first
-      this.selection.getNodes().forEach(node => {
+      this.selection.getNodes().forEach(async node => {
```
andycreeth, line 96: What's the purpose of putting the async here?

andycreeth (approved):

app/services/editor-commands/commands/add-files.ts
michelinewu: removed

app/services/editor-commands/commands/copy-nodes.ts
michelinewu: It's to made sure the node and node map entry are completed created so that the dual output source selector icons show correctly

PR #4730, opened 2023-09-19
Optimizer added back.

michelinewu:


app/components-react/windows/settings/Video.tsx
```diff
@@ -136,6 +136,7 @@ class VideoSettingsModule {
       scaleType: {
         type: 'list',
         label: $t('Downscale Filter'),
+        onChange: (val: EScaleType) => this.setScaleType(val),
```
andycreeth, line 139: I'm curious where this was setting the scale type before, as I don't see you removing it anywhere else.  Are we setting this in multiple places now?

andycreeth (approved):

PR #4731, opened 2023-09-20
Refactor how sources are filtered in source selector for dual output scene collections.

michelinewu:


xoredg (approved):
Looks good to me. Only doubts are around RunInLoadingMode (which I believe we're very cautious about using, this case might justify it though) and whether the `useMemo` calls are still needed. 

EDIT: re: RunInLoadingMode, asked and answered.

xoredg (approved):

PR #4739, opened 2023-09-26
Fix recording on dual output

gettinToasty:


app/services/settings-v2/video.ts
```diff
@@ -240,6 +240,15 @@ export class VideoSettingsService extends StatefulService<IVideoSetting> {
 
   @debounce(200)
   updateObsSettings(display: TDisplayType = 'horizontal') {
+    // confirm all vertical fps settings are synced to the horizontal fps settings
+    if (display === 'vertical') {
+      this.syncFPSSettings();
+
+      // Video.video = this.state.horizontal;
+      // Video.legacySettings = this.state.horizontal;
```
gettinToasty, line 248: errant comments

app/services/settings-v2/video.ts
```diff
@@ -240,6 +240,15 @@ export class VideoSettingsService extends StatefulService<IVideoSetting> {
 
   @debounce(200)
   updateObsSettings(display: TDisplayType = 'horizontal') {
+    // confirm all vertical fps settings are synced to the horizontal fps settings
+    if (display === 'vertical') {
+      this.syncFPSSettings();
```
gettinToasty, line 245: not sure i understand this-- so every time the vertical display is updated we make the vertical fps equal to the horizontal fps?

michelinewu: Yes, this is the actual cause of the bug. If the fps settings are not the same for both contexts, the output settings is working with mismatched values, which creates this issue with speed and duration being out of sync. When a context is created, the dual output service syncs the vertical fps settings with the horizontal one. But any time we make a change to the fps settings, we need to apply this change to both contexts to keep them synced.

app/services/settings-v2/video.ts
michelinewu: Removed

app/services/settings/settings.ts
```diff
@@ -484,9 +500,23 @@ export class SettingsService extends StatefulService<ISettingsServiceState> {
...
+   * @param forceApplyCategory - name of property to force apply settings.
+   */
+  setSettings(
+    categoryName: string,
+    settingsData: ISettingsSubCategory[],
+    forceApplyCategory?: string,
+  ) {
     if (categoryName === 'Audio') this.setAudioSettings([settingsData.pop()]);
-    if (categoryName === 'Video') return;
+    if (categoryName === 'Video' && forceApplyCategory && forceApplyCategory !== 'Video') return;
```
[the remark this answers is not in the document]
michelinewu: This parameter exists because we need to guarantee that the latest video and output settings are in the store when recording. Currently, the only time that a value is passed in is in the `refreshVideoSettings` function that is called right before starting recording. We need to force the store to load the video setting but only in this instance.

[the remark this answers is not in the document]
gettinToasty: oh i see, i didn't catch the previous context of always skipping setting video settings before, this can be disregarded

PR #4754, opened 2023-10-10
Onboarding Big WIP

xoredg:


0909183417 (approved):

app/components-react/pages/onboarding/Connect.tsx
```diff
@@ -47,88 +63,176 @@ export function Connect() {
...
...KBAVBcE8O2g3MbFVHxe20ONwyDVrrqAiBbjl5yvkrAmvlHq8rxWF2N0oSV1wCYa757VQPWloqTJ/fvqmGAOfgCIrx3e7NX7Knfr0oA7A40U8HWdpi2W9n4GBNWXpZA68s7cetTz6Qicgw+3cRkZlsAeyZ+xeRMErsjzSgdPaS8DEpRTFFIaxI009/2ONsUPJvWfDvf8pIBhtP3Xx/Fkw2xrLjiJRuBktnXLEf/PYetF793zcmCAwDgayJ6QIU5Mk4LfTKNiYwuGT0q/huyzHxYazco47TsZYaMrhABmL1uUW2Hg3Q6SF2qLBCnQv8DkEbtHBFVyiLolrWZy5cxsWqbXP2LDVT6vMBtNQmjBJj5DwA3ah2K7kg77tgDxNpRdqbP9ULTuo9YAABm2Z9pVBQeQhjOOwGQ6gIr2hC98JktqaSCr2oHls66FxIDsgEQZxNEtFelDDZVlqO93nn3Nzj+D9ZXVYCvXEmpLapeiaf7jmSZiBUuwxtN2ft6857hWlCma5T1zUuOmZ8F8FEWA1RXAxkWCJ3iNKh7wP4d45uNN2HlisJ5/sbL24DCZaFnfx4AAL4lonu9EDR+F3NBr+QuUIz5fPQsBBc2x6pdTRg7xjDoz8zihjVzrub7elysCG5zgSyYhlyYwQsb1kcSdg8kmolo7uJNIoJXBgbxzoZ1EpLuIubgT08ncOffx3zrNCtg5jUAfjP/X2BPZIYmE6grWqHlsNE7GIOvASGS3i5rKXQYJPtEJjVctT0WYw5EW9F5/jsln81ZNyalMp+vwuwcUHoklVYQIaJxpWhmayx1dl/QtRlFu7bbDr+0qxfTB8RDjexvw9gXGJrRv/d38tXtsaTXQ8lhAHWqIAj5zuW16InK/Q3Fl1Nj6Bz3/wrMIfO/A7jZKQbPx3c6pWA19sjSCnxcsTHTdooaf3X8BPZPDelgqzTG11NZw1IQICh5HZCwV/DGnCNlLt9AkAleCYD0yiA4K/3HCFLIBizEzK41bzXnOQfY+FdMRJMB+x2IOtmsm43pAJAav9hKQid45RKwSVMjEf0cSPo0ldi1tyqqtBlgMXKMiB5WMexX1rqr09UXFACG/e1E9KGuMzLjzIcZMvJeMkEDYLY3TERaXaTVaeMA0ysYnd/DBMDqz6dE1CbjpLixAfC4jKxfmf8AOofb2dg1exwAAAAASUVORK5CYII=',
```
gettinToasty, line 89: is there a reason we're using a base64 image here instead of just the usual require?

app/components-react/pages/onboarding/Connect.tsx
```diff
@@ -47,88 +63,176 @@ export function Connect() {
...
+                  <PlatformIconButton
+                    platform={EPlatform.Trovo}
+                    disabled={loading || authInProgress}
+                    loading={loading}
+                    onClick={() => authPlatform(EPlatform.Trovo, afterLogin)}
+                    title={$t('Log in with %{platform}', {
+                      platform: platformLabels(EPlatform.Trovo),
+                    })}
+                  />
+                )}
```
gettinToasty, line 173: i feel like we could maybe just add trovo to the `extraPlatforms` array if the condition is met

app/components-react/pages/onboarding/Connect.tsx
gettinToasty: nvm saw this was unchanged from base branch

app/components-react/pages/onboarding/Onboarding.tsx
```diff
@@ -7,84 +8,152 @@ import cx from 'classnames';
...
+    finish,
+    UsageStatisticsService,
+    singletonStep,
+  } = useModule(OnboardingModule);
+
+  // This would probably be easier with Vuex or RxJS but not sure how easy is to rely on return values
+  const ctx = useContext(SkipContext);
+
+  const skip = () => {
+    console.log('running on skip', ctx.onSkip);
```
gettinToasty, line 34: console log

app/services/onboarding.ts
```diff
@@ -172,56 +190,114 @@ class OnboardingViews extends ViewHandler<IOnboardingServiceState> {
...
+          EOnboardingSteps.StreamingOrRecording,
+          EOnboardingSteps.Connect,
+          EOnboardingSteps.PrimaryPlatformSelect,
+          EOnboardingSteps.FreshOrImport,
+          EOnboardingSteps.ObsImport,
+          EOnboardingSteps.HardwareSetup,
+          EOnboardingSteps.ThemeSelector,
+          EOnboardingSteps.Prime,
+          EOnboardingSteps.Tips,
+        ]);
```
gettinToasty, line 256: just kind of a fun thing we can do
```
case StreamerKnowlegeMode.BEGINNER:
case StreamerKnowledgeMode.INTERMEDIATE:
  return getSteps(...)
```

if they are the same but if we are planning for them to diverge in the future i get not doing it too

gettinToasty (approved):

app/components-react/pages/onboarding/Connect.tsx
xoredg: I don't remember exactly but I think that required more changes (like in the Extra Platforms controller) since this one probably connect as opposed to the other ones that just set the Stream settings.

app/services/onboarding.ts
xoredg: Same idea, I thought to keep them separate to make it explicit, the next time we add a step we would add it to a list instead of copying all over from the other mode first

PR #4758, opened 2023-10-16
Hide optimizer/auto-config.

michelinewu:


app/services/onboarding.ts
```diff
@@ -165,7 +167,8 @@ class OnboardingViews extends ViewHandler<IOnboardingServiceState> {
 
       return ONBOARDING_STEPS()[EOnboardingSteps.Connect];
     }
-    if (this.state.options.isOptimize) return ONBOARDING_STEPS()[EOnboardingSteps.Optimize];
+    // temporarily disable auto config until migrate to new api
+    // if (this.state.options.isOptimize) return ONBOARDING_STEPS()[EOnboardingSteps.Optimize];
```
xoredg, line 171: I believe this might be the only line that's required aside from the `canRunOptimizer` call above and the ones in tests, but I'm not sure if there are random calls from anywhere else ;) It doesn't hurt to disable all around as the only argument would be re-enabling would be harder but then again it looks like API would change so it would not be a one-liner at that point. TL;DR: looks good to me 🙏

xoredg (approved):

app/services/onboarding.ts
michelinewu: Thanks! Yeah you're with me, I wanted to be thorough and disable all around even though technically I could have just hidden the button and onboarding step.

PR #4764, opened 2023-10-26
Port Modular Layouts to React

gettinToasty:


app/components-react/editor/layouts/hooks.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+export interface IResizeMins {
+  rest: number | null;
+  bar1: number | null;
+  bar2: number | null;
+}
+
+export interface ILayoutSlotArray extends Array<ILayoutSlotArray | LayoutSlot> {}
+
+export default function useLayout(
```
holiber, line 19: Can we have more comments here, especially for the vector props?

app/components-react/editor/layouts/hooks.tsx
```diff
@@ -0,0 +1,138 @@
...
+    }
+  }
+
+  function calculateMax(restMin: number) {
+    if (!componentEl.current) return 0;
+    const { height, width } = componentEl.current.getBoundingClientRect();
+    const max = isColumns ? width : height;
+    return max - restMin;
+  }
+
```
holiber, line 136: I suggest using `useCallback` for each function inside components that are always visible in our app. Otherwise, this entire list of functions will be recreated for each `useLayout` call. That leads to performance issues.

app/components-react/editor/layouts/Default.tsx
```diff
@@ -0,0 +1,57 @@
...
+    false,
+    p.childrenMins,
+    p.onTotalWidth,
+  );
+
+  return (
+    <div className={styles.rows} ref={componentRef}>
+      <div
+        className={styles.cell}
+        style={{ height: `${100 - (resizes.bar1 + resizes.bar2!) * 100}%` }}
```
michelinewu, line 19: In what instances would `resizes.bar2` be null?

michelinewu (approved):

PR #4772, opened 2023-10-31
Mw add do tests 2

michelinewu:


test/helpers/scene-builder.ts
```diff
@@ -228,4 +243,44 @@ export class SceneBuilder {
...
+    const nodeMap = sceneNodeMaps[activeSceneId];
+    if (!nodeMap) return false;
+
+    // confirm all horizontal nodes have a partner vertical node
+    return nodes.reduce((hasPartner: boolean, node: ISceneBuilderNode) => {
+      if (node?.display === 'horizontal' && !nodeMap[node.id]) {
+        hasPartner = false;
+      }
+      return hasPartner;
+    }, true);
```
gettinToasty, line 268: this is pretty nitpicky and doesn't require a refactor but just letting you know! JS has `array.every` and `array.some` methods now that seem ideal for a lot of these places you're using reduce

gettinToasty (approved):

PR #4787, opened 2023-11-15
refactor: change "themes" to "overlays"

xoredg:


gettinToasty (changes requested):
we're going to need to change the corresponding json entries in the i18n/en-US folder as well

PR #4794, opened 2023-11-17
Dual output recording

michelinewu:


app/services/api/external-api/streaming/streaming.ts
```diff
@@ -21,8 +21,10 @@ enum EStreamingState {
 enum ERecordingState {
   Offline = 'offline',
   Starting = 'starting',
+  Start = 'start',
```
xoredg, line 24: Can we add a comment here, as to how `start` and `stop` are different?

app/services/dual-output/dual-output.ts
```diff
@@ -570,6 +633,16 @@ export class DualOutputService extends PersistentStatefulService<IDualOutputServ
     this.SET_IS_LOADING(status);
   }
 
+  /**
+   * Part of a workaround to set recording quality on the frontend to avoid a backend bug
+   * @remark TODO: remove when migrate output and stream settings to new API
+   * @param mode - single or dual output
+   * @param quality - recording quality
+   */
+  setDualOutputRecordingQuality(mode: 'single' | 'dual', quality: ERecordingQuality) {
```
xoredg, line 642: Could we extract `'single' | 'dual'` to a type so we can use it below (and probably in other places)?

app/services/recording-mode.ts
```diff
@@ -180,12 +191,22 @@ export class RecordingModeService extends PersistentStatefulService<IRecordingMo
...
+   * Add entry to recording history and show notification
+   * @param filename - name of file to show in recording history
+   * @param showNotification - primarily used when recording in dual output mode to only show the notification once
+   */
+  addRecordingEntry(filename: string, display?: TDisplayType) {
     const timestamp = moment().format();
     this.ADD_RECORDING_ENTRY(timestamp, filename);
+
+    const message = display
+      ? $t(`A new ${capitalize(display)} Recording has been completed. Click for more info`)
```
xoredg, line 204: I believe translation team would like `capitalize(display)` to be interpolated so that it can be translated. 
e.g `$t('A new %{displayType} Recording...', { displayType: $t(display /* or capitalize(display) */) }`

app/services/streaming/streaming.ts
```diff
@@ -792,45 +824,78 @@ export class StreamingService
...
-      const horizontalContext = this.videoSettingsService.contexts.horizontal;
-      const verticalContext = this.videoSettingsService.contexts.vertical;
+        const signalChanged = this.signalInfoChanged.subscribe(
+          async (signalInfo: IOBSOutputSignalInfo) => {
+            if (signalInfo.service === 'default') {
+              if (signalInfo.signal === EOBSOutputSignal.Start) {
+                // Refactor when migrate to new API, sleep to allow state to update
+                await new Promise(resolve => setTimeout(resolve, 300));
+                this.toggleRecording();
+                signalChanged.unsubscribe();
```
xoredg, line 842: If I remember correctly RxJS didn't play nicely when unsubscribing to a subject inside a `subscribe` function. I believe this could be achieved with `take(1)` or `takeWhile(condition)` and then `subscribe` which would automatically unsubscribe for us. 
Bear in mind is been a while so API might have changed or even this assumption. 
```
this.signalInfoChanged.pipe(
  take(1)
 ).subscribe(...)
 ```
I believe the filter (if takeWhile isn't needed after all, it might) and the delay could also be achieved with operators, but that's not a real concern, just the subscribe/unsubscribe behavior is.

File: app/services/streaming/streaming.ts [hunk 5]
```diff
@@ -982,6 +1061,308 @@ export class StreamingService
...
   }
 
+  /**
+   * Part of a workaround to set recording quality on the frontend to avoid a backend bug
+   * @remark TODO: remove when migrate output and stream settings to new API
+   * Because of a bug with the new API and having the `Same as stream` setting, temporarily store and set
+   * the recording quality on the frontend
+   * @param mode - single output or dual output
+   */
+  beforeStartRecording(mode: 'single' | 'dual') {
```
xoredg, line 1071: Here's another instance of the `'single' | 'dual'` type.

app/services/streaming/streaming.ts
```diff
@@ -906,14 +971,18 @@ export class StreamingService
...
             if (
               signalInfo.service === 'default' &&
               signalInfo.signal === EOBSOutputSignal.Deactivate
             ) {
               NodeObs.OBS_service_stopStreaming(false, 'vertical');
+              // Refactor when move streaming to new API
+              if (this.state.verticalStreamingStatus !== EStreamingState.Offline) {
+                this.SET_VERTICAL_STREAMING_STATUS(EStreamingState.Offline);
+              }
               signalChanged.unsubscribe();
```
xoredg, line 986: RxJS unsubscribe comment earlier.

app/services/streaming/streaming.ts
```diff
@@ -982,6 +1061,308 @@ export class StreamingService
...
+              // if the same time stamp is used, the entry will be replaced in the recording history
+              setTimeout(() => {
+                time = new Date().toISOString();
+                this.SET_RECORDING_STATUS(ERecordingState.Stopping, time, 'horizontal');
+                if (this.horizontalRecording !== null) {
+                  this.horizontalRecording.stop();
+                }
+              }, 2000),
+            );
+            recordingStopped.unsubscribe();
```
xoredg, line 1158: RxJS unsubscribe comment earlier.

app/services/streaming/streaming.ts
```diff
@@ -982,6 +1061,308 @@ export class StreamingService
...
+    // map signals to status
+    const nextState: ERecordingState = ({
+      [EOutputSignalState.Start]: ERecordingState.Recording,
+      [EOutputSignalState.Stop]: ERecordingState.Offline,
+      [EOutputSignalState.Stopping]: ERecordingState.Stopping,
+      [EOutputSignalState.Wrote]: ERecordingState.Wrote,
+    } as Dictionary<ERecordingState>)[info.signal];
+
+    // We received a signal we didn't recognize
+    if (!nextState) return;
```
xoredg, line 1275: I think logging out (or throwing errors on) these `will never happen` events would be good as we'd have the peace of mind it'll never silently happen.

blackxored (changes requested):
Most of the changes requested are cosmetic. The missing translations are probably the most important. The RxJS behavior following after, but it seems to be working correctly, so definitely up to you whether to refactor would provide value.

app/components-react/root/StudioEditor.tsx
```diff
@@ -347,30 +347,58 @@ function StudioModeControls(p: { stacked: boolean }) {
...
+          <DualOutputIcons
+            className={cx('icon-horizontal', horizontalIconVisible && 'visible')}
+            showStream={v.isHorizontalStreaming}
+            showRecord={v.isHorizontalRecording}
+          />
         </div>
       )}
-
-      {showVertical && (
+      {}
```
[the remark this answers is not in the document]
michelinewu: Removed

app/components-react/root/StudioEditor.tsx
```diff
@@ -347,30 +347,58 @@ function StudioModeControls(p: { stacked: boolean }) {
...
+      {v.showVertical && (
         <div className={styles.verticalHeader}>
           <i className="icon-phone-case" />
           <span>{$t('Vertical Output')}</span>
+          <DualOutputIcons
+            className={cx('icon-vertical', verticalIconVisible && 'visible')}
```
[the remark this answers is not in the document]
michelinewu: Done, same as above

app/components-react/root/StudioFooter.tsx
```diff
@@ -202,13 +204,7 @@ function RecordingButton() {
...
             onClick={toggleRecording}
           >
-            <span>
-              {recordingStatus === ERecordingState.Stopping ? (
-                <i className="fa fa-spinner fa-pulse" />
-              ) : (
-                <>REC</>
-              )}
-            </span>
+            <span>{recordingStopping ? <i className="fa fa-spinner fa-pulse" /> : <>REC</>}</span>
```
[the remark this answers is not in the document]
michelinewu: This is how it was originally but you bring up a good point. The limitation is that there is only room for three characters on the button. Let's bring it up to the team?

app/components-react/windows/go-live/StreamOptions.tsx
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Renders options for the stream
+ * - Vertical Recording Settings
+ **/
+export default function StreamOptions() {
+  const { DualOutputService } = Services;
+
+  const { Option } = Select;
+
+  const recordingQualities = [
```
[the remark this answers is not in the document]
michelinewu: Done

app/components-react/windows/go-live/StreamOptions.tsx
```diff
@@ -0,0 +1,83 @@
...
+              {option.name}
+            </Option>
+          ))}
+        </Select>
+      </Row>
+      <Row gutter={16} className={styles.settingsRow}>
+        <div className={styles.switcherLabel}>{$t('Vertical Recording Only')}</div>
+        <SwitchInput
+          value={v.recordVertical}
+          name={'record-vertical'}
```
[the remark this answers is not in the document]
michelinewu: Done

app/services/api/external-api/streaming/streaming.ts
michelinewu: Done

app/services/streaming/streaming.ts
michelinewu: Done

app/services/recording-mode.ts
michelinewu: Done

app/services/dual-output/dual-output.ts
michelinewu: Done

app/services/streaming/streaming.ts
michelinewu: Just for tracking, discussed

app/services/streaming/streaming.ts
michelinewu: Discussed

app/services/streaming/streaming.ts
michelinewu: Discussed

app/components-react/root/StudioFooter.tsx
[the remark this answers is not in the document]
gettinToasty: yea the character limit is the issue here, as well as i'm not certain other languages have the same shorthand that we do for this term. i think looking into an icon replacement could have value here, as there is a generally accepted icon for recording

gettinToasty (approved):

PR #4800, opened 2023-11-18
Move convert tool to scene collections

michelinewu:


app/components/windows/settings/OverlaySettings.tsx
```diff
@@ -99,6 +114,75 @@ export default class OverlaySettings extends Vue {
...
+   */
+  async convertDualOutputCollection(
+    assignToHorizontal: boolean = false,
+    exportOverlay: boolean = false,
+  ) {
+    // confirm that the active scene collection is a dual output collection
+    if (
+      !this.sceneCollectionsService?.sceneNodeMaps ||
+      (this.sceneCollectionsService?.sceneNodeMaps &&
+        Object.values(this.sceneCollectionsService?.sceneNodeMaps).length === 0)
```
xoredg, line 131: Nitpick: would prefer checking `Object.keys` instead as runtime optimizations are most likely to be in place there.

app/components/windows/settings/OverlaySettings.tsx
```diff
@@ -140,6 +224,61 @@ export default class OverlaySettings extends Vue {
...
+              })}
+            />
+            <button
+              class="button button--soft-warning"
+              onClick={async () => await this.convertDualOutputCollection()}
+              disabled={this.busy}
+            >
+              {$t('Convert')}
+            </button>
+            {/* <button
```
xoredg, line 255: I'm assuming this is disabled, would prefer a small flag on the VM instead of commenting, but again, nitpick territory ;)

xoredg (approved):

app/components/windows/settings/OverlaySettings.tsx
gettinToasty: i believe this is something being worked on but not ready to ship yet

PR #4815, opened 2023-12-12
fix(streamlabels): newlines not saved on streamlabel message format

xoredg:
Fixes inputting a newline on the message format field of a Streamlabel
showing correctly on its preview, but not on the actual source or
persisted correctly on the backend.
Might also fix the item separator issue or is no longer present since
I couldn't reproduce this one.

ref: https://app.asana.com/0/734380881425048/1185957128889367/f

### DIFF
--- a/app/services/streamlabels/index.ts
+++ b/app/services/streamlabels/index.ts
@@ -243,6 +243,7 @@ export class StreamlabelsService extends StatefulService<IStreamlabelsServiceSta
 
   getSettingsForStat(statname: string) {
     const settings = { ...this.settings[statname] };
+    console.log('getting setting for ', statname, settings);
 
     if (settings.item_separator) {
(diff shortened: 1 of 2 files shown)


app/services/streamlabels/index.ts
```diff
@@ -243,17 +243,30 @@ export class StreamlabelsService extends StatefulService<IStreamlabelsServiceSta
 
   getSettingsForStat(statname: string) {
     const settings = { ...this.settings[statname] };
+    console.log('getting setting for ', statname, settings);
```
gettinToasty, line 246: looks like there's an errant console log here but otherwise lgtm

gettinToasty (changes requested):

gettinToasty (approved):

PR #4818, opened 2023-12-13
Game window capture error handling

avoitenko-logitech:


gettinToasty:
Looks like there was a strict null error on CI here

gettinToasty (changes requested):
Tests need to be fixed here

avoitenko-logitech:
`media-backup` test is failing which is not related to my changes.

app/services/sources/sources.ts
```diff
@@ -356,7 +358,7 @@ export class SourcesService extends StatefulService<ISourcesState> {
     const source = this.views.getSource(id)!;
     const muted = obsInput.muted;
     this.UPDATE_SOURCE({ id, muted });
-    this.updateSourceFlags(source.state, obsInput.outputFlags, true);
+    this.updateSourceFlags(source, obsInput.outputFlags, true);
```
michelinewu, line 361: Why remove `.state`?

app/services/sources/sources.ts
```diff
@@ -616,16 +618,28 @@ export class SourcesService extends StatefulService<ISourcesState> {
     });
   }
 
-  private updateSourceFlags(source: ISource, flags: number, doNotEmit?: boolean) {
+  private updateSourceFlags(source: Source, flags: number, doNotEmit?: boolean) {
```
michelinewu, line 621: Why `Source` instead of `ISource`?

avoitenko-logitech: Because `ISource` does not have `getModel` method.

app/services/sources/sources.ts
avoitenko-logitech: Because I need to pass `Source` instead of `ISource`

PR #4825, opened 2023-12-18
Port LiveDock to React

gettinToasty:


app/components-react/root/LiveDock.tsx
```diff
@@ -0,0 +1,437 @@
...
+    return this.customizationService.state.hideViewerCount;
+  }
+
+  get liveDockSize() {
+    return this.customizationService.state.livedockSize;
+  }
+
+  get viewerCount() {
+    if (this.hideViewerCount) {
+      return 'viewers hidden';
```
xoredg, line 105: Do we need this translated

app/components-react/root/LiveDock.tsx
```diff
@@ -0,0 +1,437 @@
...
+
+  get hideStyleBlockers() {
+    return this.windowsService.state.main.hideStyleBlockers;
+  }
+
+  get hasChatTabs() {
+    return this.chatTabs.length > 1;
+  }
+
+  get showDefaultPlatformChat() {
```
xoredg, line 118: Cosmetic: perhaps rename to `shouldShowDefaultPlatformChat` as it's not an action?

app/components-react/root/LiveDock.tsx
```diff
@@ -0,0 +1,437 @@
...
+          };
+        }),
+    );
+    if (this.restreamService.shouldGoLiveWithRestream) {
+      tabs.push({
+        name: $t('Multistream'),
+        value: 'restream',
+      });
+    }
+    if (this.userService.state.auth.primaryPlatform === 'twitter') {
```
xoredg, line 157: Maybe we should add Instagram here as well, odds of that ending as primary platform are slim, but if we feel like being super safe.

app/components-react/root/LiveDock.tsx
```diff
@@ -0,0 +1,437 @@
...
+    collapsed,
+    isPlatform,
+    isStreaming,
+    hasChatTabs,
+    chatTabs,
+    selectedChat,
+    liveDockSize,
+    applicationLoading,
+    hideStyleBlockers,
+  } = useVuex(() => ({
```
xoredg, line 307: Cosmetic, use `pick` on keys as they're named the same as their source.

xoredg (approved):
Approved with minor/cosmetic suggestions.

app/components-react/root/LiveDock.tsx
gettinToasty: i'm going to leave this unchanged for now since that doesn't seem necessary currently, and will likely be less necessary after your changes

PR #4827, opened 2024-01-03
Use PKCE for OAuth

andycreeth:


gettinToasty (approved):
LGTM, i'm a little surprised there isn't a native `urlsafeBase64` encoding option available but looks like there isnt 🤔

andycreeth:
> LGTM, i'm a little surprised there isn't a native `urlsafeBase64` encoding option available but looks like there isn't 🤔

It exists in later versions of Node, and I think theoretically even in the version of Node we're on.  But I was getting TypeScript errors when trying to use it.  There are only 2 URL-unsafe characters in base64 so it's a trivial substitution to do manually anyway.

PR #4830, opened 2024-01-05
Batch server requests when updating scene collections from the server

gettinToasty:


xoredg (approved):
Looks good to me. At your discretion whether we want to update `collectionsToInsert` and `collectionsToUpdate` to use `Set`s for extra unneccessary dupe safety ;)

PR #4835, opened 2024-01-11
fix display jank with realm

andycreeth:


gettinToasty (approved):

gettinToasty:
very nitpicky but i think `requestAnimationFrame` might be the slightly better option in this usecase as i think it can insert directly after a repaint and not be forced to wait 16ms

PR #4853, opened 2024-01-30
Port Main Window to React

gettinToasty:


app/components-react/editor/elements/mixer/GLVolmeters.tsx
```diff
@@ -358,7 +361,7 @@ class GLVolmetersController {
     this.gl.clearColor(bg.r / 255, bg.g / 255, bg.b / 255, 1);
     this.gl.clear(this.gl.COLOR_BUFFER_BIT);
 
-    if (this.canvasWidth < 0 || this.canvasHeight < 0) return;
+    if (this.canvasWidth < 0 || this.canvasHeight < 0 || !this.sourcesOrder) return;
```
blackxored, line 364: nit: add braces, control flow can be easy to miss

app/components-react/pages/PlatformAppStore.tsx
```diff
@@ -85,14 +89,19 @@ export default function PlatformAppStore(p: { params: { appId?: string; type?: s
   return (
     <>
       <BrowserView
-        className={styles.browserView}
+        className={cx(styles.browserView, p.className)}
         style={{
           height: `calc(100% - ${
```
blackxored, line 94: could we extract this interpolation to a var? seems like we're evaluating a lot

app/components-react/root/LiveDock.tsx
```diff
@@ -364,115 +334,93 @@ function LiveDock(p: { onLeft: boolean }) {
...
+                  appId={visibleChat}
+                  pageSlot={pageSlot}
+                  key={visibleChat}
+                />
+              )}
+            </div>
+          )}
+      </div>
+      {(!ctrl.platform ||
+        (isPlatform(['youtube', 'facebook', 'twitter', 'tiktok', 'kick']) && !isStreaming)) && (
```
blackxored, line 418: not really a comment, but a note to self, lots of `isPlatform` evaluations over here, most of the time with the same arguments, that should probably be extracted in the future. It will make it easier to support newer platforms as well instead of having to track all these instances

blackxored (approved):

app/components-react/root/LiveDock.tsx
gettinToasty: i think that's slightly out of scope for this PR but yes agreed, i have a thought to handle this similar how we use `hasCapability` for platform features

gettinToasty: made a follow up PR here for when this is merged
https://github.com/streamlabs/desktop/pull/5401

michelinewu (approved):

PR #4855, opened 2024-01-31
feat(multistream): relax primary platform requirement

blackxored:


app/components-react/windows/go-live/DestinationSwitchers.tsx
```diff
@@ -42,25 +41,31 @@ export function DestinationSwitchers() {
     emitSwitch();
   }
 
+  // TODO: find a cleaner way to do this
+  const isPrimary = (platform: TPlatform) =>
+    linkedPlatforms.length === 1 ? true : isPrimaryPlatform(platform);
```
gettinToasty, line 46: could just be `linkedPlatforms.length === 1 || isPrimaryPlatform(platform)`

app/components-react/windows/go-live/DestinationSwitchers.tsx
```diff
@@ -42,25 +41,31 @@ export function DestinationSwitchers() {
...
-      ))}
-      {customDestinations?.map((dest, ind) => (
-        <DestinationSwitcher
-          key={ind}
-          destination={dest}
-          enabled={customDestinations[ind].enabled}
-          onChange={enabled => switchCustomDestination(ind, enabled)}
-        />
-      ))}
+      <div>
```
gettinToasty, line 50: not sure why an extra div was added here?

app/components-react/windows/go-live/DestinationSwitchers.tsx
```diff
@@ -80,17 +85,20 @@ function DestinationSwitcher(p: IDestinationSwitcherProps) {
   const containerRef = useRef<HTMLDivElement>(null);
   const platform = typeof p.destination === 'string' ? (p.destination as TPlatform) : null;
   const { RestreamService, MagicLinkService } = Services;
+  const canEnableRestream = RestreamService.views.canEnableRestream;
+  const canDisableDestination = canEnableRestream ? true : !p.isPrimary;
```
gettinToasty, line 89: `canEnableRestream || !p.isPrimary`

app/components-react/windows/go-live/DestinationSwitchers.tsx
```diff
@@ -80,17 +85,20 @@ function DestinationSwitcher(p: IDestinationSwitcherProps) {
...
 
   function onClickHandler(ev: MouseEvent) {
-    if (p.isPrimary) {
+    if (!canDisableDestination) {
```
gettinToasty, line 92: if we're only checking negatives for booleans we should probably reformulate the boolean to be a positive identifier ie `cannotDisableDestination`

app/components-react/windows/go-live/dual-output/DualOutputGoLiveSettings.tsx
```diff
@@ -21,31 +22,51 @@ import UserSettingsNonUltra from './UserSettingsNonUltra';
...
+      canUseOptimizedProfile: false,
+      // canUseOptimizedProfile:
+      //   VideoEncodingOptimizationService.state.canSeeOptimizedProfile ||
+      //   VideoEncodingOptimizationService.state.useOptimizedProfile,
+    };
+  });
+
+  const shouldShowPrimaryChatSwitcher = isRestreamEnabled && hasMultiplePlatforms;
+  // TODO: make sure this doesn't jank the UI
+  const leftPaneHeight = shouldShowPrimaryChatSwitcher ? '82%' : '100%';
```
gettinToasty, line 52: i think some sort of `calc` operation here would be more reliable, although admittedly a bit more expensive to compute

app/components-react/windows/go-live/dual-output/NonUltraDestinationSwitchers.tsx
```diff
@@ -99,6 +101,7 @@ interface IDestinationSwitcherProps {
   onChange: (enabled: boolean) => unknown;
   isPrimary?: boolean;
   index: number;
+  canDisablePrimary?: boolean;
```
gettinToasty, line 104: this also looks like its only used to check for the negative, see above

app/services/streaming/streaming-view.ts
```diff
@@ -248,10 +248,24 @@ export class StreamInfoView<T extends Object> extends ViewHandler<T> {
...
    * Chat url of a primary platform
+   * If the primary platform is not enabled, and we're on single stream mode,
+   * returns the URL of the first enabled platform
    */
   get chatUrl(): string {
     if (!this.userView.isLoggedIn || !this.userView.auth) return '';
-    return getPlatformService(this.userView.auth.primaryPlatform).chatUrl;
+    // TODO: transition away from primaryPlatform completely when possible
+    // TODO: handle custom destinations
+    // TODO: handle selected platform not supporting chat
```
gettinToasty, line 258: i dont think these todos are todos anymore:
1. we pivoted to primaryPlatform === primaryChat
2 and 3. we just wouldnt display chat for those

gettinToasty (changes requested):

app/components-react/sidebar/PlatformIndicator.m.less
```diff
@@ -0,0 +1,71 @@
...
+  i {
+    &,
+    &:hover {
+      transform: none !important;
+      transition: none !important;
+    }
+  }
+  // Custom destination icon
+  // TODO: why does this need to be global?
+  :global(.fa-globe) {
```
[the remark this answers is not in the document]
xoredg: I understand that, I'm just not sure why it wouldn't apply to standard DOM elements like `.fa-globe` inside a parent that has the `platform-icons` class applied to it.

app/components-react/windows/go-live/dual-output/NonUltraDestinationSwitchers.tsx
gettinToasty: actually yea thats a pretty good point, disregard

app/components-react/sidebar/PlatformIndicator.m.less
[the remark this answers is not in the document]
gettinToasty: i don't think i'm 100% clear about what you're asking, but here are how the two would compile to CSS if this helps to illustrate:
```
// LESS
.platform-icons {
  .fa-globe

// CSS
.platform-icons_sgshe534 .fa-globe_397sfe {
```
vs
```
// LESS
.platform-icons {
  :global(.fa-globe) {
  
// CSS
.platform-icons_sgshe534 .fa-globe {
```
so because fontawesome doesnt have a class named `fa-globe_397sfe` none of those styles would be applied

app/components-react/windows/go-live/dual-output/DualOutputGoLiveSettings.tsx
xoredg: My very limited CSS speaking, I'd have to figure out how tall the Primary Chat switcher is, where as this seems to work? Granted, it is a hardcoded value but I believe `calc` expression would also be? Open to suggestions.

app/components-react/sidebar/PlatformIndicator.m.less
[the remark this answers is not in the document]
xoredg: Perfect, that makes sense, though I'd swear I've seen bare classes before (or perhaps it was only elements) working before. TIL.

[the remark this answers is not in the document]
gettinToasty: bare elements definitely work, and in any file appended `.g.less` bare classes will work, but all `.m.less` files are scoped

in legacy Vue files as well scoped bare classes will still work because the way Vue applies style scopes is insane KEKW

app/components-react/windows/go-live/dual-output/DualOutputGoLiveSettings.tsx
xoredg: It doesn't, it's just a one element form with a divider.

gettinToasty (approved):

PR #4872, opened 2024-02-20
Slider input throttle

avoitenko-logitech:
Reducing number of events generated on slider movement


gettinToasty (changes requested):
looks like there are some strict null failures here

PR #4879, opened 2024-02-26
Fix Onboarding Completed Step Firing Incorrectly

gettinToasty:


xoredg (approved):
This looks good to me, we just need to verify that the events are indeed firing when the step is ran through onboarding itself. The conditions on `singletonStep` should be enough but we just need to verify with testing.

PR #4881, opened 2024-02-27
fix(build): react-devtools build folder was ignored

blackxored:
React DevTools build folder was ignored from one of our global patterns. Didn't realize this until I had to reclone the repo.


xoredg:
There's a CSRF issue identified so it might not be as simple of a review. I don't believe the param is actually user-provided but dynamically generated by the extension to work, looking like a false positive. That said, considering this and also that React Devtools doesn't work without `allowFileAccess` we should take extra care.

gettinToasty (approved):

PR #4882, opened 2024-02-27
fix(dual-output): inconsistent source visibility status, details follow

xoredg:
Switching scenes makes dual-output horizontal/vertical visibility settings get into a weird state. Since restarting the app fixes it,
it shouldn't be a persistence problem.

I've found that having `isLoading` check prepended to the visibility booleans seems to be the root cause, as when switching scenes (see [dual-output.ts#327](https://github.com/stream-labs/desktop/blob/8995f39e250b68088b02df12cebb8fcb4143a517/app/services/dual-output/dual-output.ts#L327)), this triggers a load that apparently doesn't ever re-render the `DualOputputSourceSelector` component properly.

I haven't found any issues when removing these checks, but I've definitely just fixed the symptom here. If anything, we should probably move the check to the parent component, investigate why `isLoading` doesn't end up with a valid state, or both.

ref: https://app.asana.com/0/734380881425048/1206345543167836/f


xoredg:
Closed in favor of https://github.com/stream-labs/desktop/pull/4883 which thanks to @michelinewu also simplifies loading state.

PR #4896, opened 2024-03-13
Fix game capture scaling on vertical display on import overlay.

michelinewu:


xoredg (approved):
Thanks for the catch! Just a thought: should we remove the manual setting of scaling that's on this file when the item is game capture (around the area where we create the sources based on the nodes)? It seems like `adjustTransform` will overwrite it anyways, so I'd assume it wouldn't hurt to remove it, for clarity sake.

michelinewu:
> Thanks for the catch! Just a thought: should we remove the manual setting of scaling that's on this file when the item is game capture (around the area where we create the sources based on the nodes)? It seems like `adjustTransform` will overwrite it anyways, so I'd assume it wouldn't hurt to remove it, for clarity sake.

Good call, I'll make the change

PR #4899, opened 2024-03-14
feat(overlays): preserve visiblity on export/import

xoredg:
Ensure locked status and visibility (including horizontal/vertical display visibility) is preserved when exporting and importing overlays. 

ref: https://app.asana.com/0/734380881425048/1206825809745162/f


xoredg:
Should supersede #4893 if we're fine with this approach.

gettinToasty (approved):

PR #4911, opened 2024-03-21
feat(twitch): add Branded content and Content Classification fields

blackxored:
ref: https://app.asana.com/0/home/911512908736502/1205060922063910

### DIFF
--- a/app/services/platforms/twitch.ts
+++ b/app/services/platforms/twitch.ts
@@ -21,6 +21,10 @@ import Utils from '../utils';
 import { IVideo } from 'obs-studio-node';
 import { TDisplayType } from 'services/settings-v2';
 import { TOutputOrientation } from 'services/restream';
+import {
+  ITwitchContentClassificationLabelsRootResponse,
+  TwitchContentClassificationService,
+} from './twitch/content-classification';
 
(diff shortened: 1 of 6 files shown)


michelinewu (approved):

app/services/platforms/twitch.ts
```diff
@@ -313,11 +346,24 @@ export class TwitchService
...
     const hasPermission = await this.hasScope('channel:manage:broadcast');
     const scopedTags = hasPermission ? tags : undefined;
+
+    // Twitch seems to require you to add a label with disabled to remove it
+    const labels = this.twitchContentClassificationService.options.map(option => ({
+      id: option.value,
+      is_enabled: contentClassificationLabels.includes(option.value),
+    }));
+
     await Promise.all([
```
holiber, line 356: I'm not sure why we have `Promise.all` here, lol. Should be just `await this.requestTwitch`

holiber (approved):

app/services/platforms/twitch.ts
gettinToasty: i think there previously may have been other steps or something 🤔

PR #4913, opened 2024-03-22
fix(settings): reactivity of "confirm title" needs to use Realm hook

xoredg:


gettinToasty (approved):

gettinToasty:
there might be some other stealth bugs as a result of merging Realm as well, we should be vigilant about this 👀

PR #4931, opened 2024-04-09
Fix realm thingy

andycreeth:


blackxored (approved):
Looks good. On the very cosmetic side I'd use a `typeof ... === 'string'` check to make it super obvious it could be either a string or an object, but this works.

gettinToasty (approved):

PR #4936, opened 2024-04-11
Dual Output collection: refactor create, optimize load, repair on load.

michelinewu:


app/services/dual-output/dual-output.ts
```diff
@@ -360,80 +381,291 @@ export class DualOutputService extends PersistentStatefulService<IDualOutputServ
...
+   * @remark The horizontal node id is always the key in the scene node map.
+   * The node map entry is so that the horizontal and vertical nodes can refer to each other.
+   */
+  createVerticalNode(horizontalNode: TSceneNode, verticalNodeId?: string): TSceneNode {
+    const scene = horizontalNode.getScene();
+
+    if (horizontalNode.isFolder()) {
+      // add folder and create node map entry
+      const folder = scene.createFolder(horizontalNode.name, {
+        id: verticalNodeId,
```
gettinToasty, line 428: im not sure how this works given that `verticalNodeId` is an optional parameter? how do we ensure that the folder has a valid id?

app/services/dual-output/dual-output.ts
```diff
@@ -360,80 +381,291 @@ export class DualOutputService extends PersistentStatefulService<IDualOutputServ
...
+        display: 'vertical',
+      });
+      this.sceneCollectionsService.createNodeMapEntry(scene.id, horizontalNode.id, folder.id);
+
+      this.sceneNodeHandled.next();
+      return folder;
+    } else {
+      // add item
+      const item = scene.addSource(horizontalNode.sourceId, {
+        id: verticalNodeId,
```
gettinToasty, line 438: same as above ^

app/services/dual-output/dual-output.ts
```diff
@@ -360,80 +381,291 @@ export class DualOutputService extends PersistentStatefulService<IDualOutputServ
...
   }
 
   /**
-   * Assign or confirm node contexts to a dual output scene
-   * @param sceneId - Id of the scene to map
+   * Create a horizontal node to partner with the vertical node
+   * @param verticalNode - Node to copy to the horizontal display
+   *
+   * @remark The horizontal node id is always the key in the scene node map.
+   * The node map entry is so that the horizontal and vertical nodes can refer to each other.
```
gettinToasty, line 464: in which situations are we expecting to have vertical nodes without a corresponding horizontal node?

michelinewu: Right now, there should never be a vertical node without a corresponding horizontal node. If there is for whatever reason, when the collection is loaded, the repair function will create a horizontal node to partner with the vertical node. In the scene source fix, this will introduce the instance where there is a vertical node created without a horizontal one. But this is because the scene source fix creates a vertical scene that is only ever rendered as a source for a scene-as-scene-item. More on that to come.

app/services/dual-output/dual-output.ts
michelinewu: See above

gettinToasty (approved):

app/services/dual-output/dual-output.ts
```diff
@@ -360,80 +381,291 @@ export class DualOutputService extends PersistentStatefulService<IDualOutputServ
...
+    } else {
+      // add item
+      const item = scene.addSource(horizontalNode.sourceId, {
+        id: verticalNodeId,
+        display: 'vertical',
+      });
+
+      // position all of the nodes in the upper left corner of the vertical display
+      // so that all of the sources are visible
+      item.setTransform({ position: { x: 0, y: 0 } });
```
blackxored, line 444: For the future, please ignore for now, would there be a way we can avoid forcing manipulating the scene item properties this way? If I remember correctly we have to workaround these on load to preserve position, visibility, etc. somewhere around the loading scene sequence, I wonder if we should at least move this code there at some point, so we'd look at one place for everything related to it, as opposed to being split.

app/services/dual-output/dual-output.ts
```diff
@@ -360,80 +381,291 @@ export class DualOutputService extends PersistentStatefulService<IDualOutputServ
...
+      // add folder and create node map entry
+      const folder = scene.createFolder(verticalNode.name, {
+        id: horizontalNodeId,
+        display: 'horizontal',
+      });
+      folder.placeBefore(verticalNode.id);
+
+      this.sceneCollectionsService.createNodeMapEntry(scene.id, folder.id, verticalNode.id);
+
+      folder.placeBefore(verticalNode.id);
```
blackxored, line 483: I don't know is `placeBefore` is idempotent or not, but it looks like we're calling it twice? We're not doing the same for `item` below, so I wonder if there's something special with folders.

blackxored (changes requested):
Only one real issue `folder.placeBefore` being called twice.

app/services/dual-output/dual-output.ts
michelinewu: This was a decision that was made so that all vertical items would be available to the user when converting from a single output to a dual output collection. If we preserve the properties and don't scale the horizontal scene, some scene items could be created outside of the canvas, potentially outside of the padding boundary we set in the editor in dual output mode so that scene items couldn't be dragged out of the editor. If we scale the horizontal scene, some vertical scene items could be too small to interact with. Repositioning the sources was not actually a decision I made. I'm open to another solution (such as partial scaling) but this would be a product decision.

app/services/scenes/scenes.ts
```diff
@@ -260,9 +262,12 @@ class ScenesViews extends ViewHandler<IScenesState> {
 
     return false;
   }
+
+  getNodeMap(): Dictionary<string> | undefined {
+    return this.state?.nodeMap;
```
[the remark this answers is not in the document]
michelinewu: Done

app/services/scenes/scene-item.ts
```diff
@@ -323,7 +320,7 @@ export class SceneItem extends SceneItemNode {
       blendingMethod: customSceneItem.blendingMethod,
       display,
```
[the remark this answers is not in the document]
michelinewu: Done

app/services/scene-collections/nodes/scene-items.ts
```diff
@@ -139,40 +140,17 @@ export class SceneItemsNode extends Node<ISchema, {}> {
...
     this.sanitizeIds();
 
-    // on first load, a dual output scene needs to assign displays and contexts to the scene items
-    // but if the scene item already has a display assigned, skip it
-    if (this.dualOutputService.views.hasNodeMap(context.scene.id)) {
-      // nodes must be assigned to a context, so if it doesn't exist, establish it
-      if (!this.videoSettingsService.contexts.vertical) {
-        this.videoSettingsService.establishVideoContext('vertical');
+    this.data.items.forEach(item => {
+      if (!item?.display) {
```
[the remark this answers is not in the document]
michelinewu: Done

app/services/dual-output/dual-output.ts
michelinewu: Thanks for catching! Done

blackxored (approved):

PR #4937, opened 2024-04-12
Fix Hide Viewer Count for Realm

gettinToasty:


blackxored (approved):

app/components-react/root/LiveDock.tsx
```diff
@@ -284,16 +273,21 @@ function LiveDock(p: { onLeft: boolean }) {
...
+      'currentViewers',
       'liveText',
       'isPopOutAllowed',
       'streamingStatus',
     ]),
   );
 
+  const liveDockSize = useRealmObject(Services.CustomizationService.state).livedockSize;
+  const collapsed = useRealmObject(Services.CustomizationService.state).livedockCollapsed;
+  const hideViewerCount = useRealmObject(Services.CustomizationService.state).hideViewerCount;
```
andycreeth, line 287: I think it's pretty clear that I should add some sort of selector API like we had for `useVuex` or similar to simplify this a bit.

michelinewu (approved):

PR #4942, opened 2024-04-15
add max title length to tiktok

trevorrl:
TikTok Live has a max title length of 32 characters. The API does not return an error, it simply cuts the title off at 32 characters, leading to unexpected behavior when a user uses a title longer than 32 characters. 

Currently this only works in Advanced Mode when using a custom title for TikTok, as I couldn't figure out how to make it work in simple mode / using same title across platforms. But also, 140 > 32 is such a stark drop in max characters that I felt like this might be too severe a change in behavior to do without consulting with maintainers.


gettinToasty (approved):

trevorrl:
@gettinToasty ty for approval, however I'm not sure how to get tests to actually run and it seems like I can't merge without that, so merging will have to be on one of ya'll.

PR #4946, opened 2024-04-16
fix(twitch): invalid tags prevent channel update, retry with valid

blackxored:


gettinToasty (changes requested):
Looks like there is a strict nulls regression but otherwise LGTM

michelinewu (approved):

gettinToasty (approved):

PR #4979, opened 2024-04-29
Tech debt: get base resolutions from video settings service.

michelinewu:


gettinToasty (approved):
nice super happy to see these consolidated

PR #4985, opened 2024-05-01
fix(recent-events): tooltips have been missing since port

blackxored:
Recent Events aka Mini Feed tooltips for the toolbar controls have possibly been missing since the React port. They were using `v-tooltip` and not being displayed.


gettinToasty (approved):
ty for handling, i noticed that myself but never got around to it

PR #5015, opened 2024-05-28
add twitch import button to scene collections window

andycreeth:


blackxored (approved):
Translations are not in the English file, it would seem. Otherwise, looks good.

PR #5028, opened 2024-06-05
fix(auth): SLID onboarding with YT/FB results in custom stream settings

blackxored:


andycreeth (approved):
Seems fairly harmless I think

PR #5034, opened 2024-06-07
Better unknown error handling and diag report.

michelinewu:


blackxored:
This would overwrite the custom error for invalid recording path, so we should look to bring that back in before merging.

PR #5064, opened 2024-07-10
Fix TikTok test.

michelinewu:


app/services/platforms/tiktok/api.ts
```diff
@@ -25,7 +25,7 @@ export enum ETikTokLiveScopeReason {
   APPROVED_OBS = 2,
 }
 
-export type TTikTokLiveScopeTypes = 'approved' | 'not-approved' | 'legacy' | 'denied';
+export type TikTokLiveScopeTypes = 'approved' | 'not-approved' | 'legacy' | 'denied';
```
gettinToasty, line 28: super nitpicky but i don't think this was a typo since `T` prefix indicates a TS type (as opposed to a class or something), although this does also have `Types` in the name 🤔

gettinToasty (approved):

PR #5068, opened 2024-07-12
Fix set settings and node validation order.

michelinewu:


app/services/dual-output/dual-output.ts
```diff
@@ -449,20 +429,24 @@ export class DualOutputService extends PersistentStatefulService<IDualOutputServ
...
-    this.scenesService.views.scenes.forEach(scene => {
-      this.createPartnerNodes(scene.id);
-    });
+    try {
+      // convert all scenes in the single output collection to dual output
+      this.scenesService.views.scenes.forEach(scene => {
+        this.createPartnerNodes(scene.id);
+      });
+    } catch (error: unknown) {
+      console.error('Error converting to single output collection to dual output: ', error);
```
gettinToasty, line 444: it feels like there should maybe be some sort of cleanup here to safely return the scene collection to a state of single output, but maybe that out of scope for this PR

app/services/dual-output/dual-output.ts
```diff
@@ -626,6 +619,8 @@ export class DualOutputService extends PersistentStatefulService<IDualOutputServ
           corruptedNodeIds.add(corruptedNode.id);
```
gettinToasty, line 619: i think i missed this before but since `validateSource` always returns the `partnerNode` doesn't that mean every node will have its partner marked as a corrupted node?

gettinToasty (approved):

app/services/dual-output/dual-output.ts
michelinewu: validate source only repairs and returns the node if the scene items do not have the same source

PR #5070, opened 2024-07-17
Add dual output toggle to the source selector bar.

michelinewu:


app/components-react/editor/elements/SourceSelector.tsx
```diff
@@ -632,6 +648,73 @@ class SourceSelectorController {
...
+              });
+            }
+            dualOutputToggleHandled.unsubscribe();
+          },
+        );
+
+        // only open video settings when toggling on dual output
+        const skipShowVideoSettings = this.dualOutputService.views.dualOutputMode === true;
+
+        this.dualOutputService.actions.setdualOutputMode(
```
blackxored, line 683: Cosmetic: camel case this method

app/components-react/editor/elements/SourceSelector.tsx
```diff
@@ -632,6 +648,73 @@ class SourceSelectorController {
...
+        });
+      } else if (Services.TransitionsService.views.studioMode) {
+        message.error({
+          content: $t('Cannot toggle Dual Output while in Studio Mode.'),
+          className: styles.toggleError,
+        });
+      } else {
+        const dualOutputToggleHandled = this.dualOutputService.dualOutputToggleHandled.subscribe(
+          (enabled: boolean) => {
+            console.log('enabled', enabled);
```
blackxored, line 666: dangling console, if we wanna keep it I think we should add some context as to what's enabled

app/components-react/editor/elements/SourceSelector.tsx
```diff
@@ -590,9 +597,18 @@ class SourceSelectorController {
     this.streamingService.actions.setSelectiveRecording(
       !this.streamingService.state.selectiveRecording,
     );
-    if (!this.selectiveRecordingEnabled && this.isDualOutputActive) {
-      this.dualOutputService.actions.toggleDisplay(false, 'vertical');
+    if (this.isDualOutputActive) {
+      this.dualOutputService.actions.toggleDisplay(this.selectiveRecordingEnabled, 'vertical');
```
blackxored, line 601: Had a bit of a hard time understanding this one, should we extract a variable even if it's just equal to `selectiveRecordingEnabled` to clarify what we're doing here? i.e `shouldToggleDisplay` or something similar, I'm not sure.

app/components-react/editor/elements/SourceSelector.tsx
```diff
@@ -590,9 +597,18 @@ class SourceSelectorController {
...
       this.selectionService.views.globalSelection.filterDualOutputNodes();
+
+      if (!this.selectiveRecordingEnabled) {
```
blackxored, line 604: We would use the previous one here too.

app/components-react/editor/elements/SourceSelector.tsx
```diff
@@ -632,6 +648,73 @@ class SourceSelectorController {
...
+      } else {
+        const dualOutputToggleHandled = this.dualOutputService.dualOutputToggleHandled.subscribe(
+          (enabled: boolean) => {
+            console.log('enabled', enabled);
+            if (enabled && this.selectiveRecordingEnabled) {
+              // show warning message if selective recording is active
+              remote.dialog.showMessageBox({
+                title: 'Vertical Display Disabled',
+                message: $t(
+                  'Dual Output can’t be displayed - Selective Recording only works with horizontal sources and disables editing the vertical output scene. Please disable selective recording from Sources to set up Dual Output.',
```
blackxored, line 672: Not sure about this message, it looks like it says we couldn't enable dual output but this is only on a subscription, we still proceed to enable below? Most definitely just me not understanding the logic, but wanted to make sure.

app/components-react/windows/settings/Video.tsx
```diff
@@ -473,27 +478,41 @@ class VideoSettingsModule {
...
+    ) {
+      // show warning message if selective recording is active
+      remote.dialog.showMessageBox({
+        title: 'Vertical Display Disabled',
+        message: $t(
+          'Dual Output can’t be displayed - Selective Recording only works with horizontal sources and disables editing the vertical output scene. Please disable selective recording from Sources to set up Dual Output.',
+        ),
+      });
     } else {
+      const dualOutputToggleHandled = this.dualOutputService.dualOutputToggleHandled.subscribe(
```
blackxored, line 493: Same question as before.

blackxored (approved):
Looks good, just a few minor questions trying to understand it better. We might be missing translations since I believe I've seen more new strings than on the translation file diff.

PR #5084, opened 2024-08-02
Fix typo and conditional logic for apply prompt.

michelinewu:


gettinToasty (approved):
looks like there is a bunch of commented out code left behind but otherwise looks good if that is meant to be re-implemented at a later date

michelinewu:
This is intentional - we're waiting on a change that would give us the value needed to conditionally show this text

PR #5085, opened 2024-08-06
Add TikTok replay noty.

michelinewu:


app/services/platforms/tiktok.ts
```diff
@@ -1,4 +1,4 @@
-import { InheritMutations, Inject, mutation } from '../core';
+import { InheritMutations, Inject, mutation, Service } from '../core';
```
gettinToasty, line 1: i think this is an accidental import?

gettinToasty (approved):
LGTM!

PR #5091, opened 2024-08-13
Platform link update

cmm21:


app/components-react/pages/BrowseOverlays.tsx
```diff
@@ -57,6 +58,12 @@ export default function BrowseOverlays(p: {
 
       return { action: 'deny' };
     });
+
+    view.webContents.on('did-finish-load', () => {
+      if (Utils.isDevMode()) {
+        view.webContents.openDevTools();
```
avacreeth, line 64: Ty for adding this!  It's a pain always manually doing this change in dev mode.

avacreeth (approved):
Look good! 👍

PR #5134, opened 2024-09-21
Get rid of useModules in volmeters

holiber:


app/components-react/editor/elements/mixer/GLVolmeters.tsx
```diff
@@ -133,12 +138,6 @@ class GLVolmetersModule {
     });
   }
 
-  // update volmeters subscriptions when audio sources change
-  watchAudioSources = injectWatch(
-    () => this.audioSources,
-    () => this.subscribeVolmeters(),
-  );
-
```
gettinToasty, line 141: I don't think we can simply remove this watcher? We still need the volmeters to update when the audio sources change, unless I'm missing something here.

gettinToasty (changes requested):

gettinToasty (approved):

holiber:
Closed in favor of https://github.com/stream-labs/desktop/pull/5139

PR #5135, opened 2024-09-24
feat: add collectibles API for adding source inside embed

blackxored:
Adds guest API methods for listing scenes and adding a collectible to a specific scene. This is convenience for create source depending on its type.

### DIFF
--- a/app/components-react/pages/BrowseOverlays.tsx
+++ b/app/components-react/pages/BrowseOverlays.tsx
@@ -130,8 +130,10 @@ export default function BrowseOverlays(p: {
     }
 
     if (
-      !assetURL.startsWith('https://cdn.streamlabs.com/marketplace') ||
-      !assetURL.startsWith('https://streamlabs-marketplace-staging.streamlabs.com')
+      !(
+        assetURL.startsWith('https://cdn.streamlabs.com/marketplace') ||
+        assetURL.startsWith('https://streamlabs-marketplace-staging.streamlabs.com')
+      )
     ) {
       throw new Error('Invalid asset URL');
     }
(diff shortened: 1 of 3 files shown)


app/components-react/pages/BrowseOverlays.tsx
```diff
@@ -85,11 +95,65 @@ export default function BrowseOverlays(p: {
...
+    }
+
+    const sourceSettings =
+      type === 'video' ? { is_local_file: false, input: assetURL } : { file: assetURL };
+
+    // TODO: find or create enum
+    const sourceType = type === 'video' ? 'ffmpeg_source' : 'image_source';
+
+    // TODO: do we want the caller to provide name?
+    const sourceName = SourcesService.views.suggestName('Collectible');
```
gettinToasty, line 142: I think the asset will probably have a name in the library, so maybe we should open that up as an argument?

app/services/scenes/scenes.ts
```diff
@@ -266,6 +266,16 @@ class ScenesViews extends ViewHandler<IScenesState> {
...
+
+  createAndAddSource(
+    sceneId: string,
+    name: string,
+    sourceType: TSourceType,
+    settings: Dictionary<any>,
+  ) {
+    const scene = this.getScene(sceneId);
+    return scene?.createAndAddSource(name, sourceType, settings).sceneItemId;
+  }
```
gettinToasty, line 278: we really shouldn't be altering state in a View model, that's very much an antipattern, but is there a specific reason you did it here?

gettinToasty (changes requested):

app/components-react/pages/BrowseOverlays.tsx
blackxored: Commenting again since it got lost: we eventually ended up needing to pass from caller yes.

app/services/scenes/scenes.ts
blackxored: This is now using the Scenes Service, is still an indirection to the view since we seem to need the `#createAndAddSource` from the scene object returned by that view's `getScene`. Is only for convenience though.

michelinewu (approved):

michelinewu (approved):

michelinewu (approved):

PR #5143, opened 2024-09-27
Get rid of useModule in Appearence settings

holiber:


gettinToasty (approved):

gettinToasty:
Looks like there's an extra newline being caught by the linter, but should be good to go after fixing

PR #5144, opened 2024-09-28
Hide invisible audio sources from the audio mixer

holiber:


holiber:
Hmm. This PR breaks the test "Nested scenes should provide audio sources to mixer"
There should be another more complex algorithm for getting the list of visible sources

PR #5167, opened 2024-10-17
Recording History QOL

gettinToasty:
- Adds remove function for unwanted entries
- Condenses edit functions into a modal


holiber (approved):

blackxored (approved):

blackxored:
Looks good. Has some missing translations if we could add?

PR #5196, opened 2024-10-30
feat(widgets): new ChatBox API and first setting, details follow

blackxored:
* Incremental rollout check for `core--widgets-v2--chat-box` that should indicate whether the user has access to the new ChatBox features.
* Conditional logic on the widget system to use new widget APIs when `useNewWidgetConfig` is set to true.
* Simplest way of integrating the new API by patch the response and requests to the `global` object.
* One setting to test at the beginning, "Enable Chat Notifications", which triggers an alert when there's a new chat message.
* Few hacks, this is experimental and should be generalized.

### DIFF
--- a/app/components-react/widgets/ChatBox.tsx
+++ b/app/components-react/widgets/ChatBox.tsx
@@ -45,7 +45,7 @@ export function ChatBox() {
 
 export class ChatBoxModule extends WidgetModule<IChatBoxState> {
   get meta() {
-    return {
+    const result = {
       theme: metadata.list({
         label: $t('Theme'),
         options: [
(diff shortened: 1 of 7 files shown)


app/components-react/widgets/ChatBox.tsx
```diff
@@ -111,15 +111,44 @@ export class ChatBoxModule extends WidgetModule<IChatBoxState> {
...
   protected patchBeforeSend(settings: IChatBoxState['data']['settings']) {
-    return {
+    const obj = {
       ...settings,
       message_hide_delay: Math.floor(settings.message_hide_delay / 1000),
       message_show_delay: Math.floor(settings.message_show_delay / 1000),
     };
+
+    if (this.config.useNewWidgetAPI) {
+      return { global: obj };
```
gettinToasty, line 150: do we no longer need to do math on the message delays in the new api?

gettinToasty (approved):

app/components-react/widgets/ChatBox.tsx
blackxored: We're passing the same object, in which we already did the conversion. As far as I know that is still required.

gettinToasty: good point, but it does look like we're not returning the obj in case the flag is false

blackxored: Fixed in #5197, might be dead code now.

PR #5201, opened 2024-11-01
fix(widgets): custom HTML/CSS et al bypasses widget overrides

blackxored:


app/components-react/widgets/ChatBox.tsx
```diff
@@ -147,7 +136,7 @@ export class ChatBoxModule extends WidgetModule<IChatBoxState> {
     };
 
     if (this.config.useNewWidgetAPI) {
-      return { global: obj };
+      return settingsToGlobal(obj);
```
gettinToasty, line 139: i think you could avoid this additional check by doing `super.patchBeforeSend(obj)` here but thats pretty nitpicky

gettinToasty (approved):

PR #5202, opened 2024-11-01
fix(widgets): custom code in widgets have no defaults so they break

blackxored:
Widgets can get into an invalid state when enabling custom code, no defaults were consistently provided by backend and, as a result, enabling the Custom Code option could result in empty widget HTML, CSS and JS, thus breaking the widget.
In some cases, users could recover by going through the dashboard and choosing to "Reset Defaults", but this change makes that invalid state impossible.
We fetch custom code defaults through the new static config API at `/api/v5/widgets/static/config/{widgetType}` and populate those fields if they're empty on widget settings load. 
We needed to add a new `humanType` field which is the widget type in string representation as opposed to the number enum so we can pass it to that endpoint.
Also updates the `custom_defaults` property that is supposed to be used to provide that reset functionality on Desktop, though is inconsistent where we've seen this button be available.
Backend is working towards a complete fix which will return the default code in the standard settings endpoints if they're empty to make sure it never breaks again, and also that users can't end up in invalid state by resetting them through the web UI.

### DIFF
--- a/app/services/widgets/settings/widget-settings.ts
+++ b/app/services/widgets/settings/widget-settings.ts
@@ -11,6 +11,7 @@ import {
   IWidgetSettingsServiceApi,
   IWidgetSettingsState,
   TWIdgetLoadingState,
+  WidgetDefinitions,
   WidgetsService,
 } from 'services/widgets';
 import { Subject } from 'rxjs';
@@ -23,6 +24,7 @@ export const WIDGET_INITIAL_STATE: IWidgetSettingsGenericState = {
   data: null,
   rawData: null,
   pendingRequests: 0,
+  staticConfig: null,
 };
 
 export type THttpMethod = 'GET' | 'POST' | 'PUT' | 'DELETE';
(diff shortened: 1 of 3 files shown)


holiber (commented):
Thanks for fixing that! 😊

I'm not totally sold on humanType—it feels a bit off for this context. What do you think about using apiType instead? It seems more fitting.

By the way, could you check if we even need this new field? We've already got the humps library for transforming enum keys. I made a quick example you can check out here:
https://stackblitz.com/edit/humps-enums?file=src%2FApp.tsx

blackxored:
I've updated the PR description, hopefully I didn't miss anything *too* important. I learned about `humps` today, nice library, unfortunately it wouldn't work for our case since consistency is an issue :D we have widgets that use the same widget type, and others that significantly deviate from how we name the enum and their snake case conversion, one even uses dash-case, making anything dynamic pretty much out of the question. I would love to use it, unfortunately the backend prevents us from it. As for `humanType` I'm not too strongly attached to the name, but wanted to make something new that wouldn't have implicit/overloaded meaning when looking at widget interactions elsewhere, we do have `apiType` and `type` in some other areas so I didn't want to confuse people.

gettinToasty (approved):

blackxored:
### Merge activity

* **Nov 8, 5:02 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/stream-labs/desktop/5202).

PR #5209, opened 2024-11-07
feat(ultra): dynamic info

blackxored:


app/services/user/index.ts
```diff
@@ -769,6 +803,32 @@ export class UserService extends PersistentStatefulService<IUserServiceState> {
       .catch(() => null);
   }
 
+  // TODO: why isn't this on the endpoint above, we have to do an extra request
+  async setDisplayUltraPrices() {
+    const host = this.hostsService.streamlabs;
+    // TODO: hardcoded currency, why can't backend infer
+    const url = 'https://api-id.streamlabs.dev/v1/subscriptions/plans?currency=USD';
```
davidliang781205, line 810: We use CF header and the default currency is USD.  My bad - copied and pasted too fast as this was fired from core with user preference on currency

PR #5216, opened 2024-11-14
fix(widgets): infinite loading in widget settings

blackxored:
Infinite loading on widget settings could happen by doing by following:

1. Opening a React widget settings page
2. Opening the custom code window from within that page.
3. Closing the custom code window.

After the custom code window closes, the widget settings are unable to load again
until the app is restarted.

This is caused by a combination of using the `beforeunload` handler, without taking
into account window IDs (and custom code editor is a one-off window) to execute
code conditionally, (this handler might be called for *any* window closing,
need to confirm due to the weird behavior following) and the fact that the
`WidgetModule`'s code is run again in that window (presumably because
of the `injectChild` call it uses). This is particularly relevant with the `init`
method which registers preview source, said handler, fetches data, and more.

As a result, we ended up registering two handlers that tried to cleanup the
widget preview source, with the second one (the widget settings if you closed
the custom code window first) unable to succeed. We were also attempting to
create that preview source on the custom code one-off window, which is
unnecessary and might've contributed to the issue, since only one preview
source per widget is allowed, and the code throws if we try to.

Also indirectly fixed is the `destroy` method of such module, while `cancelUnload` might
trigger without a `destroy` (if I understand correctly for window refresh in
development mode), calling both at once with only one having a null-check might
throw, and that is exactly what `destroy` did. The fact this code is ran as the
window was closing might've obscured that this could happen.

### DIFF
--- a/app/components-react/widgets/common/useWidget.tsx
+++ b/app/components-react/widgets/common/useWidget.tsx
@@ -94,13 +94,46 @@ export class WidgetModule<TWidgetState extends IWidgetState = IWidgetState> {
     const widget = this.widget;
     this.setBrowserSourceProps(widget.getSource()!.getPropertiesFormData());
 
-    // create a temporary preview-source for the Display component
-    if (this.state.shouldCreatePreviewSource) {
-      const previewSource = widget.createPreviewSource();
-      this.state.previewSourceId = previewSource.sourceId;
-    }
+    /* FIXME: we need a more robust way of handling these `onUnload` invocations,
+     *
+     * These are supposed to fire for any window that gets closed, not just
+     * the window that is displaying this component, that includes one-off windows.
+     * As a result, we destroyed the preview source when a child window such as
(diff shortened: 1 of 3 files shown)


blackxored:
* **#5216** <a href="https://app.graphite.dev/github/pr/stream-labs/desktop/5216?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/stream-labs/desktop/5216?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`

<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

michelinewu (approved):

PR #5241, opened 2024-12-05
feat: AI Highlighter

marvinoffers:


app/components-react/pages/RecordingHistory.tsx
```diff
@@ -53,6 +60,11 @@ class RecordingHistoryController {
 
   get uploadOptions() {
     const opts = [
+      {
+        label: `${$t('AI Highlighter')} ${$t('For Fortnite streams (Beta)')}`,
```
gettinToasty, line 64: im not sure i understand why these strings are separated?

app/components-react/windows/go-live/AiHighlighterToggle.tsx
```diff
@@ -0,0 +1,144 @@
...
+                      style={{
+                        width: '182px',
+                        height: '187px',
+                        display: 'grid',
+                        placeContent: 'center',
+                        backgroundImage:
+                          'url(https://slobs-cdn.streamlabs.com/media/highlighter-image.png)',
+                        backgroundPosition: 'center',
+                        backgroundSize: 'contain',
+                        backgroundRepeat: 'no-repeat',
```
gettinToasty, line 128: lots of styles that should be classes here

app/services/highlighter/ai-highlighter/ai-highlighter.ts
```diff
@@ -47,3 +48,208 @@ export interface IHighlighterMessage {
...
 }
+interface IHighlighterProgressMessage {
+  progress: number;
+}
+
+// Buffer management class to handle split messages
+class MessageBufferHandler {
+  private buffer: string = '';
+  private readonly startToken: string = '>>>>';
+  private readonly endToken: string = '<<<<';
```
gettinToasty, line 59: i feel like these should probably be constants defined outside of the class but its not a huge deal

app/services/highlighter/ai-highlighter/ai-highlighter.ts
```diff
@@ -47,3 +48,208 @@ export interface IHighlighterMessage {
...
+              }
+              resolve(aiHighlighterMessage.json as IHighlight[]);
+              break;
+            case 'highlights_partial':
+              // Handle partial highlights if needed
+              break;
+            default:
+              console.log('\n\n');
+              console.log('Unrecognized message type:', aiHighlighterMessage);
+              console.log('\n\n');
```
gettinToasty, line 156: are these just for debugging?

app/services/highlighter/ai-highlighter/ai-highlighter.ts
```diff
@@ -47,3 +48,208 @@ export interface IHighlighterMessage {
...
+    childProcess.on('exit', (code, signal) => {
+      messageBuffer.clear();
+      reject(new Error(`Child process exited with code ${code} and signal ${signal}`));
+    });
+  });
+}
+
+function parseAiHighlighterMessage(messageString: string): IHighlighterMessage | string | null {
+  try {
+    if (messageString.includes('>>>>') && messageString.includes('<<<<')) {
```
gettinToasty, line 181: yea this is a good reason to use constants as i mentioned above

app/services/highlighter/index.ts
```diff
@@ -651,16 +805,62 @@ export class HighlighterService extends PersistentStatefulService<IHighligherSta
...
+
+      //  await this.addStream({
+      //  id: 'demo123',
+      // game: 'Fortnite',
+      //title: 'Demo Stream',
+      //date: '1726234396290',
+      //state: { type: 'detection-finished', progress: 0 },
+      //});
+      //const newClips = [getSharedResource('replay123.mp4')].map(path => ({ path }));
+      //this.addClips(newClips, 'demo123', 'Manual');
```
gettinToasty, line 817: presumably to be removed?

gettinToasty (changes requested):

app/components-react/highlighter/ClipPreview.tsx
```diff
@@ -120,3 +149,24 @@ export function formatSecondsToHMS(seconds: number): string {
...
+function FlameHypeScore({ score }: { score: number }) {
+  if (score === undefined) {
+    return <></>;
+  }
+  const normalizedScore = Math.min(1, Math.max(0, score));
+  const fullFlames = Math.ceil(normalizedScore * 5);
+
+  return (
+    <div className="flex items-center gap-1" style={{ fontSize: '19px' }}>
+      {[...Array(fullFlames)].map((_, index) => (
```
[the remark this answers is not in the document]
marvinoffers: Also had to check again. For this case I need to do that: 
eg: Array(3) would create an array with 3 empty slots. map() would skip all of them. 
If i spread [...Array(3)] map is working again.  

// This won't work - map skips empty slots
Array(3).map(x => console.log(x))  // Nothing happens

[...Array(3)].map(x => console.log(x))  // Logs undefined three time


Still decided to change the code to this: Array.from({ length: fullFlames }).map...

app/components-react/highlighter/ClipsView.tsx
```diff
@@ -33,7 +37,10 @@ export default function ClipsView({
   props: IClipsViewProps;
   emitSetView: (data: IViewState) => void;
 }) {
-  const { HighlighterService, UsageStatisticsService } = Services;
+  const { HighlighterService, UsageStatisticsService, IncrementalRolloutService } = Services;
+  const aiHighlighterEnabled = IncrementalRolloutService.views.featureIsEnabled(
+    EAvailableFeatures.aiHighlighter,
+  );
```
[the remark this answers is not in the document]
marvinoffers: yeah, checked a few of the other sections where the feature flag was enabled and most times it was implemented in a similar manner. so thing its fine.

app/components-react/pages/RecordingHistory.tsx
marvinoffers: Thought it would make sense since both strings already existed on its own and i wanted to prevent duplicates. Merged to one string.

app/components-react/highlighter/StreamCard.tsx
```diff
@@ -0,0 +1,365 @@
...
+          clip.aiInfo.inputs = (clip.aiInfo as any).moments;
+        }
+        return clip;
+      }),
+  );
+  const stream = useVuex(() =>
+    HighlighterService.views.highlightedStreams.find(s => s.id === streamId),
+  );
+  if (!stream) {
+    return <>error</>;
```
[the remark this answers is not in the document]
marvinoffers:
```
  const stream = useVuex(() =>
    HighlighterService.views.highlightedStreams.find(s => s.id === streamId),
  );
  if (!stream) {
    return <>error</>;
  }
```

Its only there because the find() could technically return undefined - wich should never happen.
Are you thinking to make the error case "nicer" or check it differently?

app/components-react/highlighter/InputEmojiSection.m.less
```diff
@@ -0,0 +1,3 @@
+.description {
+  text-wrap: nowrap;
+}
```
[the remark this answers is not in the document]
marvinoffers: Hmm, there is no perfect spot beside an extra file. Inline-css is not working bcs text-wrap isnt suppoerted by react inline css. 

There will probably be more css in the future, so will keep for now

app/components-react/highlighter/StreamCard.tsx
```diff
@@ -0,0 +1,365 @@
...
+export const PlayButton = () => (
+  <svg width="38" height="38" viewBox="0 0 38 38" fill="none" xmlns="http://www.w3.org/2000/svg">
+    <path
+      fillRule="evenodd"
+      clipRule="evenodd"
+      d="M31.3111 17.05L12.9395 4.36284C11.6534 3.45661 10 4.36284 10 5.8128V31.1872C10 32.6372 11.6534 33.5434 12.9395 32.6372L31.3111 19.95C32.2296 19.225 32.2296 17.775 31.3111 17.05"
+      fill="white"
+    />
+  </svg>
+);
```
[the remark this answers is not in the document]
marvinoffers: All the beaker play pause buttons didn't match the style we were looking for. Just messaged Christina to check if we could implement our custom icons into beaker. Will implement the new icons as soon as we have them in beaker or replace with the ones we have already

app/components-react/highlighter/PreviewModal.tsx
```diff
@@ -11,76 +13,294 @@ export default function PreviewModal({
...
+      videoPlayer.current?.removeEventListener('ended', handleEnded);
+      videoPlayer.current?.removeEventListener('play', handlePlay);
+      videoPlayer.current?.removeEventListener('pause', handlePause);
+      if (audio.current) {
+        audio.current.pause();
+        audio.current.removeEventListener('ended', handleAudioEnd);
+        audio.current = null;
+      }
+    };
+  }, [playlist.length]);
```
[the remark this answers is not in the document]
marvinoffers: Not sure exactly what you mean. Could you explain a little further?

app/components-react/highlighter/HighlightGenerator.m.less
```diff
@@ -0,0 +1,50 @@
...
+}
+.option {
+  width: 100%;
+  padding: 8px 12px;
+  border-radius: 8px;
+  color: white;
+  opacity: 1;
+}
+.option:hover {
+  background-color: #4f5e65;
```
marvinoffers, line 22: # #

app/components-react/highlighter/StreamCard.tsx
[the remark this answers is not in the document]
gettinToasty: i see, in that case i would probably just recommend returning `<></>` since a blank screen is probably better than just some plaintext in terms of user visibility, the other option would be to pass through the whole stream as a prop instead of just the streamId

app/components-react/highlighter/PreviewModal.tsx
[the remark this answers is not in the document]
gettinToasty: yea, it'd look something like this
```
useEffect(() => {
if (!videoPlayer.current) return;

// existing functionality

}, [playlist.length, videoPlayer.current]);
```

app/components-react/highlighter/StreamCard.tsx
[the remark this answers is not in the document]
gettinToasty: ah ok if they arent in beacker yet its not a big deal :)

app/components-react/pages/RecordingHistory.tsx
gettinToasty: got it, yea like i mentioned before sentence structure might change for different languages to this is another place where we want to avoid concatting translations, even if it causes some duplication

app/components-react/windows/go-live/AiHighlighterToggle.tsx
```diff
@@ -67,68 +69,24 @@ export default function AiHighlighterToggle({
...
-                      <SwitchInput
-                        style={{ padding: 0, margin: 0, marginLeft: '-40px', width: '100%' }}
-                        value={useHighlighter}
-                        label=""
-                        onChange={() => HighlighterService.actions.toggleAiHighlighter()}
-                      />
+                <div className={styles.expandedWrapper}>
+                  <div className={styles.toggleTextWrapper}>
+                    <div>
+                      {' '}
```
gettinToasty, line 75: is this just a <br>?

gettinToasty (approved):

PR #5252, opened 2024-12-11
disable button if detection in progress

marvinoffers:


marvinoffers:
Right now it is possible to start multiple ai-detections via the recordings tab. This PR will prevent that.

gettinToasty (approved):

PR #5280, opened 2025-01-08
VCam custom sources settings using Realm

avoitenko-logitech:


app/services/virtual-webcam.ts
```diff
@@ -27,9 +29,26 @@ interface IVirtualWebcamServiceState {
...
+    properties: {
+      outputType: { type: 'int', default: VCamOutputType.ProgramView },
+      outputSelection: { type: 'string', default: '' },
+    },
+  };
+}
+
+VirtualCamServicePersistentState.register({ persist: true });
+
 export class VirtualWebcamService extends StatefulService<IVirtualWebcamServiceState> {
```
michelinewu, line 48: This `StatefulService` can be converted to just a `Service` now

app/services/virtual-webcam.ts
```diff
@@ -27,9 +29,26 @@ interface IVirtualWebcamServiceState {
...
   @Inject() usageStatisticsService: UsageStatisticsService;
   @Inject() sourcesService: SourcesService;
+  virtualCamSettings = VirtualCamServicePersistentState.inject();
 
   static initialState: IVirtualWebcamServiceState = { running: false };
```
michelinewu, line 53: Because the realm has been injected, this `initialState` can be removed

app/services/virtual-webcam.ts
```diff
@@ -27,9 +29,26 @@ interface IVirtualWebcamServiceState {
...
+    },
+  };
+}
+
+VirtualCamServicePersistentState.register({ persist: true });
+
 export class VirtualWebcamService extends StatefulService<IVirtualWebcamServiceState> {
   @Inject() usageStatisticsService: UsageStatisticsService;
   @Inject() sourcesService: SourcesService;
+  virtualCamSettings = VirtualCamServicePersistentState.inject();
```
michelinewu, line 51: For consistency with other services, considering renaming `virtualCamSettings` to `state`

app/services/virtual-webcam.ts
```diff
@@ -115,9 +134,23 @@ export class VirtualWebcamService extends StatefulService<IVirtualWebcamServiceS
...
+    this.virtualCamSettings.db.write(() => {
+      this.virtualCamSettings.deepPatch({
+        outputType: type,
+        outputSelection: name,
+      });
+    });
     obs.NodeObs.OBS_service_updateVirtualCam(type, name);
   }
 
+  outputType(): VCamOutputType {
```
michelinewu, line 146: Not sure if you would like this be reactive? If so, we use `get outputType()`, but the function call also works if you don't want reactivity

app/services/virtual-webcam.ts
```diff
@@ -115,9 +134,23 @@ export class VirtualWebcamService extends StatefulService<IVirtualWebcamServiceS
...
+    return this.virtualCamSettings.outputType;
+  }
+
+  outputSelection(): string {
```
michelinewu, line 150: Same as above

app/components/windows/settings/VirtualWebcamSettings.tsx
```diff
@@ -192,51 +188,40 @@ export default class VirtualCamSettings extends Vue {
...
       }
 
-      //this.settingsService.setSettingValue('Virtual Webcam', 'OutputSelection', this.outputSelectionValue.id);
       this.virtualWebcamService.update(VCamOutputType.SceneOutput, this.outputSelectionValue.id);
     } else if (value.id == VCamOutputType.SourceOutput) {
       const sources = this.virtualWebcamService.getVideoSources().map(source => ({name: source.name, id: source.sourceId}));
 
       this.outputSelectionOptions = sources;
-      const outputSelectionIndex = this.outputSelectionOptions.findIndex(val => val.id === "" /*settingsOutputSelection*/);
+      const outputSelectionIndex = this.outputSelectionOptions.findIndex(val => val.id === this.virtualWebcamService.outputSelection());
```
michelinewu, line 210: See the note in the virtual camera service regarding reactivity for getters.

app/components/windows/settings/VirtualWebcamSettings.tsx
```diff
@@ -192,51 +188,40 @@ export default class VirtualCamSettings extends Vue {
...
-
     if (value.id == VCamOutputType.SceneOutput) {
       const scenes = this.scenesService.views.scenes.map((scene) => ({
         name: scene.name,
         id: scene.id,
       }));
 
       this.outputSelectionOptions = scenes;
-      const outputSelectionIndex = this.outputSelectionOptions.findIndex(val => val.id === "" /*settingsOutputSelection*/);
+      const outputSelectionIndex = this.outputSelectionOptions.findIndex(val => val.id === this.virtualWebcamService.outputSelection());
```
michelinewu, line 198: See the note in the virtual camera service regarding reactivity for getters.

michelinewu (changes requested):

app/services/virtual-webcam.ts
avoitenko-logitech: As we extend VirtualWebcamService from the StatefulService, it already has a field named `state`. That's why if I introduce another field names `state`, I'll have a name clash. And this (already existing) state should not be stored permanently, it is not related to the settings. Currently the interface `IVirtualWebcamServiceState` only has a single field `running` which allows to know is virtual camera is ON.

gettinToasty: yea so part of moving to Realm is getting rid of the concept of StatefulService, as Micheline mentioned in her other comment, but that means we can get rid of the currently existing `state` object as well, and we can use Realm's ephemeral state to store things which aren't persisted, example is in `services/announcements.ts`

michelinewu (approved):

PR #5281, opened 2025-01-09
Refactor for flow

michelinewu:


app/services/platforms/kick.ts
```diff
@@ -39,89 +66,259 @@ export class KickService
...
-
-    this.userService.userLogout.subscribe(() => {
-      this.updateSettings({ title: this.state.settings.title, streamUrl: '', streamKey: '' });
-    });
+  /**
+   * Kick's API currently does not provide viewer count.
+   * To prevent errors, return 0 for now;
+   */
+  get viewersCount(): number {
+    return 0;
```
gettinToasty, line 103: should we hide the viewercount from livedock if they're on kick only maybe?

gettinToasty (approved):

app/services/platforms/kick.ts
michelinewu: Good call

PR #5286, opened 2025-01-22
[Highlighter] Restructured highlighter service and related files

jankalthoefer:


app/services/highlighter/models/rendering.models.ts
```diff
@@ -0,0 +1,173 @@
...
+
+// Avoid having to loop over every time for fast access
+export const transitionParams: {
+  [type in TTransitionType]?: { [key: string]: any };
+} = AVAILABLE_TRANSITIONS.reduce((params, transition) => {
+  return {
+    ...params,
+    [transition.type]: transition.params,
+  };
+}, {});
```
gettinToasty, line 173: just a note but the organizational pattern we typically use for large interface definition files is `[service]-api.ts` but i also dont hate this `models` syntax either i do like the additional separation it gives so not going to mark for change for now-- desktop team can discuss which we like more before deciding :)

app/services/highlighter/rendering/start-rendering.ts
```diff
@@ -0,0 +1,241 @@
...
+    } else {
+      setExportInfo({ error: $t('An error occurred while exporting the video') });
+
+      recordAnalyticsEvent(useAiHighlighter ? 'AIHighlighter' : 'Highlighter', {
+        type: 'ExportError',
+        error: 'Unknown',
+      });
+    }
+  } finally {
+    console.log('finally');
```
gettinToasty, line 230: console log + commented code

gettinToasty (approved):

app/services/highlighter/models/rendering.models.ts
jankalthoefer: Yes I am happy with what works for you guys. 
For me it just was important to split off the 400 line of type declarations from the service logic and ai-highlighter file itself.

PR #5298, opened 2025-02-06
Kick category required.

michelinewu:


app/services/platforms/kick.ts
```diff
@@ -325,6 +324,7 @@ export class KickService
     return jfetch<IKickStreamInfoResponse>(request)
       .then(async res => {
         const data = res as IKickStreamInfoResponse;
+        console.log('data', JSON.stringify(data, null, 2));
```
gettinToasty, line 327: got a console log here but can be cleaned up later

gettinToasty (approved):

PR #5299, opened 2025-02-06
Output settings form to react

michelinewu:


app/components-react/obs/ObsForm.tsx
```diff
@@ -216,21 +316,272 @@ export function ObsFormGroup(p: IObsFormGroupProps) {
...
+    </Collapse>
+  );
+}
+
+interface IObsInputListCustomInputProps {
+  inputProps: any;
+  options?: Omit<IObsListOption<unknown>, 'description'>[];
+}
+
+const ObsInputListCustomInput = forwardRef<{}, IObsInputListCustomInputProps>((p, ref) => {
```
gettinToasty, line 465: hmm im not sure this component should be named so generically when it has a lot of code dedicated to handling custom resolutions specifically

app/components-react/obs/ObsForm.tsx
```diff
@@ -216,21 +316,272 @@ export function ObsFormGroup(p: IObsFormGroupProps) {
...
+        style={{ width: '100%' }}
+      />
+    </InputWrapper>
+  );
+});
+
+/**
+ * @remark Note: The following is a copy of the shared component `TextInput` but with the reference forwarded.
+ * This is to allow form validation to work correctly.
+ */
```
gettinToasty, line 571: i wonder if it would make more sense to just forward the refs on these shared components?

app/components-react/shared/Tabs.tsx
```diff
@@ -36,11 +36,18 @@ export default function Tabs(p: ITabs) {
...
 
-  const data = p?.data ?? dualOutputData;
+  const data = p?.tabs ? formatTabs(p.tabs) : dualOutputData;
+
+  function formatTabs(tabs: string[]): ITab[] {
+    return tabs.map((tab: string) => ({
+      label: $t(tab),
+      key: tab,
+    }));
+  }
```
gettinToasty, line 46: is there a reason to restructure the props and format them here instead of just doing this formatting before passing in the tabs under the `data` prop?

app/components-react/windows/settings/Output.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { IObsFormType, ObsGenericSettingsForm } from './ObsSettings';
+import { useObsSettings } from './useObsSettings';
+
+export function OutputSettings() {
+  const { settingsFormData } = useObsSettings();
+
+  const type = settingsFormData[0].parameters[0].currentValue === 'Simple' ? 'collapsible' : 'tabs';
+
+  return <ObsGenericSettingsForm type={type as IObsFormType} />;
```
gettinToasty, line 10: looking at how light this component is i wonder if it would make more sense to do more of the output settings specific layout handling here with the tabs and collapsible instead of in `ObsGenericSettingsForm`, it feels a bit weird to me to put a lot of things specific to one settings page in a generic component

app/components-react/obs/ObsForm.tsx
michelinewu: In order to do that, I have to convert the shared component into a `forwardRef`. Because the `TextInput` and `NumberInput` shared components are used in so many other places, there is a risk that this could effect another form. So in order to do this, we'd have to test every form in the app, which we can do but would be a really large task.

app/components-react/shared/Tabs.tsx
michelinewu: This is how it was done in the vue component.

app/components-react/obs/ObsForm.tsx
michelinewu: Renamed

app/components-react/windows/settings/Output.tsx
michelinewu: Using this generic settings form and passing in the backend settings all at once means that the forms will automatically update on the frontend. This will be less frontend maintenance going forward when adding/updating features with settings because all we will need to do is update the realm and the fields will adjust accordingly. Moreover, adding form types to the generic settings form enables the flexibility with form layout without requiring a lot of work on the frontend since all that is needed is to pass in the type. That being said, I will refactor the `ObsTabbedFormGroup` to be less output settings specific.

gettinToasty (approved):

PR #5313, opened 2025-02-26
[Highlighter] - Educational UI

marvinoffers:


jankalthoefer (approved):

app/components-react/highlighter/EducationCarousel.m.less
```diff
@@ -0,0 +1,186 @@
+@paddingLeft: 24px; // was 24
+@paddingTop: 20px;
+@paddingRight: 24px;
+@paddingBottom: 20px;
+
+:global {
```
gettinToasty, line 6: so we don't want to have global unscoped changes in our modular css files-- when they get transpiled they'll insert global rules through the whole app. we should have some sort of `.carousel-container` to wrap these in

app/components-react/highlighter/StreamCard.tsx
```diff
@@ -49,12 +52,77 @@ export default function StreamCard({
...
+          <div style={{ display: 'flex', flexDirection: 'column', width: '280px' }}>
+            <h2>{$t('No clips found')}</h2>
+            <p style={{ marginBottom: '8px' }}>
+              {$t('Please make sure all the requirements are met:')}
+            </p>
+            <p>{$t('• Game is supported (Currently Fortnite only)')}</p>
+            <p>{$t('• Game language is English')}</p>
+            <p>{$t('• Map and Stats area is fully visible')}</p>
+            <p>{$t('• Game in fullscreen in your stream')}</p>
+            <p>{$t('• Game mode is supported (Battle Royale, Reload, Zero Build, OG)')}</p>
```
gettinToasty, line 92: i think here we would probably prefer to have a `<ul>` tag here just to keep translated strings to only what is required of them

gettinToasty (changes requested):

gettinToasty (approved):

PR #5336, opened 2025-03-25
Add Advanced Recording test.

michelinewu:


app/services/streaming/streaming.ts
```diff
@@ -1373,9 +1373,9 @@ export class StreamingService
           [OS.Mac]: filename,
           [OS.Windows]: filename.replace(/\//, '\\'),
         });
-        this.recordingModeService.actions.addRecordingEntry(parsedFilename);
-        this.markersService.actions.exportCsv(parsedFilename);
         this.recordingModeService.addRecordingEntry(parsedFilename);
+        // this.recordingModeService.actions.addRecordingEntry(parsedFilename);
```
gettinToasty, line 1377: is there a reason we aren't using the actions pattern here?

gettinToasty (changes requested):

app/services/streaming/streaming.ts
michelinewu: Updated

gettinToasty (approved):

PR #5337, opened 2025-03-25
[Highlighter] - updated publish flow

marvinoffers:


app/components-react/highlighter/Export/UploadProgress.tsx
```diff
@@ -0,0 +1,84 @@
...
+  platform: EUploadPlatform;
+  dense?: boolean;
+}) {
+  const { HighlighterService } = Services;
+  const { uploadInfo } = useVuex(() => ({
+    uploadInfo: HighlighterService.getUploadInfo(HighlighterService.views.uploadInfo, platform),
+  }));
+
+  if (uploadInfo === undefined) {
+    return <div>Something went wrong. Please contact support if the issue persists</div>;
```
gettinToasty, line 39: this needs to be translated

app/components-react/highlighter/Export/VideoPreview.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { Services } from 'components-react/service-provider';
+import { useVuex } from 'components-react/hooks';
+import styles from './ExportModal.m.less';
+import { $t } from 'services/i18n';
+
+export default function VideoPreview() {
+  const { HighlighterService } = Services;
+  const exportInfo = useRef(
+    HighlighterService.views.getCacheBustingUrl(HighlighterService.views.exportInfo.file),
+  );
```
gettinToasty, line 13: not sure you're why you're using a ref here if it's never being updated?

gettinToasty (changes requested):

app/components-react/highlighter/Export/VideoPreview.tsx
marvinoffers: if i don't use useRef the video will reload every time a re-render occurs. That is due to the old vuex logic. Need to refactor that at some point.

gettinToasty (approved):

PR #5366, opened 2025-04-12
Output native errors to console from Visual Studio Code

sandboxcoder:
When troubleshooting issues from `obs-studio-node` it can be useful to capture `std` output in console.

However, one could argue perhaps this might be annoying to those that do not care to see native logs? We could add another launch target if that's the case? For now I added it to "Launch with DevTools".


gettinToasty (approved):

gettinToasty:
I think this is ok for now :) we can revisit if it gets annoying or troublesome to FE folks

PR #5370, opened 2025-04-15
[Highlighter] -  performance improvements

marvinoffers:
- moves components to reduce re-renders
- other performance improvements
- change structure of how highlighted streams are stored in localstorage 
(so testing this branch and then switching to an older branch will cause the highlighted streams added in this branch to disappear)


app/services/highlighter/index.ts
```diff
@@ -303,8 +297,43 @@ export class HighlighterService extends PersistentStatefulService<IHighlighterSt
     return new HighlighterViews(this.state);
   }
 
+  private async migrateHighlightedStreamsToDictionary() {
```
marvinoffers, line 300: Migration

app/services/highlighter/index.ts
```diff
@@ -303,8 +299,39 @@ export class HighlighterService extends PersistentStatefulService<IHighlighterSt
...
+      // Check if current state exists and contains an array
+      if (
+        this.state &&
+        this.state.highlightedStreams &&
+        Array.isArray(this.state.highlightedStreams) &&
+        this.state.highlightedStreams.length > 0 &&
+        Object.keys(this.state.highlightedStreamsDictionary).length === 0
+      ) {
+        console.info('Migrating highlighted streams from array to dictionary structure');
+
```
michelinewu, line 313: Not sure if this logging is left over?

app/services/highlighter/index.ts
```diff
@@ -1282,7 +1309,7 @@ export class HighlighterService extends PersistentStatefulService<IHighlighterSt
   }
 
   cancelHighlightGeneration(streamId: string): void {
-    const stream = this.views.highlightedStreams.find(s => s.id === streamId);
+    const stream = this.views.highlightedStreamsDictionary[streamId];
     if (stream && stream.abortController) {
       console.log('cancelHighlightGeneration', streamId);
```
michelinewu, line 1314: Leftover console log

app/components-react/highlighter/StreamCard.tsx
```diff
@@ -14,49 +14,63 @@ import { $t } from 'services/i18n';
...
           clip.aiInfo.inputs = (clip.aiInfo as any).moments;
         }
         return clip;
-      }),
-  );
-  const stream = useVuex(() =>
-    HighlighterService.views.highlightedStreams.find(s => s.id === streamId),
-  );
+      });
+  }, [HighlighterService.views.clips.filter(clips => clips.streamInfo?.[streamId]), streamId]);
```
michelinewu, line 40: I've never seen a dependency with calculations before - is there a reason that this logic can't be moved into the use memo? Or are you trying to avoid useVuex?

ggolda: I think the problem is that useVuex with how it was made triggers an entire screen repaint that Marvin is trying to avoid here.

michelinewu (approved):

PR #5378, opened 2025-04-17
Revert 5365 revert 5285 mw migrate video realm

gettinToasty:


gettinToasty:
Currently requires a BE fix for performance analytics to follow to new API and properly update footer stats

PR #5379, opened 2025-04-17
allow for output handlers to be passed in, so that we can get debuggi…

tmaneri:
…ng info to the FE


app/services/platform-apps/api/modules/native-components.ts
```diff
@@ -14,13 +16,17 @@ export class NativeComponentsModule extends Module {
   visionUpdater = new VisionUpdater();
 
   @apiMethod()
-  async isAvatarUpdateAvailable() {
+  async isAvatarUpdateAvailable(proc: ChildProcess) {
```
avacreeth, line 19: assuming this isn't needed?

avacreeth (approved):
Seems like a good approach to me

PR #5413, opened 2025-05-14
upgrade napi-build-utils to resolve realm compile issue on Apple Silicon

sandboxcoder:
* The realm native node module has trouble building on _some_ developer machines (I never did 100% figure out why). Upgrade to latest napi-build-utils to resolve this issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "react-colorful": "^5.5.0",
     "react-devtools-electron": "^4.7.0",
     "react-transition-group": "^4.4.1",
-    "realm": "12.7.1",
+    "realm": "12.14.2",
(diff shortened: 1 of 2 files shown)


gettinToasty:
hmm i think we probably want to make this change on package.json instead of editing the yarn.lock directly, since that will typically get auto-generated and overwritten whenever we make dependency changes

sandboxcoder:
> hmm i think we probably want to make this change on package.json instead of editing the yarn.lock directly, since that will typically get auto-generated and overwritten whenever we make dependency changes

Thanks for your feedback. Only solution I can think of is to upgrade `realm`? This way the native node tools are upgraded.

gettinToasty (approved):

PR #5415, opened 2025-05-20
feat(sources): add search to Add Source modal

blackxored:


package.json
```diff
@@ -251,4 +251,4 @@
     "got@^9.6.0": "11.8.5"
   },
   "packageManager": "yarn@3.1.1"
-}
```
michelinewu: To prevent merge conflicts with staging can we remove this line break?

app/components-react/windows/source-showcase/SourceGrid.tsx
```diff
@@ -69,33 +70,125 @@ export default function SourceGrid(p: { activeTab: string }) {
...
+        const widgetPlatforms = WidgetDisplayData(primaryPlatform)[WidgetType[type]]?.platforms;
+        if (!widgetPlatforms) return true;
+        return linkedPlatforms?.some(platform => widgetPlatforms && widgetPlatforms.has(platform));
+      })
+      .filter(type => {
+        // show only supported widgets
+        const whitelist = primaryPlatformService?.widgetsWhitelist;
+        if (!whitelist) return true;
+        return whitelist.includes(WidgetType[type]);
+      });
```
michelinewu, line 91: Is it possible to do this in one pass with reduce?

app/components-react/windows/source-showcase/SourceGrid.tsx
```diff
@@ -69,33 +70,125 @@ export default function SourceGrid(p: { activeTab: string }) {
...
+
+  type FuseItem = {
+    type: string;
+    name: string;
+    description: string;
+    shortDesc: string;
+  };
 
-  const availableSources = useMemo(() => {
+  const useSearchMemo = <T extends string[] | IObsListOption<TSourceType>[]>(
```
michelinewu, line 153: Would it be worthwhile to move this to our hooks utils?

blackxored: For now it is tied to sources and widgets, see the `isSourceTypeList` requirement and all of the type massaging around `IObsListOption` and friends. We could pass those in but at that point I need a second user of the code which we don't have :D

app/components-react/windows/source-showcase/SourceGrid.tsx
blackxored: For performance, yes. For readability IMHO I don't believe the intermediate assignments it would require and the longer code branch in general would provide an improvement, it is also memoized, what do you think?

michelinewu: Makes sense to keep it this way because of readability and memoization. Let's leave as is

app/components-react/windows/source-showcase/SourceGrid.tsx
michelinewu: Makes sense then to wait and revisit if we need to with a second use case

michelinewu (approved):

PR #5446, opened 2025-06-12
Update Desktop to support the new mac virtual camera system extension

sandboxcoder:
### Description
Updates the virtualcam included in obs30 to work with externally outside of OBS.app by configuring it with the `com.streamlabs.slobs` bundle identifier and enabling the swift camera-extension to be downloaded and built by a different repo for `codesign`.

### Motivation and Context
The former legacy DAL plugin was deprecated by Apple. Enabled the mac obs30 virtual camera to be used with Streamlabs Desktop. This way we are using the same implementation as obs.

### How Has This Been Tested?
* In OSN, added a small test case that initialized the virtual cam. However, this only works locally due to System extension(s) security restrictions (cant install those on a build agent).
* In Streamlabs-Desktop, verified this works in non-codesigned builds.
* Tested mostly outside of app bundles via the `VIRTUALCAM_BYPASS_SYSTEM_CHECK` environment var (I updated cmake to look for this var when its building the project for xcode). Also tested within an app bundle but w/o codesign.

### DIFF
--- a/app/services/core/signals.ts
+++ b/app/services/core/signals.ts
@@ -1,0 +1,46 @@
+import { EOutputCode } from '../../../obs-api';
+
+export enum EOBSOutputType {
+  Streaming = 'streaming',
+  Recording = 'recording',
+  ReplayBuffer = 'replay-buffer',
+  VirtualCam = 'virtual-camera',
+}
+
+export enum EOBSOutputSignal {
+  Starting = 'starting',
+  Start = 'start',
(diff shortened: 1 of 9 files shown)


sandboxcoder:
TODO: found this issue after posting the PR. Need to refactor `OBS_service_connectOutputSignals` to be shared between virtualcamService & StreamingService.

michelinewu (approved):

PR #5451, opened 2025-06-16
Merge stream avatar working branch into staging

avacreeth:
Merging our working stream avatar branch into staging so we don't diverge too much.

- introduces the concept of "highly privileged apps" which have full node access, as well as access to special modules
- highly privileged apps are explicitly listed by app id
- adds avatar updater and runtime service (will change)


app/services/platform-apps/api/modules/streaming-recording.ts
```diff
@@ -51,6 +51,16 @@ export class StreamingRecordingModule extends Module {
     return this.serializeStreamInfo();
   }
 
+  @apiMethod()
```
blackxored, line 54: I reckon we haven't wanted to use start/stop streaming methods for a while, but as these are it could be confusing and get into invalid states, between `startStreaming` and `stopStreaming` here, one calling part of what `toggleStreaming` does while the other one just calls it by name. I think we should either use `toggleStreaming` on both (and expose that API instead) or force these methods to check if we're live and no-op if needed, what it probably doesn't happen but it would be confusing at the very least is that a call to `stopStreaming` could end up starting a stream here.

app/services/platform-apps/index.ts
```diff
@@ -165,6 +172,16 @@ class PlatformAppsViews extends ViewHandler<IPlatformAppServiceState> {
   getDelisted(appId: string) {
     return this.getApp(appId).delisted;
   }
+
+  isAppHighlyPrivileged(appId: string) {
```
blackxored, line 176: Minor nitpick: would prefer this to be done server-side at the cost of a small API call, for a) security if needed b) ease of maintenance if these ever need to change

app/services/app/app.ts
```diff
@@ -185,9 +187,10 @@ export class AppService extends StatefulService<IAppState> {
     this.START_LOADING();
     this.loadingChanged.next(true);
     this.tcpServerService.stopListening();
-
     window.setTimeout(async () => {
       obs.NodeObs.InitShutdownSequence();
+      this.streamAvatarService.stopAvatarProcess();
```
blackxored, line 192: Are these supposed to be called unconditionally?

app/services/platform-apps/api/modules/streaming-recording.ts
```diff
@@ -51,6 +51,16 @@ export class StreamingRecordingModule extends Module {
...
+  startStreaming() {
+    this.streamingService.goLive();
+  }
+
```
michelinewu, line 58: Curious about why this directly goes live instead of toggling streaming?

blackxored (changes requested):

app/services/platform-apps/api/modules/streaming-recording.ts
avacreeth: The idea was that since the avatar is managing the stream, it was meant to be a hands free experience that skips the go live window.  This won't work for all platforms so it's definitely something to think about whether this is the right behavior long term.

app/services/platform-apps/api/modules/streaming-recording.ts
avacreeth: I was under the assumption that `goLive` would be a no-op if already live, but I'll be honest it's been a while since I've taken a deep dive into the streaming service.

app/services/platform-apps/index.ts
avacreeth: Yep it should be a database flag long term for sure.

app/services/app/app.ts
avacreeth: Yeah it will be a no-op if avatar isn't running.

package.json
```diff
@@ -95,6 +99,7 @@
     "realm": "12.7.1",
     "rimraf": "^2.6.1",
     "semver": "^5.5.1",
+    "sharp": "^0.33.5",
```
gettinToasty, line 102: just a note on this since we've tried adding the sharp package multiple times for ai highlighter and it consistently crashes on launch for macos, however i also don't see it actually being used anywhere in the PR so i'm wondering why we're adding it here?

gettinToasty (changes requested):

PR #5472, opened 2025-06-30
[Highlighter] - Add pubg and CS

marvinoffers:
- just adds the new game PUBG


marvinoffers:
Also added counter-strike

michelinewu (approved):

PR #5491, opened 2025-07-14
Check for undefined error in diag report.

michelinewu:
This fixes a bug where some users were unable to upload a diag report.

### DIFF
--- a/app/components-react/windows/settings/Video.tsx
+++ b/app/components-react/windows/settings/Video.tsx
@@ -545,7 +545,6 @@ export function VideoSettings() {
 
   return (
     <>
-      <ObsGenericSettingsForm />
       <div className={styles.videoSettingsHeader}>
         <h2>{$t('Video')}</h2>
         <DualOutputToggle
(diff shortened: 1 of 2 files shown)


gettinToasty (approved):

app/services/diagnostics.ts
```diff
@@ -1063,6 +1063,7 @@ export class DiagnosticsService extends PersistentStatefulService<IDiagnosticsSe
         if (
           s?.type === 'Single Output' &&
           platforms.includes('tiktok') &&
+          s?.error &&
```
blackxored, line 1066: can we changed to just s?.error?.split... but no biggie

app/components-react/windows/settings/Video.tsx
```diff
@@ -12,6 +12,7 @@ import { invalidFps, IVideoInfoValue, TDisplayType } from 'services/settings-v2/
 import { AuthModal } from 'components-react/shared/AuthModal';
 import Utils from 'services/utils';
 import DualOutputToggle from '../../shared/DualOutputToggle';
+import { ObsGenericSettingsForm } from './ObsSettings';
```
blackxored, line 15: unused import?

app/services/diagnostics.ts
michelinewu: So this is actually the bug caused by only doing `s?.error?.split` because it was trying to split something undefined

PR #5492, opened 2025-07-14
update setVideoSetting to also sync the fps setting

sandboxcoder:
Bug fix for issue where the fps performance counter did not show the correct value 

* add optional parameter to updateObsSettings that will invoke `syncFPSSettings` after the settings have been applied to ensure these two operations are executed sequentially. Previously, syncFPSSettings()
would run *before* updateObsSettings (because it has the debounce modifier applied).


sandboxcoder:
ignore this PR; need to redo

PR #5500, opened 2025-07-16
Fix inconsistent go live experiences with custom stream settings and …

gettinToasty:
…dual output

Fixes the below issue as brought up by support, which was actually expected behavior due to how the stream to custom ingest setting works, but is incompatible with dual output

> I also found a small issue when using the "disable dual output" option in the go live settings when the settings > stream is set to "stream to custom ingest". It seems this doesn't default back to the normal go live screen with the ability to toggle on and off platforms. Will attach gif in comments
Streamlabs_OBS_0iZolohp1X

### DIFF
--- a/app/services/settings/settings.ts
+++ b/app/services/settings/settings.ts
@@ -381,7 +381,9 @@ export class SettingsService extends StatefulService<ISettingsServiceState> {
   }
 
   showSettings(categoryName?: string) {
-    this.navigationService.setSettingsNavigation(categoryName);
+    if (categoryName) {
+      this.navigationService.setSettingsNavigation(categoryName);
+    }
     this.windowsService.showWindow({
       componentName: 'Settings',
       title: $t('Settings'),


app/services/dual-output/dual-output.ts
```diff
@@ -336,25 +348,54 @@ export class DualOutputService extends PersistentStatefulService<IDualOutputServ
...
+  setDualOutputModeIfPossible(
+    status: boolean = true,
+    skipShowVideoSettings: boolean = false,
+    showGoLiveWindow?: boolean,
+  ) {
+    if (!this.userService.isLoggedIn) return;
+
+    // If a user is not in protected mode (ie using "Stream to a Custom Ingest")
+    // Dual Output will not function correctly and try to overwrite the settings
+    console.log(status, this.streamSettingsService.protectedModeEnabled);
```
michelinewu: Not sure if this log is needed?

package.json
```diff
@@ -252,4 +252,4 @@
     "napi-build-utils": "2.0.0"
   },
   "packageManager": "yarn@3.1.1"
-}
\ No newline at end of file
+}
```
michelinewu: I think the linter accidentally removed this?

michelinewu (changes requested):

PR #5509, opened 2025-07-24
fix [ARCH] for virtual cam installer tar.gz

sandboxcoder:
* fix mac x86_64 builds to download the correct virtual camera installer for their ARCH
* refactor reusable logic from `install-native-deps` (where it will lookup darwin ARCH)
* bump slobs-vcam-installer, `macVirtualCamUrl` in package.json, to resolve invalid code signature error due to missing camera extension
* bump obs-studio-node to `0.25.44` (for error messages, legacy DAL Plugin uninstall, virtual cam fixes)
* Display error/warning messages wheen virtual cam is installed
image

# How was this tested?
* Locally, I made a codesigned build and verified mac virtual camera worked 
* I rebuilt desktop (remove `node_modules` and run `yarn install` to force the script to redownload the dependencies from `script/repositories.json`) to verify my change to `install-native-deps`

### DIFF
--- a/scripts/install-native-deps.js
+++ b/scripts/install-native-deps.js
@@ -1,7 +1,7 @@
 const sh = require('shelljs');
 const colors = require('colors/safe');
 const fs = require('fs');
-const os = require('os');
+const findDarwinArch = require('./find-darwin-arch');
 const path = require('path');
 const stream = require('stream');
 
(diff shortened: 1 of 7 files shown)


scripts/install-native-deps.js
```diff
@@ -1,7 +1,7 @@
 const sh = require('shelljs');
 const colors = require('colors/safe');
 const fs = require('fs');
-const os = require('os');
```
sandboxcoder, line 4: This statement was unused

PR #5533, opened 2025-08-21
Factory API Migration

michelinewu:


app/components-react/highlighter/SettingsView.tsx
```diff
@@ -112,6 +110,25 @@ export default function SettingsView({
...
 
+  function handleToggleHighlighter() {
+    if (disableAIHighlighter) {
+      promptAction({
+        title: $t('Vertical Recording Active'),
+        message: $t(
+          'Vertical recording is in-progress. Would you like to stop the recording to enable AI Highlighter?',
+        ),
+        btnText: $t('Stop Recording'),
+        fn: StreamingService.actions.toggleRecording,
```
gettinToasty: do we want this to also toggle ai highlighter on afterward? or are they supposed to hit it twice?

app/components-react/root/StudioFooter.tsx
```diff
@@ -116,7 +120,7 @@ export default function StudioFooterComponent() {
   return (
     <div className={cx('footer', styles.footer)}>
       <div className={cx('flex flex--center flex--grow flex--justify-start', styles.footerLeft)}>
-        <Tooltip placement="left" title={$t('Open Performance Window')}>
+        <AntdTooltip placement="left" title={$t('Open Performance Window')}>
```
gettinToasty, line 135: kinda wondering why we're not using our wrapper here?

michelinewu: We're not using our wrapper because the nesting prevents the tooltip from rendering correctly. The extra `div` positions the tooltip inside the parent component so looks squished.

app/components-react/highlighter/SettingsView.tsx
michelinewu: Good call. Updated

PR #5538, opened 2025-08-25
moving the source message callback up a level, so it doesn't get repl…

tmaneri:
…aced for each smart source


tmaneri:
`obs.NodeObs.RegisterSourceMessageCallback` only supports a single callback, so each time a smart source loads it ends up overriding it, and preventing other smart sources from getting notified of changes

PR #5597, opened 2025-10-01
Switcher and go live UI.

michelinewu:


app/components-react/modals.tsx
```diff
@@ -46,7 +46,9 @@ export function confirmAsync(
  * alert('This is Alert').then(() => console.log('Alert closed'))
  *
  */
-export function alertAsync(p: Omit<ModalFuncProps, 'afterClose'> | string): Promise<void> {
+export function alertAsync(
+  p: (Omit<ModalFuncProps, 'afterClose'> | string) & { afterCloseFn?: () => void },
```
gettinToasty, line 50: why are we not just using `ModalFuncProps | string` here with the existing `afterClose` function prop?

app/components-react/modals.tsx
```diff
@@ -137,16 +148,33 @@ export function promptAction(p: {
       <ModalLayout
         footer={
           <Form layout={'inline'} className={styles.actionModalFooter}>
-            <Button onClick={Modal.destroyAll}>{$t('Skip')}</Button>
+            {!p.cancelBtnPosition ||
+              (p.cancelBtnPosition && p.cancelBtnPosition === 'left' && (
+                <Button onClick={Modal.destroyAll}>{p.cancelBtnText ?? $t('Skip')}</Button>
```
gettinToasty: this doesn't seem to include the `cancelFn` callback like the right side button

app/components-react/root/StartStreamingButton.tsx
```diff
@@ -44,10 +67,78 @@ export default function StartStreamingButton(p: { disabled?: boolean }) {
...
+
+      const isMobileRemote = streamShiftStreamId && /[A-Z]/.test(streamShiftStreamId);
+      console.debug(
+        'Desktop stream id: ' +
+          streamShiftStreamId +
+          '\nEvent:' +
+          JSON.stringify(event) +
+          '\nSource: ' +
+          (isMobileRemote ? 'Mobile' : 'Desktop'),
+      );
```
gettinToasty, line 97: console debug

app/components-react/root/StartStreamingButton.tsx
```diff
@@ -44,10 +67,78 @@ export default function StartStreamingButton(p: { disabled?: boolean }) {
...
-  async function toggleStreaming() {
+  useEffect(() => {
+    if (!isDualOutputMode && isPrime) {
+      fetchStreamShiftStatus();
+    }
+
+    const streamShiftEvent = StreamingService.streamShiftEvent.subscribe(event => {
+      const streamShiftStreamId = RestreamService.state.streamShiftStreamId;
+
+      const isMobileRemote = streamShiftStreamId && /[A-Z]/.test(streamShiftStreamId);
```
gettinToasty: i think i might need some clarification on this regex check-- any id with a capital letter signifies a mobile remote setup?

app/components-react/root/StartStreamingButton.tsx
```diff
@@ -89,6 +180,33 @@ export default function StartStreamingButton(p: { disabled?: boolean }) {
...
+      // Only check for Stream Shift for ultra users
+      if (isLoggedIn && isPrime && canUseStreamShift) {
+        setIsLoading(true);
+        const isLive = await fetchStreamShiftStatus();
+        setIsLoading(false);
+
+        const message = isDualOutputMode
+          ? $t(
+              'A stream on another device has been detected. Would you like to switch your stream to Streamlabs Desktop? If you do not wish to continue this stream, please end it from the current streaming source.',
+            ) + $t('Dual Output will be disabled since not supported in this mode.')
```
gettinToasty: we shouldn't be concatenating translated strings in case grammatical patterns of different languages require a restructure

app/components-react/root/StartStreamingButton.tsx
```diff
@@ -136,7 +274,11 @@ export default function StartStreamingButton(p: { disabled?: boolean }) {
         StreamSettingsService.isSafeToModifyStreamKey()
       );
     }
-  }
+  }, [
+    UserService.state.auth?.primaryPlatform,
+    StreamingService.views.isMultiplatformMode,
+    CustomizationService.state.updateStreamInfoOnLive,
```
gettinToasty: these won't be reactive unless they're hooked into useVuex

app/components-react/root/StartStreamingButton.tsx
```diff
@@ -145,43 +287,56 @@ export default function StartStreamingButton(p: { disabled?: boolean }) {
...
     </button>
   );
 }
 
-function StreamButtonLabel(p: {
-  streamingStatus: EStreamingState;
-  delaySecondsRemaining: number;
-  delayEnabled: boolean;
-}) {
+const StreamButtonLabel = forwardRef<
```
gettinToasty, line 339: not sure why there's a ref being forwarded here if you aren't using one

app/components-react/shared/DisplaySelector.tsx
```diff
@@ -69,20 +84,32 @@ export default function DisplaySelector(p: IDisplaySelectorProps) {
...
+      return acc;
+    }, {} as Dictionary<IRadioMetadata>);
+  }, [displays]);
+
+  const name = `${p.platform || `destination${p.index}`}Display`;
+  const value = displayDict[display]?.value || 'horizontal';
+
   return (
     <RadioInput
+      // data-title={p.title}
```
gettinToasty: comment

app/components-react/shared/inputs/RadioGroup.tsx
```diff
@@ -0,0 +1,53 @@
...
+        gapsize={0}
+        defaultValue={p.defaultValue}
+        options={p.options}
+        onChange={(key: string) => (value: any) => p.onChange(value)}
+        value={p.value}
+        className={p?.className}
+        style={p?.style}
+        icons={true}
+      />
+    </InputWrapper>
```
gettinToasty: i dont really understand why there's an InputWrapper around a component that already has an InputWrapper, or why just using RadioInput didn't work here

app/components-react/windows/settings/Stream.tsx
```diff
@@ -43,31 +47,7 @@ function censorEmail(str: string) {
  * A Redux module for components in the StreamSetting window
  */
 class StreamSettingsModule {
```
gettinToasty, line 49: im a little confused-- it looks like we took out all the references to this module but we're still keeping it around?

app/components-react/windows/settings/Stream.tsx
```diff
@@ -563,7 +554,11 @@ function Platform(p: { platform: TPlatform }) {
  * Renders a custom destinations list
  */
 function CustomDestinationList() {
-  const { isPrime, customDestinations, editCustomDestMode, addCustomDest } = useStreamSettings();
+  const { customDestinations, editCustomDestMode, addCustomDest } = useStreamSettings();
```
gettinToasty, line 557: ^ oh nvm the above found it, but if we're decoupling from the module that much i think it might just be worth to do it all the way

app/services/platforms/youtube.ts
```diff
@@ -465,6 +529,20 @@ export class YoutubeService
...
+
+    //   if (broadcasts.length) {
+    //     const streamShiftBroadcast = broadcasts.find(b => b.id === this.state.settings.broadcastId);
+
+    //     // If for some reason the broadcast is still live, end it
+    //     if (streamShiftBroadcast && streamShiftBroadcast.status.lifeCycleStatus === 'live') {
+    //       await this.stopBroadcast(streamShiftBroadcast.id);
+    //     }
+    //   }
+    // }
```
gettinToasty, line 592: not sure if this is unneeded or not

app/services/restream.ts
```diff
@@ -408,6 +506,88 @@ export class RestreamService extends StatefulService<IRestreamState> {
...
+  async checkIsLive(): Promise<boolean> {
+    if (!this.canUseStreamShift) return false;
+
+    const status = await this.fetchLiveStatus();
+    console.debug('Stream Shift Status', status);
+
+    if (status.isLive) {
+      this.streamSettingsService.setGoLiveSettings({ streamShift: true });
+      this.SET_STREAM_SWITCHER_STATUS('pending');
+      this.SET_STREAM_SWITCHER_TARGETS(status.targets);
```
gettinToasty, line 518: it feels like a code smell that a status checking function would cause mutations as a side effect, not sure entirely what the intention of this function is

app/services/settings/settings.ts
```diff
@@ -427,7 +427,7 @@ export class SettingsService extends StatefulService<ISettingsServiceState> {
   getCategories(): CategoryName[] {
     let categories: CategoryName[] = obs.NodeObs.OBS_settings_getListCategories();
     // insert 'Multistreaming' after 'General'
-    categories.splice(1, 0, 'Multistreaming');
+    // categories.splice(1, 0, 'Multistreaming');
```
gettinToasty, line 430: are we just removing this settings page entirely?

app/services/streaming/streaming.ts
```diff
@@ -365,6 +400,7 @@ export class StreamingService
       // Note: Because the horizontal video context is the default, it does not need
       // to be validated.
 
+      console.log('this.views.isDualOutputMode', this.views.isDualOutputMode);
```
gettinToasty: console log

app/components-react/root/LiveDock.tsx
```diff
@@ -214,11 +221,25 @@ class LiveDockController {
 
   popOut() {
     this.platformAppsService.popOutAppPage(this.store.selectedChat, this.pageSlot);
-    this.store.setState(s => {
+    this.store.setState((s: any) => {
```
blackxored: Did TypeScript drop the ball on this file? These types shouldn't be needed for any of the `setState` calls in this file. If they did drop the ball and can't infer (wouldn't be surprised) can you add a comment to the first one stating so?

app/components-react/root/StartStreamingButton.tsx
```diff
@@ -44,10 +67,78 @@ export default function StartStreamingButton(p: { disabled?: boolean }) {
     }
   }, [delaySecondsRemaining, streamingStatus, delayEnabled]);
 
-  async function toggleStreaming() {
+  useEffect(() => {
+    if (!isDualOutputMode && isPrime) {
+      fetchStreamShiftStatus();
+    }
+
+    const streamShiftEvent = StreamingService.streamShiftEvent.subscribe(event => {
```
blackxored, line 83: So this initially was a comment about unsubscribing from this event, but then I found it does do that so nevermind, the only useful part of that was the nudge that some RxJS operators might help with some of what we're doing in the function body (things like filtering and case switching), but totally optional and more of a nice-to-have.

app/components-react/root/StartStreamingButton.tsx
```diff
@@ -44,10 +67,78 @@ export default function StartStreamingButton(p: { disabled?: boolean }) {
...
+
+      if (event.type === 'switchActionComplete') {
+        const remoteStreamId = event.data.identifier;
+
+        if (remoteStreamId !== streamShiftStreamId) {
+          UsageStatisticsService.recordAnalyticsEvent('StreamShift', {
+            stream: 'mobile-to-desktop',
+          });
+
+          promptAction({
```
blackxored: These two calls to `promptAction` seem to be the same except for the `fn` and the `message`, 2 is technically my limit with dupes but considering how many lines this is in an already non-trivial function, I wouldn't mind if we made an exception and abstract this and parameterize those two, for desktop/mobile/etc (wink wink). so we can reduce line count.

app/components-react/shared/AddDestinationButton.tsx
```diff
@@ -66,32 +68,63 @@ export default function AddDestinationButton(p: IAddDestinationButtonProps) {
...
+      <PlusIcon style={{ paddingLeft: '17px', fontSize: '24px' }} />
+      <span style={{ flex: 1 }}>{$t('Add Destination')}</span>
+    </Button>
+  );
+}
+
+function SmallAddDestinationButton(p: { className?: string; onClick: () => void }) {
+  return (
+    <Button
+      data-test="default-add-destination"
```
blackxored: We should really decide if we're gonna use `data-test` or `data-testid`. Nit: add a `button` or `btn` suffix, when reading a test it would easier to spot that is a button (apart from the testing API that presumably clicks it).

app/components-react/shared/StreamShiftToggle.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+export default function StreamShiftToggle(p: IStreamShiftToggle) {
+  const { isPrime, isStreamShiftMode, setStreamShift } = useGoLiveSettings();
+
+  useEffect(() => {
+    // Ensure that non-ultra users have the stream switcher disabled
+    if (!isPrime && isStreamShiftMode) {
+      setStreamShift(false);
+    }
+  }, [isPrime, isStreamShiftMode, setStreamShift]);
```
blackxored: Does `setStreamShift` need to be a dependency? Also, depending on how it is defined, it could cause this to evaluate everytime, since it could create a function everytime or fail an equality test.

app/components-react/shared/inputs/RadioGroup.tsx
```diff
@@ -0,0 +1,53 @@
...
+      icon?: string;
+    }[];
+  },
+  string
+>;
+
+export const RadioGroup = InputComponent((p: TRadioGroupProps) => {
+  return (
+    <InputWrapper
+      // label={p.label}
```
blackxored: Left-over comment

app/components-react/windows/go-live/GoLiveSettings.tsx
```diff
@@ -33,16 +37,25 @@ export default function GoLiveSettings() {
...
     canAddDestinations,
     canUseOptimizedProfile,
-    showSelector,
     showTweet,
     hasMultiplePlatforms,
     hasMultiplePlatformsLinked,
     enabledPlatforms,
     primaryChat,
     recommendedColorSpaceWarnings,
+    isPrime,
```
blackxored, line 46: There are 3 fields related to stream shift we have to pull, deref and subsequently use, with no semantic relation other than their names, we have always done it like this so no action, but more of a thought provocation to maybe start grouping these in objects and pass those around.

app/components-react/windows/go-live/GoLiveWindow.tsx
```diff
@@ -86,54 +85,76 @@ function ModalFooter() {
...
+  async function handleGoLive() {
+    if (isPrime && canUseStreamShift) {
+      try {
+        setIsFetchingStreamStatus(true);
+        const isLive = await Services.RestreamService.checkIsLive();
+        setIsFetchingStreamStatus(false);
+
+        // Prompt to confirm stream switch if the stream exists
+        if (isLive) {
+          promptAction({
```
blackxored, line 111: Definitely (probably?) have seen this function before, so technically my 2 dupes rule is no longer valid :D

app/services/settings/settings.ts
michelinewu: Yeah, in favor of a less cluttered menu since it no longer had any settings managed by it

app/components-react/shared/AddDestinationButton.tsx
gettinToasty: data-name is the convention we use for most of our form selectors, i think it could probably be used here too

app/services/platforms/youtube.ts
michelinewu: I coded out two possible solutions for a bug. This is the second one, the first one is currently in testing. If the first fix seems to work, we can take this out. I'll add a TODO to remove it if the first fix is working and stable

app/components-react/root/StartStreamingButton.tsx
michelinewu: Yes, that's correct

app/components-react/modals.tsx
michelinewu: Because we use the spread operator when applying props and we need to use the `afterClose` prop to execute the function, to prevent possible issues with using the same prop name I added the `afterCloseFn` prop

PR #5707, opened 2026-01-15
Validate the encoder for dual output before going live

mhoyer-streamlabs:
A user is able to select an encoder in Output settings based on the main stream service, but if they go live with Dual Output that encoder is never validated for the additional service. If the encoder is invalid, the stream fails but the user isn't notified and no errors are logged. Added a function to the backend to validate encoders and use that to prevent this from happening. On `Go Live`, validate the encoder if using Dual Output but NOT using the restream server - that is handled elsewhere. If the encoder is invalid, prompt the user to default to x264 or select their own in settings.

To reproduce:

1. Go to Stream settings and select `Stream to custom ingest`
2. Select `Streaming Services` as stream type and select YouTube RTMPS
3. Click `Use recommended settings`
4. Go to Output settings and select the encoder `NVIDIA NVENC AV1`
5. `Go Live` and select `Dual Output`, choose YouTube and any other destination (I tested with Twitch and Kick)
6. `Confirm and Go Live` will prompt for an encoder change

image


app/services/streaming/streaming.ts
```diff
@@ -386,6 +386,46 @@ export class StreamingService
...
+      if (ytSettings?.display === 'both') {
+        streamToCheck = 'Stream';
+      }
+
+      const validEncoder = this.settingsService.isValidEncoder(streamToCheck as 'Stream' | 'StreamSecond' | 'Both');
+      if (!validEncoder) {
+        const choice = remote.dialog.showMessageBoxSync(this.windowsService.windows.child, {
+          type: 'error',
+          message:
+            'The currently configured encoder is not supported for the selected Dual Output destinations. Would you like to reset the stream encoder to Software (x264) or select a different encoder in settings?',
```
sandboxcoder, line 406: we need to add this message to those locale .ini files and update it to use the `$t()` pattern like so:
```
$t('The currently configured encoder is not supported for the selected Dual Output destinations. Would you like to reset the stream encoder to Software (x264) or select a different encoder in settings?'),
```
Same for other strings displayed to the user

mhoyer-streamlabs: Added strings to `app/i18n/en-US/settings.json` since from other PRs that's where it looks like new strings are added.

sandboxcoder (approved):

PR #5750, opened 2026-02-11
Port SceneTransitions to React

gettinToasty:


app/components-react/windows/scene-transitions/ConnectionsTable.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { ITransitionConnection } from 'services/transitions';
+import { ColumnsType } from 'antd/lib/table';
+import styles from './SceneTransitions.m.less';
+import Tooltip from 'components-react/shared/Tooltip';
+
+export default function ConnectionsTable(p: { setInspectedConnection: (id: string) => void; setShowConnectionModal: (val: boolean) => void; }) {
+  const { TransitionsService, EditorCommandsService, ScenesService } = Services;
+
+  const { transitions, connections } = useVuex(() => ({
+    transitions: TransitionsService.state.transitions,
```
michelinewu: Should these be views?

app/components-react/shared/ModalLayout.tsx
```diff
@@ -80,6 +81,7 @@ export function ModalLayout(p: TProps) {
     <div
       className={cx('ant-modal-content', currentTheme, p.className)}
       style={{ ...wrapperStyles, ...p.wrapperStyle }}
+      id={p.id}
```
michelinewu, line 84: Looks like `id` was removed from ModalLayout, unless it's already a prop in antd's `ModalProps`?

app/components-react/windows/scene-transitions/ConnectionSettings.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  const { sceneOptions, transitionOptions, connection } = useVuex(() => ({
+    sceneOptions: [
+      { label: $t('All'), value: 'ALL' },
+      ...ScenesService.views.scenes.map(scene => ({
+        label: scene.name,
+        value: scene.id,
+      })),
+    ],
+    transitionOptions: TransitionsService.state.transitions.map(transition => ({
```
michelinewu: Same as above for views vs state

app/components-react/windows/scene-transitions/ConnectionSettings.tsx
```diff
@@ -0,0 +1,47 @@
...
+import FormFactory from 'components-react/shared/inputs/FormFactory';
+import { useVuex } from 'components-react/hooks';
+import { Services } from 'components-react/service-provider';
+import { $t } from 'services/i18n';
+
+export default function ConnectionSettings(p: { connectionId: string }) {
+  const { ScenesService, TransitionsService, EditorCommandsService } = Services;
+
+  const { sceneOptions, transitionOptions, connection } = useVuex(() => ({
+    sceneOptions: [
```
michelinewu, line 12: Just for my own edification, is there a difference in reactivity/rerenders between doing the logic in the hook vs outside the hook?

app/components-react/windows/scene-transitions/TransitionsTable.tsx
```diff
@@ -0,0 +1,115 @@
...
+import Tooltip from 'components-react/shared/Tooltip';
+
+export default function TransitionsTable(p:
+  { setInspectedTransition: (id: string) => void, setShowTransitionModal: (val: boolean) => void; }
+) {
+  const { TransitionsService, EditorCommandsService } = Services;
+
+  const { transitions, defaultId } = useVuex(() => ({
+    transitions: TransitionsService.state.transitions,
+    defaultId: TransitionsService.state.defaultTransitionId,
```
michelinewu: Same as above for views vs state

app/components-react/windows/scene-transitions/ConnectionSettings.tsx
```diff
@@ -0,0 +1,47 @@
...
+  };
+
+  const values = {
+    from: connection?.fromSceneId || '',
+    transition: connection?.transitionId || '',
+    to: connection?.toSceneId || '',
+  }
+
+  function handleChange(key: string) {
+    return (val: string) => EditorCommandsService.executeCommand('EditConnectionCommand', p.connectionId, {
```
michelinewu: Should these be actions?

michelinewu (changes requested):

app/components-react/windows/scene-transitions/ConnectionSettings.tsx
gettinToasty: yea probably

app/components-react/windows/scene-transitions/ConnectionSettings.tsx
gettinToasty: if you do it outside of the hook i suppose it risks recalculating every render (unless using like useMemo or something), which i don't think happens if you put the logic in useVuex but tbh i haven't looked super closely at the useVuex implementation

app/components-react/windows/scene-transitions/ConnectionsTable.tsx
gettinToasty: in this case it doesn't make much difference since the view would just be
```
get transitions() {
  return this.state.transitions
}
```
which is technically more in pattern but we don't gain a ton

app/components-react/shared/ModalLayout.tsx
gettinToasty: this PR adds it back in

app/components-react/windows/scene-transitions/TransitionsTable.tsx
```diff
@@ -0,0 +1,137 @@
...
+  //  * scene transition creation options.
+  const lockStates = useMemo(() => TransitionsService.getLockedStates(), []);
+  function canEdit(id: string) {
+    return !lockStates[id];
+  }
+
+  function getEditableMessage(id: string) {
+    if (canEdit(id)) return;
+    return $t('This scene transition is managed by an App and cannot be edited.');
+  }
```
wesrupert, line 35: This return type seems off: `undefined | 'This scene transition is managed by an App and cannot be edited.'`

app/components-react/windows/scene-transitions/TransitionSettings.tsx
```diff
@@ -0,0 +1,55 @@
...
+          label: $t('Duration'),
+          displayed: values.type !== 'obs_stinger_transition' && values.type !== 'cut_transition',
+        }),
+      },
+    }),
+  };
+
+  const formData = useMemo(() => {
+    return TransitionsService.getPropertiesFormData(p.transitionId);
+  }, [p.transitionId, values.type]);
```
wesrupert, line 37: Having to track the `values.types` dependency directly, when it's not used in the closure, has a huge code smell to me. Is there a better way to capture the cross-dependency here? Maybe move this into the `useVuex` call alongside `values`?

If not, there should probably be a comment explaining why `TransitionsService.getPropertiesFormData(p.transitionId)` has a dependency on a "separate" object.

app/components-react/windows/scene-transitions/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  const [activeTab, setActiveTab] = useState('transitions');
+  const [showConnectionModal, setShowConnectionModal] = useState(false);
+  const [showTransitionModal, setShowTransitionModal] = useState(false);
+  const [inspectedTransition, setInspectedTransition] = useState('');
+  const [inspectedConnection, setInspectedConnection] = useState('');
+
+  const { transitionsEnabled } = useVuex(() => ({
+    transitionsEnabled: ScenesService.views.scenes.length > 1,
+  }));
```
wesrupert, line 25: More a style question than a suggestion, but I've seen this style in a few places now and wonder why structure/destructure like this, vs directly returning/capturing the value like below?

```typescript
  const transitionsEnabled = useVuex(() => ScenesService.views.scenes.length > 1);
```

gettinToasty: i think its just habit tbh

app/components-react/windows/scene-transitions/TransitionSettings.tsx
gettinToasty: yea so this is a weird one in that the `ObsForm` fields change in response to the type of transition, which lives in a different form

basically its the interaction between two different form constructors which in an ideal world shouldn't co-exist but currently do

wesrupert (approved):

PR #5822, opened 2026-04-01
chore(widgets): port credits to react

wesrupert:
- Migrate Credits widget to React
- Clean up widget form types

Screenshot 2026-04-01 at 12 42 47


app/components-react/shared/inputs/metadata.ts
```diff
@@ -47,29 +46,20 @@ export const metadata = {
...
-  | ITextMetadata
-  | INumberMetadata
-  | ISliderMetadata
-  | ITextBoolMetadata
-  | ICheckboxGroupMetadata
-  | IRadioGroupMetadata
-  | IListMetadata<T>;
-
-interface IBaseMetadata {
+export interface IBaseMetadata {
```
wesrupert, line 50: This union type was hiding some potential bugs, so I moved to using the underlying base interface as our main export:
- There was no way for other widgets to reference the base class for input types that don't have a hand-written type yet
- The inference from `TInputMetadata<unknown>` was casting custom types to incorrect subtypes, like SponsorBanner's `imagepicker` input was casting as a `IListMetadata` since it was the only type with `.options`

app/components-react/widgets/Credits.tsx
```diff
@@ -0,0 +1,220 @@
...
+      roll_time: number;
+      loop_credits: boolean;
+    };
+  };
+}
+
+type TCreditsMeta = PartialRec<
+  keyof ICreditsState['data']['settings'] | 'themes' | `_${string}`,
+  IBaseMetadata
+>;
```
wesrupert, line 53: After struggling a bit to figure out how widget settings are synced via this system, I wanted to make it explicit in the types. This record type forces each section to either (a) use a synced value defined above, or (b) be explicit in its unsynced behavior via an underscore. See `creditsMeta` for more details.

app/components-react/widgets/Credits.tsx
```diff
@@ -0,0 +1,224 @@
...
+            '{top_donor}',
+            '{top_donated_amount}',
+            '{top_cheer_donor}',
+            '{username}',
+            '{top_cheer_amount}',
+            '{new_subscriber_count}',
+            '{new_follower_count}',
+          ].join(', '),
+      }),
+      _includes: {
```
wesrupert, line 154: This is where the `TCreditsMeta` magic helps make the config explicit. If this were named `includes`, TypeScript will now emit the following:
> ```
> Diagnostics:
> Object literal may only specify known properties,
> and 'includes' does not exist in type 'Partial<Record<
>   "theme" | "credit_title" | "credit_subtitle" | "background_color"
>     | "font_color" | "font_size" | "font" | "muted_chatters" | "bits"
>     | "subscribers" | "moderators" | "donations" | ... 11 more ...
>     | `_${string}`,
>   IBaseMetadata>
> >'. [2353]
> ```

Same thing for typos like `font_size` vs `fontsize`, and even widget API mismatches like Credits' `font_size` vs Event List's `text_size`.

app/components-react/widgets/Credits.tsx
```diff
@@ -0,0 +1,224 @@
...
+  get visualMeta() {
+    return fromMeta({
+      theme: metadata.list({
+        label: $t('Theme'),
+        options: Object.entries(this.widgetData?.themes || {}).map(([theme, val]) => ({
+          label: val.label,
+          value: theme,
+        })),
+      }),
+      background_color: metadata.color({
```
wesrupert, line 190: Review question: how does the new system handle previews? I didn't see anywhere in the reference code where event list was handling preview.

app/services/widgets/widgets-config.ts
```diff
@@ -476,9 +477,31 @@ export function getWidgetsConfig(
...
-    // },
+    [WidgetType.Credits]: {
+      type: WidgetType.DonationTicker,
+
+      defaultTransform: {
+        width: 600,
+        height: 200,
+        x: 1,
+        y: 1,
+        anchor: AnchorPoint.North,
```
wesrupert: I guessed at this value. Was it defined in the old code somewhere?

wesrupert:
Rebased on current master

app/components-react/widgets/Credits.tsx
gettinToasty: its all handled by WidgetLayout iirc

app/services/widgets/widgets-config.ts
gettinToasty: should be in `widgets-data` in the `WidgetDefinitions` function i think

gettinToasty (approved):

PR #5847, opened 2026-04-16
Highlighter/jk replay integration

jankalthoefer:


jankalthoefer:
@copilot review with the latest changes.

app/components-react/highlighter/migration/FeatureCarousel.tsx
```diff
@@ -0,0 +1,282 @@
...
+    headline: 'Auto record gameplay',
+    description:
+      'Not streaming? No problem! Your gameplay gets recorded automatically — if you want it to.',
+    topColor: '#1E0101',
+    bottomColor: '#1E0101',
+    previewImage: `${IMAGE_PATH}/auto-record.webp`,
+    blobColor: '#FF4655',
+    iconUrl: `${IMAGE_PATH}/auto-record-icon.svg`,
+  },
+];
```
gettinToasty, line 74: these need to be translated, but that can be done in another PR later

app/components-react/highlighter/migration/FeatureCarousel.tsx
```diff
@@ -0,0 +1,282 @@
...
+  const isListHoveredRef = useRef(false);
+  const isAnimatingRef = useRef(false);
+  const currentIndexRef = useRef(currentIndex);
+  const transitionStateRef = useRef(transitionState);
+  const incomingIndexRef = useRef(incomingIndex);
+
+  currentIndexRef.current = currentIndex;
+  isAnimatingRef.current = transitionState === 'animating';
+  transitionStateRef.current = transitionState;
+  incomingIndexRef.current = incomingIndex;
```
gettinToasty: why are these being reassigned after they are initialized with the same values?

app/components-react/highlighter/migration/FeatureCarousel.tsx
```diff
@@ -0,0 +1,282 @@
...
+        requestAnimationFrame(() => {
+          requestAnimationFrame(() => {
+            setTransitionState('animating');
+          });
+        });
+      }
+
+      setTimeout(() => finishAnimation(index), duration);
+    },
+    [finishAnimation],
```
gettinToasty: not sure why this is in the dependency graph since its not something that changes?

app/components-react/highlighter/migration/ModalInstallationFlow.tsx
```diff
@@ -0,0 +1,165 @@
...
+  if (step === 'error') {
+    return (
+      <div className={cx(styles.installationFlow, styles.installationFlowModal)}>
+        <SectionHeader title={$t('Installing Highlighter')} onClose={onCancel} />
+        <div className={styles.installContent}>
+          <div className={styles.errorTitle}>{$t('Installation interrupted')}</div>
+          <p className={styles.installSubtext}>
+            {$t("It seems the installation didn't finish.")}
+            <br />
+            {$t("Let's figure this out together and")}{' '}
```
gettinToasty: we shouldn't be using concatenated strings for translations since it makes it difficult/impossible for the translations team to use proper context. we have the `Translation />` react component to allow us to insert links into translated strings

app/components-react/highlighter/HypeWrapper.tsx
```diff
@@ -0,0 +1,224 @@
...
+          gradientUnits="userSpaceOnUse"
+          gradientTransform="translate(-3.79108 8.32046) rotate(66.1069) scale(11.3591 46.8348)"
+        >
+          <stop stopColor="#3771C8" />
+          <stop offset="0.128" stopColor="#3771C8" />
+          <stop offset="1" stopColor="#6600FF" stopOpacity="0" />
+        </radialGradient>
+      </defs>
+    </svg>
+  );
```
gettinToasty: we already have platform icons available for every platform in the `<PlatformIcon />` component

gettinToasty (changes requested):
see above

app/components-react/highlighter/HypeWrapper.tsx
jankalthoefer: Okay, logos are replaced with the default logos, except Discord.

app/components-react/highlighter/migration/ModalInstallationFlow.tsx
jankalthoefer: done

app/components-react/highlighter/migration/FeatureCarousel.tsx
jankalthoefer: already added translation now

app/components-react/highlighter/migration/FeatureCarousel.tsx
jankalthoefer: not too sure, I see how this approach was a bit confusing.
Did a new approach with just one reducer that handles the state and the transition a little more subtle.

Let me know if you see any issues with that.

app/components-react/highlighter/migration/FeatureCarousel.tsx
jankalthoefer: also gone in the new approach

gettinToasty (approved):

PR #5858, opened 2026-04-22
Add realm properties hooks and reduce main window re-renders.

michelinewu:
**Overview**                                                         
This PR introduces new React hooks (useRealmProperty and useRealmProperties) to minimize unnecessary component re-renders across the application, particularly in the main window.         
                                                                   
**Issues Addressed**                                                                                                          
1. Excessive re-renders - Multiple components were re-rendering when realm properties changed in the realm object, not just the property referenced
2. Stale values in resize handlers - Chat and other resizable components had debounce/cleanup issues with stale state
3. Performance in main window and dock - The main window and live dock components were re-rendering too frequently

**Solutions Implemented**
 _New Hooks_ (app/components-react/hooks/realm.ts)
- Added useRealmProperty - selectively subscribe to individual realm properties
- Added useRealmProperties - subscribe to multiple specific properties
- These mirror the existing useVuex hook pattern for consistency

_Component Refactoring_ 
- Refactored Main.tsx (primary change: +84/-44 lines) to use new hooks
- Updated LiveDock.tsx, SideNav.tsx, and other frequently-rendered components
- Replaced broader state subscriptions with targeted property subscriptions

_Bug Fixes_
- Fixed chat resize bug by adding debounce and cleanup checks to prevent stale values


app/components-react/modals/onboarding/Onboarding.tsx
```diff
@@ -35,9 +35,11 @@ export default function Onboarding() {
...
   const [processing, setProcessing] = useState(false);
 
-  const currentStep = useRealmObject(OnboardingV2Service.state).currentStep;
-  const currentIndex = useRealmObject(OnboardingV2Service.state).currentIndex;
-  const showOnboarding = useRealmObject(OnboardingV2Service.state).showOnboarding;
+  const { currentStep, currentIndex, showOnboarding } = useRealmProperties({
+    currentStep: prop(OnboardingV2Service.state, 'currentStep'),
+    currentIndex: prop(OnboardingV2Service.state, 'currentIndex'),
+    showOnboarding: prop(OnboardingV2Service.state, 'showOnboarding'),
+  });
```
wesrupert, line 42: Out of curiousity, is there a benefit to this approach vs the `useRealmObjectProperties` multi-key signature? From my read of realm.ts changes, this would also be a valid call:

```typescript
const { currentStep, currentIndex, showOnboarding } = useRealmObjectProperty(
  OnboardingV2Service.state,
  'currentStep',
  'currentIndex',
  'showOnboarding'
);
```

Comment applies throughout, so I'll just leave it on this one.

app/components-react/editor/elements/mixer/MixerItem.tsx
```diff
@@ -6,14 +6,14 @@ import { useVuex } from 'components-react/hooks';
...
-import { useRealmObject } from 'components-react/hooks/realm';
+import { useRealmObjectProperty } from 'components-react/hooks/realm';
 
 export default function MixerItem(p: { audioSourceId: string; volmetersEnabled?: boolean }) {
   const volmetersEnabled = p.volmetersEnabled ?? true;
 
   const { CustomizationService, EditorCommandsService, SourcesService, AudioService } = Services;
 
-  const performanceMode = useRealmObject(CustomizationService.state).performanceMode;
+  const { performanceMode } = useRealmObjectProperty(CustomizationService.state, 'performanceMode');
```
wesrupert, line 16: Nit, but throughout - this gets pretty verbose in some files, especially when renaming the tracked value. I'd recommend two changes:

1. Rename `useRealmObjectProperty` -> `useRealmProperty`, to match `useRealmProperties`
2. If there is only one key (or just make this the "one key only" version, for simplicity/maintainability), return the value directly.

Together, this line would become:
```typescript
const performanceMode = useRealmProperty(CustomizationService.state, 'performanceMode');
```

---

The benefits are better realized in other files, like [app/components-react/windows/Settings.tsx](https://github.com/streamlabs/desktop/pull/5858/changes#diff-3fce5274326da1ea04576a5aa901a4fe4e36fc31ede4972b3fea8c3ed01ea6c9):

This:
```typescript
const { currentSettingsTab: currentTab } = useRealmObjectProperty(
  NavigationService.state,
  'currentSettingsTab',
);
```

Becomes this:
```typescript
const currentTab = useRealmProperty(NavigationService.state, 'currentSettingsTab');
```

wesrupert (approved):
LGTM, with two small suggestions

PR #5861, opened 2026-04-23
Fix dual stream with factory API.

michelinewu:
Fixes for:

- [Crashing every time streaming dual mode to YT ](https://app.asana.com/1/1083097041131/project/1207748235152481/task/1214164136932701?focus=true)
- [Stream crashes immediately after going live with Twitch both enabled](https://app.asana.com/1/1083097041131/task/1214164136821739)
- [Vertical Recording Records Horizontal](https://app.asana.com/1/1083097041131/task/1214214613788960)

### DIFF
--- a/app/services/streaming/streaming.ts
+++ b/app/services/streaming/streaming.ts
@@ -299,9 +299,12 @@ export class StreamingService
       // Update the replay buffer duration if the setting has been changed while active
       if (patch?.Output?.RecRBTime && this.isReplayBufferActive) {
         for (const contextName in this.contexts) {
-          if (this.contexts[contextName as TDisplayType].replayBuffer !== null) {
-            this.contexts[contextName as TDisplayType].replayBuffer.duration =
-              patch.Output.RecRBTime;
+          const display = contextName as TDisplayType;
+          if (
+            this.isAdvancedReplayBuffer(this.contexts[display].replayBuffer) ||
+            this.isSimpleReplayBuffer(this.contexts[display].replayBuffer)
+          ) {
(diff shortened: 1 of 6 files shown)


app/services/streaming/streaming.ts
```diff
@@ -3375,17 +3441,64 @@ export class StreamingService
...
+
+        try {
+          this.destroyFactoryInstance(context, type, instance);
+        } catch (e: unknown) {
+          console.error(`Error destroying ${type} for ${context} during shutdown:`, e);
         }
-      }),
-    );
+
+        this.contexts[context][type] = (null as unknown) as (
```
wesrupert: This cast has a code smell to me. Is there a reason we don't have the `| null` type annotation on `this.contexts[c][t]`, if it's expected to get nulled out here? Where is `contexts` defined?

michelinewu: `contexts` is defined on line [192](https://github.com/streamlabs/desktop/blob/2d271afba98536ee8f68b75dd829229a227864d2/app/services/streaming/streaming.ts#L192) to be used as a private store for the instances. The casting is for strictnulls. Without it there is the error:
`TS2352: Conversion of type 'null' to type '(ISimpleStreaming | IAdvancedStreaming) & ((ISimpleRecording | IAdvancedRecording) & (ISimpleReplayBuffer | IAdvancedReplayBuffer))' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.`

gettinToasty (approved):

PR #5863, opened 2026-04-24
Fix apply settings for delay, reconnect, network, and replay buffer duration.

michelinewu:
Fixes for:

- [UI does not react appropriately when reconnecting](https://app.asana.com/1/1083097041131/task/1214236759482283)
- [Stream Delay is not Counting down all the way](https://app.asana.com/1/1083097041131/task/1214214613788968)
- Apply network settings correctly
- Update replay buffer duration while replay buffer is active

### DIFF
--- a/app/services/settings/settings.ts
+++ b/app/services/settings/settings.ts
@@ -705,23 +705,25 @@ export class SettingsService extends StatefulService<ISettingsServiceState> {
     // This function represents a cleaner API we would like to have
     // in the future.
 
-    (Object.entries(patch) as [TCategoryName, Dictionary<any>][]).forEach(([categoryName, category]) => {
-      const formSubCategories = this.fetchSettingsFromObs(categoryName).formData;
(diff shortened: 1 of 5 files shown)


app/services/settings/settings.ts
```diff
@@ -698,7 +705,7 @@ export class SettingsService extends StatefulService<ISettingsServiceState> {
     // This function represents a cleaner API we would like to have
     // in the future.
 
-    Object.keys(patch).forEach((categoryName: TCategoryName) => {
+    (Object.keys(patch) as TCategoryName[]).forEach((categoryName: TCategoryName) => {
       const category: Dictionary<any> = patch[categoryName];
```
wesrupert: Nit this cast should have stronger typing:
```suggestion
    (Object.entries(patch) as [TCategoryName, Dictionary<any>][]).forEach(([categoryName, category]) => {
```

Image

wesrupert (approved):

PR #5878, opened 2026-04-29
Add YouTube dual stream error handling and chat link.

michelinewu:
# Fix YouTube Dual Stream & Enhanced Broadcasting Signal Handling

## Issue

YouTube dual streaming and Twitch Enhanced Broadcasting had several interacting bugs in dual output mode:

1. **YouTube dual stream start order** — The vertical YouTube stream was being started before the horizontal stream, which caused it to be treated as the primary stream in the YouTube mobile app (the horizontal stream should always start first so the vertical stream appears as the mobile-optimized view).

2. **Broken signal handling for Enhanced Broadcasting + dual output combinations** — The `handleStartDualOutputStream` signal handler had no dedicated path for Twitch Enhanced Broadcasting alongside other platforms. The logic was inlined and covered only a narrow set of cases, silently misbehaving when Twitch dual stream was enabled alongside another platform, or when Twitch used single-display Enhanced Broadcasting while another display also had targets.

3. **Missing vertical display Live status** — In several Enhanced Broadcasting dual output flows, the non-Enhanced Broadcasting display instance started successfully but its `EStreamingState.Live` status was never set, leaving the UI stuck at `Starting`.

---

#### Signal handler refactor (`f885ab175`)

- Extracted `handleStartEnhancedBroadcastingDualOutput` — a shared handler for all Enhanced Broadcasting dual output start signals.
- Added `displayNeedsNonEnhancedBroadcastingInstance(display)` helper to determine which displays need their own streaming instance alongside Enhanced Broadcasting.
- Dual streaming has per platform handling: less DRY but easier for debugging in the future.
- Improved `handleStopStreaming` to gracefully recover when no horizontal/vertical streaming instances exist (resets stale non-Offline statuses rather than always surfacing an error).

#### YouTube Dual Stream vertical stream link (`ae308f1b5`, `911dc1269`)

- Added optional `verticalStreamPageUrl` to `IPlatformService` interface.
- Updated `ShareStreamLink` to include the vertical YouTube watch URL as a second copyable link when YouTube is dual streaming.


app/services/platforms/youtube.ts
```diff
@@ -570,6 +576,13 @@ export class YoutubeService
       );
     }
 
+    console.log(
+      'YouTube stream key set in settings, streamKey: ',
+      streamKey,
+      ' context: ',
+      context,
+    );
+
```
gettinToasty: couple of logs but nothing blocking

gettinToasty (approved):

gettinToasty (approved):

PR #5904, opened 2026-05-07
revert replay-buffer save

sandboxcoder:
Richard Osborne	Revert "osn-replay-buffer: invoke output handler to save replays (#1689)". Fix issue where the same replay file is overwritten.


sandboxcoder:
It turns out the bug is in a different spot in code

PR #5908, opened 2026-05-11
Move check for global rescale before target setup.

michelinewu:
# Move global rescale check when going live

## Issue
In `StreamingService.goLive()`, `dualOutputService.disableGlobalRescaleIfNeeded()` was being called after platform and custom destination targets had already been configured When a user had "Rescale Output" enabled globally and then went live in dual output mode, the rescale setting was applied while building the per-target video contexts for the vertical/horizontal displays — causing scaling issues on targets using the vertical display before being disabled too late to take effect.

## Fix
Removed all checks to disable global rescale settings in dual output mode to allow global rescale for streaming instances the use the horizontal display and disable only for streaming instances that use the vertical display. This is a temporary solution until backend changes are made for the vertical display to have its own global rescale settings.

### DIFF
--- a/app/services/dual-output/dual-output.ts
+++ b/app/services/dual-output/dual-output.ts
@@ -11,7 +11,7 @@ import { ScenesService, SceneItem, TSceneNode } from 'services/scenes';
 import { TDisplayType, VideoSettingsService } from 'services/settings-v2/video';
 import { TPlatform } from 'services/platforms';
 import { Subject } from 'rxjs';
-import { IVideoInfo } from 'obs-studio-node';
+import { EScaleType, IVideoInfo } from 'obs-studio-node';
 import { ICustomStreamDestination, StreamSettingsService } from 'services/settings/streaming';
 import {
   ISceneCollectionsManifestEntry,
@@ -505,9 +505,17 @@ export class DualOutputService extends PersistentStatefulService<IDualOutputServ
         'Streaming',
         'Rescale',
       );
-      if (globalRescaleOutput) {
+      const globalRescaleFilter = this.settingsService.findSettingValue(
+        output,
+        'Streaming',
+        'RescaleFilter',
+      );
+      if (globalRescaleOutput || globalRescaleFilter !== EScaleType.Disable) {
         // `Output` not a typo, it is different from above
-        this.settingsService.setSettingValue('Output', 'Rescale', false);
+        this.settingsService.setSettingsPatch({
+          Output: { Rescale: false, RescaleFilter: EScaleType.Disable },
+        });
+
         // TODO: find a cleaner way to make dual output recalculate its settings for the vertical display
         // since even after disabling "rescale output" its settings persists, and looks stretched.
         this.settingsService.refreshVideoSettings();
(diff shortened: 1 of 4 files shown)


aleksandr-voitenko:
This workaround should work.

The only concern I have:
The `disableGlobalRescaleIfNeeded()` clears only legacy `Rescale`. Recently a modern version of this option was introduced called `RescaleFilter `.

app/services/dual-output/dual-output.ts
```diff
@@ -505,9 +505,17 @@ export class DualOutputService extends PersistentStatefulService<IDualOutputServ
...
+      const globalRescaleFilter = this.settingsService.findSettingValue(
+        output,
+        'Streaming',
+        'RescaleFilter',
+      );
+      if (globalRescaleOutput || globalRescaleFilter !== EScaleType.Disable) {
         // `Output` not a typo, it is different from above
-        this.settingsService.setSettingValue('Output', 'Rescale', false);
+        this.settingsService.setSettingsPatch({
+          Output: { Rescale: false, RescaleFilter: EScaleType.Disable },
```
aleksandr-voitenko: This fixes the vertical leak, but it also clears the user's global streaming rescale for the horizontal output. In Dual Output, a user may intentionally rescale horizontal to 852x480 while expecting vertical to remain portrait. Can we avoid mutating the saved global Output settings here and instead ignore rescale only when building the vertical streaming settings?

app/services/dual-output/dual-output.ts
```diff
@@ -505,9 +505,17 @@ export class DualOutputService extends PersistentStatefulService<IDualOutputServ
         'Streaming',
         'Rescale',
       );
-      if (globalRescaleOutput) {
+      const globalRescaleFilter = this.settingsService.findSettingValue(
+        output,
+        'Streaming',
+        'RescaleFilter',
+      );
+      if (globalRescaleOutput || globalRescaleFilter !== EScaleType.Disable) {
```
aleksandr-voitenko: If `RescaleFilter` is missing or unavailable, this condition treats `undefined` as enabled rescale. Could we guard that explicitly, e.g. `globalRescaleFilter !== undefined && globalRescaleFilter !== EScaleType.Disable`, so older/partial settings do not trigger unnecessary settings writes and refreshes?

app/services/dual-output/dual-output.ts
michelinewu: Refactored to ignore rescale only when building the vertical streaming instance

app/services/dual-output/dual-output.ts
michelinewu: Removed all global rescale checks in favor of ignore rescale only when building vertical streaming instance

app/services/dual-output/dual-output.ts
```diff
@@ -11,7 +11,7 @@ import { ScenesService, SceneItem, TSceneNode } from 'services/scenes';
 import { TDisplayType, VideoSettingsService } from 'services/settings-v2/video';
 import { TPlatform } from 'services/platforms';
 import { Subject } from 'rxjs';
-import { IVideoInfo } from 'obs-studio-node';
+import { EScaleType, IVideoInfo } from 'obs-studio-node';
```
sandboxcoder, line 14: (trivial/optional) we can remove this change

aleksandr-voitenko (approved):

sandboxcoder (approved):

PR #5949, opened 2026-05-28
Fix highlighter recording broken by twitch game_id instead of game_name

jankalthoefer:
With a previous change the `this.streamingService.views.game ` now passes a _game_id_ instead of the _game_name_. 
I now added the game_name as an additional views variable to the _streamingService_, which is only used in the _highlighterService_ now.

### DIFF
--- a/app/services/streaming/streaming-view.ts
+++ b/app/services/streaming/streaming-view.ts
@@ -110,6 +110,15 @@ export class StreamInfoView<T extends Object> extends ViewHandler<T> {
     );
   }
 
+  get gameName() {
+    return (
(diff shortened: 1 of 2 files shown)


app/services/streaming/streaming-view.ts
```diff
@@ -110,6 +110,10 @@ export class StreamInfoView<T extends Object> extends ViewHandler<T> {
     );
   }
 
+  get gameName() {
+    return (this.platforms.twitch?.enabled && this.platforms.twitch.gameName) || '';
+  }
+
```
jankalthoefer, line 121: Assuming that game is now consistently storing only gameIds I added gameName here with only twitch.

gettinToasty (approved):

PR #5956, opened 2026-06-02
MacOS: fix setInputValue to work + enable multiple tests to run

sandboxcoder:
Follow up to my [pull request](https://github.com/streamlabs/desktop/pull/5818) in which laid the ground work to run tests to run on MacOS. However, any test that interacted with an input field would not work properly because the text field was not being properly cleared. This PR addresses this and another issue where after the first test passed, the test runner would stall.

* fix `setInputValue()` to work properly on MacOS. Now it properly clears the pre-existing text. Legacy Windows code path is unaffected.
* `runner-utils.js`: fix issue on MacOS, after first recording.js test completes, the runner would lock up (no other tests run). This is a shared code path, apparently Windows somehow dodged this situation? I am not  certain if tests might occassionally stall on Windows or not but I definitely observed this behavior on macOS.
* `test/helpers/webdriver/index.ts`: fix error: ` ✖ afterEach.always hook for Recording The log-file has errors.` so all recording tests will pass.

## How Was this tested?
All tests pass on MacOS when I run `yarn test:file test-dist/test/regular/recording.js` & `yarn test:file test-dist/test/regular/obs-importer.js`. This PR will not fix all tests to run on MacOS (some, afterall, are testing the Windows-only plugins etc)- but does enable a great many to run. This PR makes it much more feasible to write tests on MacOS.

### DIFF
--- a/main.js
+++ b/main.js
@@ -1,6 +1,22 @@
 const appStartTime = Date.now();
 let lastEventTime = 0;
 
+// On macOS, writing to a closed stdout/stderr pipe throws EPIPE. Replace write with a noop
+process.stdout.on('error', err => {
+  if (err.code === 'EPIPE') {
+    process.stdout.write = () => true;
+  } else {
+    throw err;
+  }
+});
+process.stderr.on('error', err => {
+  if (err.code === 'EPIPE') {
+    process.stderr.write = () => true;
+  } else {
+    throw err;
+  }
+});
+
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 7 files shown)


aleksandr-voitenko:
The `test/helpers/form-monkey.ts` helper still uses `Control+a` / `Backspace` to clear inputs. On macOS, that selection step may not select the field contents, so `FormMonkey`-based tests can still append instead of replacing text.

aleksandr-voitenko (approved):

PR #5985, opened 2026-06-18
Add shutdown watchdog for stalled Electron workers

aleksandr-voitenko:
Adds a coordinated, bounded shutdown flow across Electron’s main process and worker renderer. The main process waits up to 10 seconds for acknowledgement, 30 seconds for worker teardown, and 10 seconds for Electron to exit after completion, force-exiting when a phase stalls.

Worker shutdown now persists local scene and session state before destructive OBS teardown, continues remaining steps after individual failures, bounds analytics flushing to three seconds, and reports whether shutdown completed cleanly. File flushing now supports concurrent shutdown callers and writes that race an in-flight flush.

If a second instance starts during shutdown, its arguments are preserved and one relaunch is scheduled; the active watchdog continues to bound the original process’s exit. Scene-collection cloud synchronization is deferred to startup so network stalls cannot block shutdown.


sandboxcoder (approved):

summeroff (approved):

main.js
```diff
@@ -272,10 +275,21 @@ app.on('ready', () => {
...
 
+function forceShutdown(reason) {
+  console.warn(`[Shutdown] Force exiting application: ${reason}`);
+  allowMainWindowClose = true;
+
+  BrowserWindow.getAllWindows().forEach(window => {
+    if (!window.isDestroyed()) window.destroy();
+  });
+
+  app.exit(0);
```
summeroff, line 290: is 0 intentional? 
it is not exactly clean exit, but not a full failure too.

main.js
```diff
@@ -462,19 +481,35 @@ async function startApp() {
...
+  app.on('quit', () => {
+    shutdownCoordinator.finishShutdown();
+  });
+
   ipcMain.on('acknowledgeShutdown', () => {
-    if (appShutdownTimeout) clearTimeout(appShutdownTimeout);
+    shutdownCoordinator.acknowledgeShutdown();
   });
 
   ipcMain.on('shutdownComplete', () => {
```
summeroff, line 492: as we now send additional information with this message. 
```
          electron.ipcRenderer.send('shutdownComplete', {
            clean: report.clean,
            failedSteps: report.failures.map(failure => failure.name),
          });
```
probably be useful to log it? 

```
ipcMain.on('shutdownComplete', (event, payload) => {
  const { clean, failedSteps } = payload;

```

PR #5995, opened 2026-06-25
Show per-display bitrate in dual output mode during streaming

aleksandr-voitenko:
> [!NOTE]
> This PR's target branch is `mw_fix_btrt` 

> [!WARNING]
> This PR depends on https://github.com/streamlabs/obs-studio-node/pull/1738 

image

## Motivation and Context

Desktop currently shows a single combined bitrate metric in Performance Metrics. That is confusing in dual output mode because horizontal and vertical outputs can have different bitrate behavior, and the combined value does not make it clear what each display is actually sending.

This also matters for Twitch Enhanced Broadcasting / Dual Format. `obs-studio-node` now exposes per-display enhanced broadcasting stats, so `desktop` can present the horizontal and vertical bitrate separately instead of relying only on an aggregate stream value.

Single-output streaming should keep the existing presentation.

## Changes

- Added streaming performance stats helpers for aggregating active output instances.
- Added optional per-display bitrate/data-output aggregation for dual output mode.
- Uses Enhanced Broadcasting `displayStats` when available.
- Falls back to output context mapping for normal dual streaming across platforms.
- Keeps the existing combined bitrate metric when dual output mode is disabled.
- Updates Performance Metrics bandwidth display to `H: <num> V: <num>` in dual output mode.
- Added tests for streaming stats aggregation and Performance Metrics metadata.

### DIFF
--- a/app/components-react/shared/performance-metrics.ts
+++ b/app/components-react/shared/performance-metrics.ts
@@ -1,4 +1,5 @@
 import { IPinnedStatistics } from 'services/customization';
+import { $t } from 'services/i18n';
 
 export interface IPerformanceMetricValues {
   cpuPercent: string;
@@ -21,18 +22,15 @@ export interface IPerformanceMetricMetadata {
   icon: string;
 }
 
-type TTranslate = (key: string, params?: Record<string, unknown>) => string;
-
 export function getPerformanceMetricMetadata(
   values: IPerformanceMetricValues,
-  translate: TTranslate,
 ): IPerformanceMetricMetadata[] {
   const metrics: IPerformanceMetricMetadata[] = [
     {
       key: 'cpu',
       attribute: 'cpu',
       value: `${values.cpuPercent}%`,
-      label: translate('CPU'),
+      label: $t('CPU'),
       icon: 'icon-cpu',
     },
     {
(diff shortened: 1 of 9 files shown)


app/components-react/shared/performance-metrics.ts
```diff
@@ -0,0 +1,73 @@
...
+  value: string | number;
+  label: string;
+  icon: string;
+}
+
+type TTranslate = (key: string, params?: Record<string, unknown>) => string;
+
+export function getPerformanceMetricMetadata(
+  values: IPerformanceMetricValues,
+  translate: TTranslate,
```
michelinewu: Instead of passing in the translate function we can use the `$t` function directly from the `i18n` service

app/services/streaming/streaming.ts
```diff
@@ -3514,38 +3523,33 @@ export class StreamingService
...
+        display,
+        displayStats,
+        droppedFrames: instance.droppedFrames,
+        totalFrames: instance.totalFrames,
+        kbitsPerSec: instance.kbitsPerSec,
+        dataOutput: instance.dataOutput,
+      });
     }
 
     // TODO: Add UI to show bitrate by display but for now average the all instances, which is more accurate
```
michelinewu: Nitpick: this TODO can be removed I think

test/regular/components-react/performance-metrics.ts
```diff
@@ -0,0 +1,52 @@
+import test from 'ava';
+import { getPerformanceMetricMetadata } from '../../../app/components-react/shared/performance-metrics';
+
+const translate = (key: string) => key;
+
+test('performance metric metadata uses one combined bandwidth metric outside dual output mode', t => {
```
michelinewu: Nitpick to capitalize the first letter in the test name so it looks more like a title

michelinewu (changes requested):

michelinewu (approved):

PR #6000, opened 2026-06-29
[Highlighter] Highlighter settings update

jankalthoefer:
- In highlighter settings it's possible to install Replay now
- small fix for the progress update for the installation
- a small styling change for the FeatureCarousel
- wording change


app/components-react/highlighter/SettingsView.tsx
```diff
@@ -112,6 +133,30 @@ export default function SettingsView({
...
+          onChange={handleToggleHighlighter}
+        />
+      );
+    }
+    return (
+      <Button
+        style={{ width: 'fit-content' }}
+        type="primary"
+        onClick={() => {
+          HighlighterService.actions.installAiHighlighter(true, 'Highlighter-tab');
```
michelinewu, line 152: Outside of scope nitpick: for analytics we're trying to move to typing the various strings in the usage statistics service to make it simpler to reference.

jankalthoefer: But only for the Event names or also for every property sent to the analtics?

gettinToasty (approved):

PR #6029, opened 2026-07-17
Consolidate smart source websocket subscriptions. (Rebased)

michelinewu:
# Consolidate smart source websocket subscriptions

>[!NOTE]
> Replaces [Consolidate smart source websocket subscriptions. #5899](https://github.com/streamlabs/desktop/pull/5899)

## Issue

Each `SmartBrowserSourceManager` independently subscribed to the websocket and called `visionService.ensureRunning()` on init. With N smart browser sources active, this created N parallel RxJS subscriptions to the same `visionEvent`/`userStateUpdated` events and N redundant Vision startup calls.

## Fix

Moved event forwarding and Vision startup out of `SmartBrowserSourceManager` and into `ReactiveDataService`, which already held a single socket subscription:

- `ReactiveDataService` now forwards `visionEvent`/`userStateUpdated` events to every smart source via `forwardEventToSources()`.
- Vision startup is centralized behind `ensureVisionRunning()`, which reads `isRunning`/`isStarting` from local state and only calls `visionService.ensureRunning()` when Vision is not already up.
- `SmartBrowserSourceManager` loses its own `socketSub` subscription, its `WebsocketService` injection, and its `destroy()` unsubscribe.

### Smart-source lookup helpers (`SourcesService`)

Added two helpers used by the centralized path:

- `hasSmartSources()` — cheap existence check (`Array.some` on `propertiesManagerType === 'smartBrowserSource'`); used to guard Vision startup on login and on scene-collection switch so we don't start Vision for collections with no smart sources.
- `getSmartSources()` — filters the source models by `propertiesManagerType` **before** constructing `Source` instances, so the hot fan-out path (`forwardEventToSources`, which can fire frequently while live) avoids building `Source` objects for non-smart sources.

## What changed

| Area | Before | After |
|---|---|---|
| Socket subscriptions for vision/user-state events | N (one per smart source) | 1 (in `ReactiveDataService`) |
| Event delivery to sources | each source subscribes and messages itself | one subscription fans out via `getSmartSources()` → `obsSource.sendMessage()` |
| `visionService.ensureRunning()` on startup | N (once per source init) | guarded; fired at login / collection-switch (only when `hasSmartSources()`) and on first source registration, skipped when Vision is already running/starting |
| `destroy()` / unsubscribe overhead | N (one `socketSub.unsubscribe()` per source) | 0 |

All of the above runs in the **worker** window — the invisible renderer that hosts the services layer and the OBS backend. `WebsocketService`, `VisionService`, `SourcesService`, and `ReactiveDataService` are all singleton services, and `socketEvent` is a plain RxJS `Subject` on `WebsocketService`, so these subscriptions and `sendMessage()` / `ensureRunning()` calls are **in-process RxJS / OBS calls, not per-source IPC**. The win is eliminating redundant in-process subscriptions and duplicate Vision startup work, plus a single centralized fan-out — not a reduction in IPC crossings.

## Files touched

- `app/services/reactive-data/index.ts` — centralized subscription + forwarding, `ensureVisionRunning()`, `forwardEventToSources()`, login/collection-switch guards.
- `app/services/sources/properties-managers/smart-browser-source-manager.ts` — removed per-source subscription, injection, and `destroy()`.
- `app/services/sources/sources.ts` — added `getSmartSources()` and `hasSmartSources()`.


app/services/reactive-data/index.ts
```diff
@@ -115,6 +118,11 @@ export class ReactiveDataService extends Service {
         this.writeState({ schemaFlatJson: JSON.stringify(schemaFlat) });
       });
 
+      // If the active scene collection has smart sources, ensure Vision (aka Streamlabs AI) is active on login
+      if (this.sourcesService.hasSmartSources()) {
+        this.ensureVisionRunning();
+      }
+
```
wesrupert: From offline conversation, this can be removed now.

It is now handled here: [app/services/vision/index.ts](https://github.com/streamlabs/desktop/blob/e4fba6ca2f7ca420f6ec4432cc0c1de983c69ed9/app/services/vision/index.ts#L143-L164)

```suggestion
```

app/services/reactive-data/index.ts
```diff
@@ -134,13 +142,18 @@ export class ReactiveDataService extends Service {
 
       if (!this.userService.isLoggedIn) return;
 
+      if (this.sourcesService.hasSmartSources()) {
+        this.ensureVisionRunning();
+      }
+
```
wesrupert:
```suggestion
```

app/services/reactive-data/index.ts
```diff
@@ -208,6 +221,8 @@ export class ReactiveDataService extends Service {
               sourceId: key,
             });
 
+            this.ensureVisionRunning();
+
```
wesrupert:
```suggestion
```

app/services/reactive-data/index.ts
```diff
@@ -243,6 +258,38 @@ export class ReactiveDataService extends Service {
...
+   */
+  private ensureVisionRunning() {
+    if (!this.visionService.state.isRunning && !this.visionService.state.isStarting) {
+      // Don't await the `ensureRunning` promise since we don't need to wait for it to finish here
+      this.visionService.ensureRunning().catch(e => {
+        console.error('Error validating if Vision (aka Streamlabs AI) is running:', e);
+      });
+    }
+  }
+
```
wesrupert:
```suggestion
```

app/services/reactive-data/index.ts
```diff
@@ -6,6 +6,7 @@ import {
   UserService,
   WebsocketService,
 } from 'app-services';
+import { VisionService } from 'services/vision';
```
wesrupert:
```suggestion
```

app/services/reactive-data/index.ts
```diff
@@ -91,6 +93,7 @@ export class ReactiveDataService extends Service {
   @Inject() websocketService: WebsocketService;
   @Inject() sourcesService: SourcesService;
   @Inject() sceneCollectionsService: SceneCollectionsService;
+  @Inject() visionService: VisionService;
```
wesrupert:
```suggestion
```

wesrupert (approved):

sandboxcoder (approved):

PR #6063, opened 2026-08-04
Provision the shared graphics hook directory from the installer

summeroff:
Fixes the installer half of HackerOne #3319516 (local privilege escalation, High / 7.8).

`%ProgramData%\obs-studio-hook` holds the graphics hook we inject into the processes we capture, and it is registered as an implicit Vulkan layer, so the loader maps whatever sits there into every process that renders. Releases up to 1.21 leave it writable by `BUILTIN\Users`, and a directory created by an unelevated run is owned outright by whichever account got there first — on this machine, without the app ever having been run elevated, the owner was a standard user with Full control. The app runs unelevated and cannot repair that. We are the elevated part of the install, so `customInstall` provisions the directory instead.

## What it does

- Creates the directory with the final protected descriptor already attached — `O:BA D:PAI` granting Full to SYSTEM and Administrators, Read/Execute to Users, `ALL APPLICATION PACKAGES` and `ALL RESTRICTED APPLICATION PACKAGES`. Not `CreateDirectory(..., NULL)` followed by `icacls`: one created under `%ProgramData%` inherits an entry letting users write to it, and that would stand for as long as it takes to spawn a process.
- Moves an existing directory aside rather than repairing it in place. Rewriting a DACL does not revoke handles opened before the change, so whoever created it could otherwise rename the hardened directory away afterwards and put their own back under the same name. Only a directory this run created is accepted.
- Installs `graphics-hook{32,64}.dll` and `obs-vulkan{32,64}.json` from the signed installer payload rather than from `$INSTDIR` — the user chooses that path, so a standard user may be able to rewrite it while this elevated process is running.
- Removes the Vulkan layer registration if anything could not be put in place, rather than leaving it pointing at a directory we cannot stand behind.

## Notes for review

- Unlike the app and the updater, NSIS has no plugin here that can tell an administrator-provisioned directory from a planted one, so the quarantine happens whenever a directory exists. That means the installer cannot take part in newest-hook-wins arbitration with other OBS-derived applications — nothing survives to arbitrate against. The updater settles that on its next run, and runs far more often than this does.
- `icacls` is invoked by full path, since a bare name would be resolved against the search path in an elevated process, and never with `/T`: recursing through a directory standard users could write until a moment ago means walking whatever links they left, and `icacls` follows link targets.
- Ownership and the DACL are reset only for a file we actually wrote. Doing it for one we merely found would make somebody else's file look administrator-installed to the app, which decides what to trust on exactly that basis.
- The shell's all-users `$APPDATA` is used rather than the `ProgramData` environment variable, which is inherited from whoever launched us.

## Ship order

Pairs with streamlabs/obs-studio#756 (app side: validate before trusting, quarantine an untrusted directory) and streamlabs/a-files-updater#140 (repairs the installed base, since NSIS does not re-run on auto-update).

**This should not ship ahead of the obs-studio change.** Today's `add_aap_perms()` re-grants `BUILTIN\Users` write on the hook directory on any elevated app launch, which would undo what this installer just did. Merge after obs-studio-node carries the matching hook and `repositories.json` is bumped.

## Testing

Verified on an elevated end-to-end install of 1.22.0-preview.1 (Windows 11 26200, x64) against a deliberately staged vulnerable directory — ownership reassigned away from Administrators to a user account, `BUILTIN\Users` granted Full, and `graphics-hook64.dll` replaced with a marker file. How the directory got into that state does not matter to the code path: the quarantine keys on the directory existing, not on who owns it.

Before:

```
Owner : <machine>\<user>
C:\ProgramData\obs-studio-hook NT AUTHORITY\SYSTEM:(OI)(CI)(F)
                               BUILTIN\Administrators:(OI)(CI)(F)
                               BUILTIN\Users:(OI)(CI)(F)
                               APPLICATION PACKAGE AUTHORITY\ALL APPLICATION PACKAGES:(OI)(CI)(RX)
                               APPLICATION PACKAGE AUTHORITY\ALL RESTRICTED APPLICATION PACKAGES:(OI)(CI)(RX)

graphics-hook64.dll  7 bytes  sha256=1D29EE7BEDD603FDB36DA6F156C363AA29011A95CAE04A78B92D0726E8049123
```

After:

```
Owner  : BUILTIN\Administrators
SDDL   : O:BA...D:PAI(A;OICI;FA;;;SY)(A;OICI;FA;;;BA)(A;OICI;0x1200a9;;;BU)(A;OICI;0x1200a9;;;AC)(A;OICI;0x1200a9;;;S-1-15-2-2)
Reparse: False

C:\ProgramData\obs-studio-hook NT AUTHORITY\SYSTEM:(OI)(CI)(F)
                               BUILTIN\Administrators:(OI)(CI)(F)
                               BUILTIN\Users:(OI)(CI)(RX)
                               APPLICATION PACKAGE AUTHORITY\ALL APPLICATION PACKAGES:(OI)(CI)(RX)
                               APPLICATION PACKAGE AUTHORITY\ALL RESTRICTED APPLICATION PACKAGES:(OI)(CI)(RX)

graphics-hook32.dll  248680 bytes  owner=BUILTIN\Administrators  sha256=5BEA21AB3A4DF3FA35AF3B203EB59FEAFC4A193A4BA121EA0F8D42833E6CF104
graphics-hook64.dll  303464 bytes  owner=BUILTIN\Administrators  sha256=F89ADE59480DE0D55FCC100495D78E8E491BED469FAB4DBD3D430837B58B27A5
obs-vulkan32.json       514 bytes  owner=BUILTIN\Administrators  sha256=C16E0707AE66AD71E8A0720AEB6E6997A1017F19762333452AEF692115A9AB41
obs-vulkan64.json       514 bytes  owner=BUILTIN\Administrators  sha256=7B6B0813FB58B276847A8583EB5C3F94AEE7D7AD0AE3A1EF6133D5D8771F20F4
```

- The descriptor is protected (`D:PA`, no inherited ACEs) and owned by Administrators; `Users` is demoted to Read/Execute; both AppContainer SIDs are kept at Read/Execute.
- The marker is gone — all four files carry the payload hashes and administrator ownership.
- The directory's creation time advanced across the install, and no `obs-studio-hook.quarantine*` survives.
- A Basic User restricted token (`runas /trustlevel:0x20000`) is denied `OpenWrite` on `graphics-hook64.dll`.

The sequence itself comes from the USN journal, because `DetailPrint` and `nsExec::ExecToLog` produce nothing observable in a release build — electron-builder sets `ShowInstDetails nevershow` and `SetDetailsPrint none`:

```
Rename: old name    obs-studio-hook
Rename: new name    obs-studio-hook.quarantine0
File delete         obs-studio-hook.quarantine0
File create         obs-studio-hook              <- new file reference number
Security change     obs-studio-hook              <- /setowner
Security change     obs-studio-hook              <- /inheritance:r /grant:r
```

The new file reference number is the load-bearing part: the directory the app ends up trusting is one this install created, not the staged one with its DACL rewritten.

Earlier checks, unchanged:

- The `CopyFileW` and `SECURITY_ATTRIBUTES` marshalling were exercised against a real `makensis` build (`good=1 missing=0`), since NSIS `System::Call` gets these wrong silently.
- `mklink /J` confirms directory junctions need no elevation, which is what the reparse-point check and the no-`/T` rule are there for.
- `icacls /setowner` on a file owned by another account fails without `SeRestorePrivilege`, which is what makes remove-protection → tamper → restore-protection incomplete.

Not covered: the `Rename /REBOOTOK` staging path for a hook mapped into a running process, and the junction branch end to end — only the `mklink /J` premise it rests on was checked.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/installer.nsh
+++ b/installer.nsh
@@ -1,11 +1,9 @@
 !include MUI2.nsh
 
-; Ownership and the DACL are reset only for a file we actually wrote. Doing it
-; for one we merely found would take somebody else's file and make it look
-; administrator-installed to the app, which decides what to trust on exactly
-; that basis. CopyFileW can carry explicit ACEs from the source, so changing
-; only the owner is not enough: /reset makes the file inherit the protected
-; directory descriptor.
+; Only ever for a file we just wrote. On one we merely found, this would make
+; somebody else's file look administrator-installed to the app, which decides
+; what to trust on exactly that basis. /reset as well as /setowner, because
+; CopyFileW can carry explicit ACEs over from the source.
 !macro SecureHookFile FileName
   nsExec::ExecToLog '"$SYSDIR\icacls.exe" "$R7\${FileName}" /setowner "*S-1-5-32-544" /Q'
   Pop $R8
@@ -22,29 +20,21 @@
   ${EndIf}
 !macroend
 
-; Copies one graphics hook file into the locked-down shared directory ($R7)
-; from the install directory ($R9), setting $R6 if the file could not be put in
-; place right now.
+; Copies one graphics hook file from $R9 into the locked-down shared directory
+; $R7, setting $R6 if it could not be put in place right now.
 ;
-; Each destination is deleted before it is written, and the copy refuses to
-; overwrite. An entry left there while the directory was writable may be a hard
-; link sharing its data with some other file, and copying onto it would put our
-; bytes into that file - an arbitrary write, since the installer is elevated.
-; Deleting removes the directory entry, not the link target.
+; $R9 is the installer's own plugin directory, extracted from the signed
+; installer. Do not copy these out of $INSTDIR: the user chooses that path, so
+; a standard user may be able to rewrite it while we are running elevated.
 ;
-; CopyFiles takes a destination directory rather than a destination file, which
-; the staged path needs, so this goes through CopyFileW directly for a real
-; return value. The target can be loaded by another process at this very moment
-; - the vulkan layer pulls the hook into anything that renders - so a locked
-; file is staged beside it and swapped in by Windows on the next reboot. That
-; swap is link-safe: it replaces the directory entry rather than writing
-; through it, and the file we leave behind until then keeps whoever put it
-; there as its owner, so the app will not use it.
+; Delete first, then copy without overwrite. An entry left there while the
+; directory was writable may be a hard link, and writing through it would put
+; our bytes into the link target - an arbitrary write, since we are elevated.
+; CopyFileW rather than CopyFiles, which takes a destination directory and
+; gives no usable return value.
 ;
-; The source is embedded in the signed installer and extracted to its private
-; plugin directory. Do not copy these files out of $INSTDIR: the installer lets
-; the user choose that path, so a standard user may be able to rewrite it while
-; this elevated process is running.
+; The vulkan layer pulls the hook into anything that renders, so the target may
+; be loaded right now; a locked file is staged beside it and swapped on reboot.
 !macro InstallHookFile FileName
   Delete "$R7\${FileName}"
   System::Call 'kernel32::CopyFileW(w "$R9\${FileName}", w "$R7\${FileName}", i 1) i .s'
@@ -69,19 +59,15 @@
 ; Moves an existing hook directory aside instead of repairing it where it
 ; stands, and records the result in $R4 (empty if nothing was moved) and $R6.
 ;
-; Rewriting a DACL does not revoke handles that are already open. Whoever
-; created the directory can hold one with delete access, let us harden it, and
-; rename it away afterwards - then put their own back at the same path. The app
-; would reject what it finds there, but the vulkan loader does not repeat its
-; checks. A directory that did not exist when they opened their handle is not
-; reachable that way.
+; An ACL change does not revoke handles opened before it, so whoever created
+; the directory could rename the hardened one away afterwards and put their own
+; back at the same path.
 ;
-; Unlike the app and the updater we cannot tell an administrator-provisioned
-; directory from a planted one - that needs an ownership check, and NSIS has no
-; plugin for it here - so this happens whenever one exists. It does mean the
+; This happens whenever a directory exists, because unlike the app and the
+; updater we cannot tell an administrator-provisioned one from a planted one -
+; that needs an ownership check and NSIS has no plugin for it here. So the
 ; installer cannot take part in the newest-hook-wins arbitration the other two
-; do: nothing survives to arbitrate against. The updater sorts that out on its
-; next run, and it runs far more often than this does.
+; do; the updater settles that on its next run.
 !macro QuarantineHookDir
   StrCpy $R4 ""
   System::Call 'kernel32::GetFileAttributesW(w "$R7") i .s'
@@ -104,9 +90,8 @@
       StrCpy $R6 "unverified"
     ${Else}
       DetailPrint "Moved the existing hook directory to $R4"
-      ; Only the names we know. RMDir /r would follow a junction left inside
-      ; and delete whatever is on the other side of it. Anything else in there
-      ; keeps the directory alive, which is untidy but harmless.
+      ; Only names we own. RMDir /r would follow a junction left inside and
+      ; delete whatever is on the other side of it.
       Delete /REBOOTOK "$R4\graphics-hook32.dll"
       Delete /REBOOTOK "$R4\graphics-hook64.dll"
       Delete /REBOOTOK "$R4\obs-vulkan32.json"
@@ -141,11 +126,10 @@
   ${EndIf}
 
   ; The graphics hook is injected into other processes out of a shared
-  ; directory, so only administrators may write to it. We are the elevated
-  ; part of the install; the app itself runs unelevated and cannot provision
-  ; this. Releases up to 1.21 granted BUILTIN\Users write access here, so an
-  ; existing directory may be owned by, and full of files planted by, a
-  ; standard user.
+  ; directory, so only administrators may write to it. We are the elevated part
+  ; of the install; the app runs unelevated and cannot provision this. Releases
+  ; up to 1.21 granted BUILTIN\Users write access here, so an existing
+  ; directory may be owned by, and full of files planted by, a standard user.
   Push $R1
   Push $R2
   Push $R3
@@ -158,16 +142,15 @@
 
   StrCpy $R6 ""
 
-  ; Not the ProgramData environment variable: that is inherited from whoever
-  ; launched us, so a standard user could point this at a directory of their
-  ; choosing and have an elevated process provision it. Under the all-users
-  ; context $APPDATA is CSIDL_COMMON_APPDATA, which comes from the shell.
+  ; Not the ProgramData environment variable, which is inherited from whoever
+  ; launched us. Under the all-users context $APPDATA is CSIDL_COMMON_APPDATA,
+  ; which comes from the shell.
   SetShellVarContext all
   StrCpy $R7 "$APPDATA\obs-studio-hook"
 
   ; A junction left behind by whoever owned this directory before us would send
-  ; the icacls calls and the copies below to its target instead. RMDir without
-  ; /r unlinks the junction itself and leaves whatever it pointed at alone.
+  ; every step below to its target instead. RMDir without /r unlinks the
+  ; junction and leaves whatever it pointed at alone.
   System::Call 'kernel32::GetFileAttributesW(w "$R7") i .s'
   Pop $R8
   ${If} $R8 <> -1
@@ -192,15 +175,12 @@
   ${EndIf}
 
   ${If} $R6 == ""
-    ; Only accept a directory created by this call. If a standard user wins
-    ; the name after quarantine, hardening their directory would leave any
-    ; delete-capable handle they already opened alive.
-    ;
-    ; The descriptor goes on at creation rather than being fixed up by the
-    ; icacls calls below. A directory created under %ProgramData% inherits an
-    ; entry letting users create files in it, and that would stand for as long
-    ; as it takes to spawn icacls. This is the same descriptor, and the same
-    ; reasoning, as hook_dir_create() in obs-studio's hook-dir-security.h.
+    ; Same descriptor, and the same reasoning, as hook_dir_create() in
+    ; obs-studio's hook-dir-security.h. It goes on at creation rather than being
+    ; fixed up by the icacls calls below: a directory created under
+    ; %ProgramData% inherits an entry letting users create files in it, and that
+    ; would stand for as long as it takes to spawn icacls. A failure here means
+    ; somebody won the name after quarantine, and must not be repaired in place.
     System::Call 'advapi32::ConvertStringSecurityDescriptorToSecurityDescriptorW(w "O:BAD:PAI(A;OICI;FA;;;SY)(A;OICI;FA;;;BA)(A;OICI;FRFX;;;BU)(A;OICI;FRFX;;;AC)(A;OICI;FRFX;;;S-1-15-2-2)", i 1, *p .r13, p 0) i .s'
     Pop $R8
     ${If} $R8 == 0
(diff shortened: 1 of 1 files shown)


installer.nsh
```diff
@@ -20,6 +140,162 @@
...
+    ${If} $R8 == 0
+      DetailPrint "Could not build the hook directory descriptor"
+      StrCpy $R6 "unverified"
+    ${Else}
+      ; SECURITY_ATTRIBUTES {nLength, lpSecurityDescriptor, bInheritHandle};
+      ; 12 bytes because NSIS installers are 32-bit
+      System::Call '*(i 12, p r13, i 0) p .r12'
+      System::Call 'kernel32::CreateDirectoryW(w "$R7", p r12) i .s'
+      Pop $R8
+      System::Free $R2
```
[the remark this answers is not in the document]
summeroff: This one is a false positive — `r12` *is* `$R2`.

The System plugin's register syntax maps `r0`–`r9` to `$0`–`$9` and `r10`–`r19` to `$R0`–`$R9`. So `*(i 12, p r13, i 0) p .r12` stores the `SECURITY_ATTRIBUTES` pointer into `$R2`, and `System::Free $R2` releases exactly that allocation. By the same mapping `r13` is `$R3`, which holds the security descriptor and is released by the `LocalFree(p r13)` on the following line.

Rather than take that from documentation, I checked it against the same `makensis` this project builds with:

```nsis
StrCpy $R2 "R2-untouched"
StrCpy $2  "2-untouched"
System::Call '*(i 12, p 0, i 0) p .r12'
; -> $R2 = 7843168
; -> $2  = 2-untouched
```

`$R2` receives the pointer; `$2` is untouched. No change made.

The other comment was right and is fixed in e0c9541 — the uninstaller now resolves the same `SetShellVarContext all` + `$APPDATA` path the install side provisions. While there I also stopped it recursing through a reparse point: `RMDir /r` follows a junction and deletes what is on the other side, and the uninstaller is elevated, so a directory left behind by a release that let standard users write to it is exactly the case to be careful about.

installer.nsh
```diff
@@ -20,6 +125,154 @@
...
+    nsExec::ExecToLog '"$SYSDIR\icacls.exe" "$R7" /inheritance:r /grant:r "*S-1-5-18:(OI)(CI)(F)" "*S-1-5-32-544:(OI)(CI)(F)" "*S-1-5-32-545:(OI)(CI)(RX)" "*S-1-15-2-1:(OI)(CI)(RX)" "*S-1-15-2-2:(OI)(CI)(RX)"'
+    Pop $R8
+    ${If} $R8 <> 0
+      DetailPrint "Could not secure $R7"
+      StrCpy $R6 "unverified"
+    ${EndIf}
+  ${EndIf}
+
+  ${If} $R6 == ""
+    SetOutPath "$PLUGINSDIR\obs-studio-hook"
```
aleksandr-voitenko, line 240: This PR treats `$PLUGINSDIR` as trusted staging for hook manifests/DLLs while the installer runs elevated. The problem is that it can also be compromised due to bugs in the older NSIS versions.

Our current version has a known $PLUGINSDIR creation race: a standard user can win creation of the temporary directory and plant files subsequently loaded with administrator privileges. NSIS 3.11 fixed this race, and 3.12 fixed an additional elevated/SYSTEM temporary-directory issue.

Please see `3.12` and `3.11` security sections here:
https://nsis.sourceforge.io/Docs/AppendixF.html#v3.12

If we do not involve low level hacking, the proper fix is to upgrade the electron-builder version. We need at least `26.12.0` which introduced NSIS version bundle selection. Then update the builder config to use
```js
const base = {
  toolsets: {
    nsis: '1.2.1',
  },
  ...
```
NSIS toolset `1.2.1` maps to NSIS `3.12` which has both required security fixes.
https://github.com/electron-userland/electron-builder-binaries/releases/tag/nsis%401.2.1

summeroff: good reason to revive https://github.com/streamlabs/desktop/pull/5694

installer.nsh
```diff
@@ -47,8 +300,26 @@ Function un.LeaveUnWelcome
...
+    System::Call 'kernel32::GetFileAttributesW(w "$APPDATA\obs-studio-hook") i .s'
+    Pop $1
+    ${If} $1 <> -1
+      IntOp $1 $1 & 0x400 ; FILE_ATTRIBUTE_REPARSE_POINT
+      ${If} $1 <> 0
+        RMDir "$APPDATA\obs-studio-hook"
+      ${Else}
+        RMDir /r /REBOOTOK "$APPDATA\obs-studio-hook"
+      ${EndIf}
+    ${EndIf}
```
aleksandr-voitenko, line 357: Uninstall steps recreate the original privilege-escalation condition.

Deleting`$APPDATA\obs-studio-hook` directory undoes the ACL fix. Once the name is free, a standard user can recreate `%ProgramData%\obs-studio-hook` and put malicious manifests/DLLs there.

`C:\ProgramData` normally allows standard users to create new subdirectories. While obs-studio-hook exists with an administrator-owned, non-writable ACL, a standard user cannot replace its manifests or DLLs. If an elevated uninstaller deletes it, the pathname becomes unclaimed:
```
Protected directory deleted
        ↓
Standard user recreates obs-studio-hook
        ↓
New directory/files are controlled by that user
        ↓
Any surviving Vulkan registration now points to attacker-controlled code
```

More details:
The dangerous leftover is the machine-wide Vulkan registration.

1. While Streamlabs runs elevated, it creates an enabled registry value similar to:

   ```text
   HKLM\SOFTWARE\Khronos\Vulkan\ImplicitLayers
     C:\ProgramData\obs-studio-hook\obs-vulkan64.json = DWORD 0
   ```

   The code registers both architecture variants in the corresponding registry views. [Registration code](https://github.com/streamlabs/obs-studio/blob/5953b92b9c24410007add7581f8fb971336aa9b5/plugins/win-capture/game-capture-file-init.c#L270-L330)

2. Uninstall deletes `C:\ProgramData\obs-studio-hook`, including its protected ACL, but leaves those HKLM values behind. [Uninstaller code](https://github.com/streamlabs/desktop/blob/e0c9541d5e678170f46ae7829eec490cd4520b23/installer.nsh#L314-L322)

3. A standard user can now recreate that directory under the normal `C:\ProgramData` permissions. Because they created it, they control its contents.

4. The attacker places:

   ```text
   C:\ProgramData\obs-studio-hook\obs-vulkan64.json
   C:\ProgramData\obs-studio-hook\malicious-layer.dll
   ```

   The JSON identifies a valid Vulkan implicit layer and names `malicious-layer.dll` as its library.

5. Later, any elevated 64-bit application that initializes Vulkan causes the Vulkan loader to scan the machine-wide `ImplicitLayers` registry key. A DWORD value of zero is enabled; the loader opens the JSON named by the registry value and loads its referenced DLL. Elevated applications ignore user-level layer registrations, but they still trust HKLM registrations. [Khronos loader documentation](https://github.com/KhronosGroup/Vulkan-Loader/blob/main/docs/LoaderLayerInterface.md#windows-layer-discovery)

6. The malicious DLL consequently executes inside that unrelated Vulkan application with its elevated privileges.

So the uninstall creates a vulnerable dormant state that can be triggered later.

**The fix:**
Either remove both registry-view registrations first or leave the directory and files protected.

installer.nsh
```diff
@@ -47,8 +300,26 @@ Function un.LeaveUnWelcome
...
+    ; are elevated here. A directory left behind by a release that let standard
+    ; users write to it may be one.
+    System::Call 'kernel32::GetFileAttributesW(w "$APPDATA\obs-studio-hook") i .s'
+    Pop $1
+    ${If} $1 <> -1
+      IntOp $1 $1 & 0x400 ; FILE_ATTRIBUTE_REPARSE_POINT
+      ${If} $1 <> 0
+        RMDir "$APPDATA\obs-studio-hook"
+      ${Else}
+        RMDir /r /REBOOTOK "$APPDATA\obs-studio-hook"
```
aleksandr-voitenko: Another reason to not delete directories recursively, and do not delete protected `obs-studio-hook` at all is the ability of malicious user to recreate `obs-studio-hook` with a junction.

A Windows directory junction is a special folder entry that redirects access to another directory.

```text
C:\ProgramData\obs-studio-hook\       ← normal directory
└── payload\                          ← junction
    redirects to C:\SomeSensitiveFolder
```

The installer checks only `obs-studio-hook` itself, so it appears safe. But `RMDir /r` enters `payload`, follows the redirect, and can delete files from `C:\SomeSensitiveFolder` using the uninstaller’s elevated permissions.

The safest fix is to remove `RMDir /r from uninstall`; never traverse an unverified shared directory; preserve the hardened container and, only after verifying it is trusted, delete exact known files.

summeroff:
the `ImplicitLayers` values now go first, in both registry views, and the files are deleted by exact name instead of recursively, a recursive delete really does destroy the target of a nested junction. 

The directory itself is fully removed rather than kept hardened: it is only ours while we are installed, and other apps on same obs code would fail the copy into an admin-owned dir, so a leftover would break any other OBS-derived app Vulkan capture for good.

installer.nsh
```diff
@@ -1,5 +1,110 @@
...
+      DetailPrint "Could not install ${FileName} into $R7"
+    ${Else}
+      !insertmacro SecureHookFile "${FileName}.new"
+      Rename /REBOOTOK "$R7\${FileName}.new" "$R7\${FileName}"
+      DetailPrint "${FileName} could not be replaced now; it is staged for the next reboot"
+    ${EndIf}
+    StrCpy $R6 "unverified"
+  ${Else}
+    !insertmacro SecureHookFile "${FileName}"
+  ${EndIf}
```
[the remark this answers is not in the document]
summeroff: `CopyFileW` doesn't carry the source's ACEs over, so this can't happen. It
creates the destination fresh, which takes the inheritable entries off the
parent and nothing else — and the destination is always fresh here, since
`Delete` runs first and the copy uses `bFailIfExists=1`, so an existing file
fails the copy rather than being overwritten:

    SOURCE   hook.dll   BUILTIN\Users:(W)            ← explicit write ACE
    DEST DIR            BUILTIN\Users:(OI)(CI)(RX)   ← protected, as the hook dir is
    DEST     hook.dll   <me>:(I)(F)  BUILTIN\Users:(I)(RX)

So a failed `/reset` cannot leave a writable hook — the file only ever has what
the directory grants, which is RX for Users.

You were reading our own comment, though, which claimed the opposite. That was
wrong and is corrected in ebb54ec17. `/setowner` is the call that matters, since
the app decides what to trust by owner.

aleksandr-voitenko (approved):

summeroff:
should be release in preview with this osn release https://github.com/streamlabs/desktop/pull/6063

nbarrett-logitech:
hey, I'm putting together the hackerone update for #3319516 and want to make sure I don't tell them it's fixed before it actually is.

what I can see from the outside:

1. all three merged. obs-studio #756 and a-files-updater #140 on 8/7, this one 8/12
2. but there's been a lot of follow up since. #762 and #763 on obs-studio, and #144 through #152 on the updater, with #152 landing today. #763 and #145 read like an actual bug in the trust check rather than cleanup. there's also a diagnose-hook-dir-trust-failures branch out there
3. both preview tags were cut 8/12, so preview.0 and preview.1 predate all of that

so what I need to know:

1. is the fix done, or are you still chasing failures? the commits landing today are the main thing holding me up
2. which build should h1 retest against? I don't think it's either of the current previews
3. the ship order note in this PR says it shouldn't go ahead of the obs-studio change, or add_aap_perms re-grants Users write on elevated launch and undoes the installer work. the gate was osn carrying the matching hook and repositories.json bumped. I can't confirm that from the repo since package.json points at the floating osn-0.0.0-release.tar.gz on both 1.21.9 and 1.22.0-preview.1. did that land?
4. did the elevated end to end install run happen before merge? description said you'd post the icacls output here first and I don't see it
5. when does an updater build with #140 plus the follow ups actually ship? that's the only elevated code that reaches machines already installed since nsis doesn't re-run on auto update. and any read on 1.22.0 stable? still showing 1.21.9 as latest

no rush on all of it, 1 through 3 are what I need before I can put anything on the report

PR #6075, opened 2026-08-11
Remove stale translations and add missing ones

gettinToasty:
Used Claude to audit our i18n folder and there was a lot of stuff we don't use anymore as well as some things that slipped under the radar. This isn't a 100% holistic fix as there are some strings we get from obs that I did not include in the audit for the sake of expediency, but this should help us and the translation team a lot.


app/i18n/en-US/grow.json
michelinewu, line 1: Since this is being removed, maybe the whole grow tab should be? And if so, maybe split off removing this translations file into that PR?

michelinewu (approved):

PR #6082, opened 2026-08-12
First pass adding copilot instructions to check translations

gettinToasty:


.github/instructions/translations.instructions.md
```diff
@@ -0,0 +1,39 @@
...
+If the same string wrapped in a `$t()` function cannot be found in the `app/i18n/en-US/**` directory, it should be flagged as a missing translation.
+
+# Tracking Stale Translations
+
+## Removed strings
+
+When code reviewing a PR, check for any strings wrapped in a `$t()` function that have been removed in the changes.
+When those strings are removed, they should have a corresponding removal of the same key/value pair in the `app/i18n/en-US/**` directory.
+If no such key/value pair has been removed in the PR, it should be flagged as a stale translation.
+
```
michelinewu, line 35: Should we also check for json files that exist but have not been added to `fallback.ts`?

.github/instructions/translations.instructions.md
```diff
@@ -0,0 +1,39 @@
+applyTo: "app/components/**,app/components-react/**,app/services/**,app/i18n/en-US/**"
+
+Read the entire instructions file before carrying out any checks.
+
+# Tracking Missing Translations
+
+## Plaintext Strings
+
+When code reviewing a PR, check for new plaintext strings added to a tsx or vue template that are not wrapped in a `$t()` function.
+These strings need to be wrapped in a `$t()` function to ensure they are translated properly.
```
michelinewu, line 10: Will this also pick up strings in the Translate component? A plaintext string could accidentally be wraped in `"` like `<Translate message="..."`

michelinewu (changes requested):

.github/instructions/translations.instructions.md
gettinToasty: definitely! good catch :D

.github/instructions/translations.instructions.md
gettinToasty: i think we actually want that to run on the `Translate` component-- even though it is named `Translate` it doesn't actually interact with the i18n layer itself, it takes what should be an already translated string and injects html into it

michelinewu (approved):

wesrupert (approved):

PR #6084, opened 2026-08-13
[WIP] Add logging for non-bug-related test failures in automated tests.

michelinewu:
# Specify details for automated test failure cases

Automated test failure logs were too generic to rule out non-bug related failures.

---

## Issues

**Every shutdown failure reported the same message.**Errors arrive as `unknown` causing webdriver to reject with `Error` objects when sometimes the api-client rejects with plain strings and with serialized error objects. The `catch` in `stopApp` emitted a blanket `'Crash on shutdown'` regardless of what was thrown. That one message covers at least four unrelated conditions:

1. The app never exited when `waitForElectronInstancesExist` timed out.
2. The app exited early so the webdriver session was already gone.
3. The window disappeared before `closeWindow` reached it.
4. the connection to the app's api server dropped. Only this last one is a crash.

**Account-caused failures looked identical to app bugs.**
Failures due to issues from the user pool accounts, rather than app-related bugs, were logged the same in summary, requiring triage by hand in the logs for the specific test failure. This is unnecessarily time consuming. Specific cases:

1. The test failed to obtain an account from the user pool after all retries completed.
2. An account with YouTube is not enabled for live streaming.
3. An account with YouTube cannot stream due to rate limiting.
4. An account with YouTube failed for another account-related reason.

**3. The analytics POST failing was silent.**
`sendJobToAnalytics` wraps its request in a `catch` that logs and swallows it, so the entire body, including failed tests and stats, was discarded with no trace beyond one console line. This meant that test failures due to failure to post analytics was logged as `ECONNABORTED` and not distinguished from test failures due to app bugs.

## Fixes

**Shutdown errors are classified** An error is passed as `unknown` and is now identified so it can be logged in a readable way. Note: the fallback case uses `util.inspect` rather than `String()` or `JSON.stringify`. `String()` renders any object without a `.message` as `[object Object]`, so the message will not actually be logged. `JSON.stringify` is unsafe because it returns the value `undefined` for `undefined`/symbols, and it throws outright on circular references. `util.inspect` handles both and matches how `main.js` and `app/app.ts` already serialize unknown values into the log.

**Account failures are detected and recorded.** A new `TAccountFailure` union names the reasons a run failed because of its accounts:

| Reason                         | Detected from                                                        |
| ------------------------------ | -------------------------------------------------------------------- |
| `NO_ACCOUNT_AVAILABLE`         | flag set in `reserveUserFromPool` where it gives up after 5 attempts |
| `YOUTUBE_STREAMING_DISABLED`   | `liveStreamingNotEnabled` in the app log                             |
| `YOUTUBE_ACCOUNT_RATE_LIMITED` | `userRequestsExceedRateLimit` in the app log                         |
| `YOUTUBE_ACCOUNT_FAILURE`      | token expired / platform validation failure in the app log           |
| `HEROKU_ANALYTICS_SEND_FAILED` | the analytics POST giving up                                         |

The user pool case `NO_ACCOUNT_AVAILABLE` is flagged at its source rather than string-matched because the harness never sees the AVA error. The YouTube cases come from the log because that is the only place the reason surfaces.

Detection reads `currentTestLogs`, which accumulates only the slice written since the current test began because the pre-existing `lastLogs` holds the whole file and leaks across tests whenever the cache directory is reused.

**Results are written to `test-dist/account-failures.json`** For extensibility, write results under a shape shared by both writers, described by the exported `IAccountFailures`:

```json
{
  "tests": { "Go live to YouTube": "YOUTUBE_ACCOUNT_RATE_LIMITED" },
  "job": ["HEROKU_ANALYTICS_SEND_FAILED"]
}
```

`tests` is keyed by test name and written by the harness; `job` holds run-wide reasons and is written by the runner. Both sides read through a normalizer that fills in the missing halves, so an absent, empty, or older flat-shaped file degrades to empty instead of throwing. Job reasons are appended rather than assigned, so an earlier reason is not lost behind a later one.

**Analytics carries the reason.** Each entry in `failedTests` gains an `accountFailure` field, and the body gains a job-level `accountFailure` chosen off an explicit `ACCOUNT_FAILURE_SEVERITY` ordering rather than first-seen.

**File(s) changed:** `test/helpers/webdriver/index.ts`, `test/helpers/webdriver/runner-utils.ts`, `test/helpers/webdriver/user.ts`, `test/helpers/runner.js`

## Performance Implications

None. All work happens in `afterEach.always` on the failure path only, after the test has already failed. Detection is a few regex matches over a string already held in memory, and the file write is a few hundred bytes to `test-dist/`. No change to the passing path.

### DIFF
--- a/test/helpers/runner.js
+++ b/test/helpers/runner.js
@@ -105,13 +105,13 @@ function readTestStats() {
  * @returns - An object with the test failure details
  */
 function readFailureReason() {
-  let failures = {};
   try {
-    failures = JSON.parse(fs.readFileSync(failureReasonsFile, 'utf8'));
+    const failures = JSON.parse(fs.readFileSync(failureReasonsFile, 'utf8'));
+    return { tests: failures?.tests || {}, job: failures?.job || [] };
   } catch (e) {
-    failures = {};
+    console.error('Failed to read the failure reasons file', e);
+    return { tests: {}, job: [] };
   }
-  return { tests: failures.tests || {}, job: failures.job || [] };
 }
 
 async function sendJobToAnalytics(failedTests) {
--- a/test/helpers/runner.js
+++ b/test/helpers/runner.js
@@ -13,6 +13,7 @@ const fetch = require('node-fetch');
 
 const failedTestsFile = 'test-dist/failed-tests.json';
 const testStatsFile = 'test-dist/test-stats.json';
+const accountFailuresFile = 'test-dist/account-failures.json';
 const args = process.argv.slice(2);
 const TIMEOUT = 3; // timeout in minutes
 const {
@@ -24,6 +25,16 @@ const {
   BUILD_DEFINITIONNAME,
   SLOBS_TEST_RUN_CHUNK,
 } = process.env;
+
+// Order the account failure errors from highest to lowest severity
+const ACCOUNT_FAILURE_SEVERITY = [
+  'NO_ACCOUNT_AVAILABLE',
+  'HEROKU_ANALYTICS_SEND_FAILED',
+  'YOUTUBE_ACCOUNT_FAILURE',
+  'YOUTUBE_ACCOUNT_RATE_LIMITED',
+  'YOUTUBE_STREAMING_DISABLED',
+];
+
 let retryingFailed = false;
 
 const RUN_TESTS_CMD = !args.length ? `yarn test --timeout=${TIMEOUT}m ` : args.join(' ') + ' ';
@@ -33,6 +44,7 @@ const RUN_TESTS_CMD = !args.length ? `yarn test --timeout=${TIMEOUT}m ` : args.j
   try {
     rimraf.sync(failedTestsFile);
     rimraf.sync(testStatsFile);
+    rimraf.sync(accountFailuresFile);
     await createTestTimingsFile();
     execSync(RUN_TESTS_CMD, { stdio: [0, 1, 2] });
   } catch (e) {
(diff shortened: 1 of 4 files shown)


test/helpers/webdriver/runner-utils.ts
```diff
@@ -93,6 +115,29 @@ export function removeFailedTestFromFile(testName: string) {
...
+
+/**
+ * Write to the logs test failure that was caused by the test account to indicate that
+ * the failure is not from a bug
+ * @param testName - Name of the test with the failure
+ * @param reason - The account failure reason
+ */
+export function saveAccountFailureToFile(testName: string, reason: TAccountFailure) {
+  const failures = readAccountFailures();
+  failures.tests[testName] = reason;
```
summeroff, line 149: Storing account failures by test name makes the reason survive into the retry run. If the first attempt fails with a YouTube/account issue and the retry then fails for a real app bug, `sendJobToAnalytics()` will still attach the stale `accountFailure` to that unrecovered failure. We should clear this test's account-failure entry before retrying/rerunning it, or record attempts separately, so final failures are not mislabeled as account noise.

test/helpers/runner.js
```diff
@@ -114,12 +159,30 @@ async function sendJobToAnalytics(failedTests) {
...
     slice: SLOBS_TEST_RUN_CHUNK,
     stats: readTestStats(),
+    accountFailure: getJobAccountFailure(testsToSend, accountFailures.job),
   };
   log(body);
   try {
     await requestUtilityServer('job', 'post', body);
   } catch (e) {
     console.error('failed to send analytics', e);
+    saveJobAccountFailure('HEROKU_ANALYTICS_SEND_FAILED');
```
summeroff: This records `HEROKU_ANALYTICS_SEND_FAILED` only after the analytics POST has already failed. The runner reads `account-failures.json` before this request, and the Azure job does not publish `test-dist`, so this reason never reaches analytics or any artifact; it only lands in a local file that disappears with the job. Either include this condition in something observable before/when the POST fails, or drop the file write and rely on the existing console error.

PR #6091, opened 2026-08-14
More selectively kick off CI runs

gettinToasty:
This PR intends to limit automated CI runs of somethings, notably renovate PRs which we can manually run before merging and Crowdin PRs as well as changes to config files which do not cause changes to the outcome of tests

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -2,6 +2,22 @@ trigger:
   - staging
   - master
 
+pr:
+  branches:
+    include:
+      - staging
+      - master
+  paths:
+    include:
+      - '*'
+    exclude:
+      - package.json
+      - yarn.lock
+      - app/i18n/*
+      - .github/*
+      - '*.md'
+      - '*.yml'
+
 variables:
   CI: true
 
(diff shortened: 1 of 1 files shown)


wesrupert (approved):

aleksandr-voitenko (approved):

azure-pipelines.yml
```diff
@@ -2,6 +2,22 @@ trigger:
   - staging
   - master
 
+pr:
+  branches:
+    include:
```
michelinewu, line 7: By only including targeting master and staging, this means CI won't run on branches targeting other branches, so will never run on stacked PRs. Highlighting because we should decide this is definitely something we want to do

michelinewu (changes requested):

azure-pipelines.yml
wesrupert: I'm a big fan of stacked PRs so I'd miss the extra branch CI, but _in theory_ GH's stacked PR support will operate off of the same branch, so if we ever get into the beta/it releases to open audience, this will be a moot point.

michelinewu (approved):

PR #6099, opened 2026-08-19
feat(automations): tier-based enable limit + Ultra/Ultra+ upgrade modal

AnkhHeart:
## Why

Automations shipped with no limit and no paywall — any user can enable unlimited automations. This gates the **enabled** count by subscription tier:

| Tier | Max enabled | Over the limit |
| --- | --- | --- |
| Free | 5 | Upgrade-to-Ultra modal |
| Ultra | 10 | Upgrade-to-Ultra+ modal |
| Ultra+ | 30| Toast, no upsell (hard cap) |

**Nothing is ever refused outright.** Only *turning an automation on* is blocked — an automation created over the cap is still created, just switched off. No user work is lost.

### `tier` was already on the wire, just being dropped

`/api/v5/slobs/prime` returns `tier`, but `setPrimeStatus()` declared the response shape inline (twice) and omitted the field. Lifted it into a named `IPrimeStatusResponse`, added `tier?: string`, and store it via a new `SET_TIER` mutation. No new network call.

- `UserViews.tier` returns `'free'` when logged out, and falls back to `isPrime ? 'ultra' : 'free'` if the field is absent, so free/Ultra stays correct either way.
- Also set on the `streamlabs_prime_subscribe` socket event, so the cap lifts right after purchase without a restart.
- `isPrime` is untouched — no existing call site changes.

### One shared helper, three call sites

`automations-limits.tsx` holds the limits and `checkEnableLimit()`, which returns `false` and has *already* shown either the `promptAction` upgrade modal or the hard-cap toast.

| Path | Behaviour |
| --- | --- |
| Row toggle | Blocks turning **on** only; turning off is never gated |
| Editor save | New automation saves disabled when over cap |
| Template bulk add | Prompts once up front, creates the overflow disabled |

Plus a `3 of 5 enabled` counter and an "Upgrade for more" link in the filter bar, mirroring the Collab Cam guest counter.

### Ultra+ checkout

`linkToPrime` gained an optional `tier` in its config. When present it short-circuits to `/ultra?checkout=1&tier=…&plan=month&refl=…&os=…`, since the dashboard magic link only ever resolves to the plain Ultra plan. `recordUltra` still fires first so attribution is unchanged. Purely additive — the five existing callers pass no `tier` and behave exactly as before.

### Drive-by

`Modals.m.less` had an `.action-modal-icon` class written for `promptAction`'s icon slot that nothing had ever used — no caller had passed an `icon` before. It was sized for a font icon (`18px`, `vertical-align: super`), which stretched the SVG and pushed it off the baseline. Rewrote it as `inline-flex`/`middle` with a proper gap, wired it up inside `promptAction`, and added a sibling rule that indents the message **only when an icon is present** so it lines up under the title. The five icon-less callers are unaffected.

## Needs backend confirmation

- **`ULTRA_PLUS_TIER = 'ultra_plus'`** is used both to compare against `UserService.views.tier` and as the `tier` param in the checkout URL. The checkout side is confirmed; if the prime endpoint reports a different string for an Ultra+ account, the cap silently stays at 5. One constant to change.
- **`stream-avatar-api` `AutomationController.ts:43`** has `tierBasedEnabled = tier === "ultra" ? 10 : 30`, which is inverted vs. this spec. Server limits should be reconciled separately.

## Out of scope

Server enforcement. `fetchAll()` can still hydrate more enabled automations than the cap (created on web or another machine) and the engine will run all of them. This is a UX guardrail, not security.

## Test plan

- [x] `yarn typecheck` — passes clean
- [x] `yarn eslint` — passes clean
- [x] **Free:** enable 5, header reads `5 of 5 enabled`; a 6th shows the Ultra modal and the switch stays off
- [x] "Upgrade to Ultra" opens the dashboard magic link with `refl=slobs-automations`; "Not now" closes with no leftover style blocker (OBS preview must not paint over the modal)
- [x] Creating a 6th via the editor: saves, lands in the list **disabled**, modal shown
- [x] Templates: select 8 with 3 slots free → 3 enabled, 5 disabled, exactly one modal
- [x] Disabling is never blocked at any count
- [x] **Ultra:** same at 10; modal says Ultra+ and opens `?checkout=1&tier=ultra_plus&plan=month`
- [x] **Ultra+:** toast at 30, no upgrade CTA (force via `UserViews.tier` until backend ships it)
- [x] Confirm the raw `/api/v5/slobs/prime` response actually carries `tier` on an Ultra account


app/components-react/windows/stream-avatar-automations/EditAutomations.tsx
```diff
@@ -100,6 +108,8 @@ export default function EditAutomations() {
 
   function toggleEnabled(automation: TAutomationExport) {
     if (!automation.id) return;
+    // Turning one off is always allowed; only turning one on is capped.
+    if (!automation.enabled && !checkEnableLimit(1, 'toggle')) return;
```
wesrupert, line 112: Confirming that this limit is also enforced on the backend?

app/services/magic-link.ts
```diff
@@ -53,14 +53,22 @@ export class MagicLinkService extends Service {
...
+    config?: { redirectToCheckout?: boolean; event?: TAnalyticsEvent; tier?: string },
   ) {
     // TODO: this is only here to accommodate ultra checkout A/B test requiring OS
     // remove this and the parameter from {getDashboardMagicLink} after.
     const os = byOS({ [OS.Windows]: 'windows', [OS.Mac]: 'mac' });
 
     this.usageStatisticsService.recordUltra(refl, config?.event);
 
+    // Tiers above Ultra go straight to checkout — the dashboard magic link only
+    // resolves to the plain Ultra plan.
```
wesrupert: The magic link actually _does_ support ultra plus now! Igor added that support when I updated it to skip the select plan dialog. We just didn't need the support in desktop at the time, so I didn't wire it up on desktop.

You can add the following to magic-link.ts to add support:
```diff
diff --git a/app/services/magic-link.ts b/app/services/magic-link.ts
index e6d392399..869c0d38d 100644
--- a/app/services/magic-link.ts
+++ b/app/services/magic-link.ts
@@ -21,20 +21,23 @@ export class MagicLinkService extends Service {
   @Inject() hostsService: HostsService;
   @Inject() usageStatisticsService: UsageStatisticsService;
 
-  async getDashboardMagicLink(subPage = '', source?: string, os?: string) {
+  async getDashboardMagicLink(subPage = '', source?: string, os?: string, tier?: string) {
     const token = (await this.fetchNewToken()).login_token;
-    // TODO: we really need `qs` or similar
-    const sourceString = source ? `&refl=${source}` : '';
-    const osString = os ? `&os=${os}` : '';
 
     if (subPage === 'multistream') {
       // TODO: remove this if statement when multistream settings are implemented
       return `https://${this.hostsService.streamlabs}/content-hub/post/how-to-multistream-the-ultimate-guide-to-multistreaming?login_token=${token}`;
     }
 
-    return `https://${this.hostsService.streamlabs}/slobs/magic/dashboard?login_token=${token}&r=${
-      subPage ?? ''
-    }${sourceString}${osString}`;
+    const params = new URLSearchParams({
+      login_token: token,
+      r: subPage ?? '',
+      ...(source ? { refl: source } : {}),
+      ...(os ? { os } : {}),
+      ...(tier ? { tier } : {}),
+    });
+
+    return `https://${this.hostsService.streamlabs}/slobs/magic/dashboard?${params.toString()}`;
   }
 
   private fetchNewToken(): Promise<ILoginTokenResponse> {
```

app/components-react/windows/stream-avatar-automations/EditAutomations.tsx
AnkhHeart: Yes

wesrupert (approved):

app/components-react/windows/stream-avatar-automations/AutomationEditor.tsx
```diff
@@ -81,7 +82,9 @@ export default function AutomationEditor({ initial, onClose }: Props) {
       ],
   );
   const actions = rows.filter(r => r.action.type).map(r => r.action);
-  const enabled = initial?.enabled ?? true;
+  // New automations start enabled unless that would exceed the tier cap, in which case
+  // checkEnableLimit() has already shown the upgrade modal and we save it disabled.
+  const enabled = initial?.enabled ?? undefined;
```
gettinToasty, line 87: np: not sure why this is being explicitly set as undefined

app/components-react/windows/stream-avatar-automations/EditAutomations.tsx
```diff
@@ -256,6 +271,51 @@ export default function EditAutomations() {
   );
 }
 
+function UsageMeter(p: { count: number; max: number; tier: string }) {
+  const pct = p.max > 0 ? Math.min(100, Math.round((p.count / p.max) * 100)) : 0;
```
gettinToasty, line 275: why would we be sending in 0 or less than 0 as a prop max?

gettinToasty (approved):

PR #6100, opened 2026-08-19
Migrate layouts to relative coordinates

aleksandr-voitenko:
The ultimate goal of the task is:

> Enable safe base canvas resolution changes in settings by introducing relative scene coordinates across `obs-studio`, `obs-studio-node`, and `desktop`, preserving the visual layout of shared horizontal/vertical scene graph.

https://github.com/user-attachments/assets/a544dc8c-0529-45eb-825a-d5c90d67166e

### Description
This is the Desktop part of a three-repository change:

1. `obs-studio` implements canvas-aware relative transforms and crop references.
https://github.com/streamlabs/obs-studio/pull/764
2. `obs-studio-node` exposes the native lifecycle and metadata.
https://github.com/streamlabs/obs-studio-node/pull/1764
3. `desktop` migrates collections and applies Base Resolution changes transactionally.

Previously, scene-item transforms were tied to absolute canvas pixels. Changing the Base (Canvas) Resolution could therefore produce incorrect positioning, scaling, and crop behavior, especially across horizontal and vertical canvases or after saving and reloading a collection.

OBS now stores normalized transforms internally while preserving the existing absolute-pixel API. Desktop must coordinate collection persistence, legacy migration, canvas resizing, output lifecycle, and transform-cache refreshes around that new contract.

### What changed

- Introduced scene collection schema v5, recording:
  - Relative-coordinate mode.
  - Independent horizontal and vertical authored base resolutions.
- Added fail-closed migration for legacy absolute-coordinate collections:
  - Restores the saved base resolution before rebuilding the scene graph.
  - Creates a one-time `.absolute.bak` backup.
  - Awaits nested scene and source loading.
  - Refuses to persist a partial v5 collection when a source is missing or unsupported.
- Made Base (Canvas) Resolution changes transactional:
  - Serializes and coalesces width/height updates.
  - Prevents output startup during the transaction.
  - Invalidates native transform caches and refreshes Desktop’s absolute-pixel models.
  - Persists and flushes the result atomically.
  - Restores resolution, transforms, and persisted data if any step fails.
- Kept canvas resolution changes seamless; no intermediate loading screen is shown.
- Kept Output (Scaled) Resolution independent from scene-item migration.
- Preserved independent horizontal and vertical canvas routing across save/reload.
- Persisted nested-scene crop reference dimensions internally without changing the public scene-transform API.
- Added collection-switch preflight while video output is active:
  - Primary and backup collection files are checked before removing the current scene graph.
  - A collection requiring a different base resolution is rejected while recording or another video output is active.
- Updated output-start coordination for streaming, recording, replay buffer, and virtual camera.
- Updated Desktop to use the coordinated OSN initialization and scene-item APIs.

### User-facing behavior

- Base canvas resizing remains seamless.
- Items scale only with the canvas to which they are assigned.
- Horizontal and vertical canvases retain independent resolutions and transforms.
- Output resolution changes do not move or resize scene items.
- Unsafe canvas changes and collection switches are rejected while video output is active, without disturbing the current scene graph or output.
- Legacy collections are either migrated completely or left unmodified with an absolute backup.

### OSN initialization

Desktop now uses the object-based OSN initializer:

```ts
OBS_API_initAPI({
  language,
  appDataPath,
  version,
  crashServerUrl,
});

### DIFF
--- a/app/components-react/windows/settings/Video.tsx
+++ b/app/components-react/windows/settings/Video.tsx
@@ -60,6 +60,7 @@ class VideoSettingsModule {
   userService = Services.UserService;
   dualOutputService = Services.DualOutputService;
   streamingService = Services.StreamingService;
+  virtualWebcamService = Services.VirtualWebcamService;
   tiktokService = Services.TikTokService;
 
   get display(): TDisplayType {
@@ -67,7 +68,12 @@ class VideoSettingsModule {
   }
 
   get cantEditFields(): boolean {
-    return this.streamingService.views.isStreaming || this.streamingService.views.isRecording;
+    return (
+      this.streamingService.views.isStreaming ||
+      this.streamingService.views.isRecording ||
+      this.streamingService.views.isReplayBufferActive ||
+      this.virtualWebcamService.views.running
+    );
   }
 
   get values(): Dictionary<TInputValue> {
@@ -342,7 +348,14 @@ class VideoSettingsModule {
           settings[`${otherPrefix}Height`] = Number(height);
         }
       }
-      this.service.actions.setSettings(settings, display);
+      Promise.resolve(this.service.actions.setSettings(settings, display)).catch(
+        (error: unknown) => {
+          message.error({
+            content:
+              error instanceof Error ? error.message : $t('Failed to update the video resolution.'),
+          });
+        },
+      );
     }
   }
 
--- a/app/components-react/windows/settings/Video.tsx
+++ b/app/components-react/windows/settings/Video.tsx
@@ -336,14 +336,12 @@ class VideoSettingsModule {
           settings[`${otherPrefix}Height`] = Number(height);
         }
       }
-      Promise.resolve(this.service.actions.setSettings(settings, display)).catch(
-        (error: unknown) => {
-          message.error({
-            content:
-              error instanceof Error ? error.message : $t('Failed to update the video resolution.'),
-          });
-        },
-      );
+      this.service.actions.return.setSettings(settings, display).catch((error: unknown) => {
+        message.error({
+          content:
+            error instanceof Error ? error.message : $t('Failed to update the video resolution.'),
+        });
+      });
     }
   }
 
--- a/app/services/scene-collections/nodes/root.ts
+++ b/app/services/scene-collections/nodes/root.ts
@@ -5,16 +5,16 @@ import { TransitionsNode } from './transitions';
 import { HotkeysNode } from './hotkeys';
 import { NodeMapNode } from './node-map';
 import { Inject } from 'services/core';
-import { VideoService } from 'services/video';
 import { StreamingService } from 'services/streaming';
 import { OS } from 'util/operating-systems';
 import { GuestCamNode } from './guest-cam';
 import { VideoSettingsService } from 'services/settings-v2/video';
 import { DualOutputService } from 'services/dual-output';
-import { SettingsService } from 'services/settings';
-import { SceneCollectionsService } from '../scene-collections';
+import { ISceneCollectionLoadContext } from './load-session';
+import { SceneCollectionOperationalError } from '../errors';
 
 interface ISchema {
+  relativeCoordinates: boolean;
   /**
    * this is for backward compatibility with single output collections
    */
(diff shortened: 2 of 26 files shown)

### CODE REVIEW

File: app/services/scene-collections/nodes/root.ts [hunk 1]
```diff
@@ -159,7 +145,21 @@ export class RootNode extends Node<ISchema, {}> {
     }
     // Added multiple displays with individual base resolutions in version 4
     if (version < 4) {
-      this.data.baseResolutions = this.videoSettingsService.baseResolutions;
+      this.data.baseResolutions = {
```
michelinewu, line 148: Instead of modifying `version < 4`, add 5, otherwise anyone on version 4 won't get the new assignment. for backwards compatibility if switching between versions, you may need `version === 4` instead but I'm not 100% sure

app/services/scene-collections/nodes/sources.ts
```diff
@@ -383,15 +393,18 @@ export class SourcesNode extends Node<ISchema, {}> {
...
+        );
       }
 
       this.sourceFiltersService.loadFilterData(sourceInfo.id, sourceInfo.filters.items);
     });
 
-    return new Promise(resolve => {
-      Promise.all(promises).then(() => resolve());
-    });
+    return Promise.all(promises).then(() => undefined);
```
michelinewu: Why return `undefined` here to resolve the promise?

app/components-react/windows/settings/Video.tsx
```diff
@@ -329,7 +336,14 @@ class VideoSettingsModule {
           settings[`${otherPrefix}Height`] = Number(height);
         }
       }
-      this.service.actions.setSettings(settings, display);
+      Promise.resolve(this.service.actions.setSettings(settings, display)).catch(
+        (error: unknown) => {
+          message.error({
+            content:
+              error instanceof Error ? error.message : $t('Failed to update the video resolution.'),
```
michelinewu: Needs to be added to translations

app/components-react/windows/settings/Video.tsx
```diff
@@ -329,7 +336,14 @@ class VideoSettingsModule {
           settings[`${otherPrefix}Height`] = Number(height);
         }
       }
-      this.service.actions.setSettings(settings, display);
+      Promise.resolve(this.service.actions.setSettings(settings, display)).catch(
```
michelinewu: You may need `actions.return` for the async behavior

app/services/scene-collections/scene-collections.ts
```diff
@@ -571,84 +636,218 @@ export class SceneCollectionsService extends Service implements ISceneCollection
...
+    if (!loadedData || !loadedResult) {
+      await this.create({ auto: true });
+      return;
+    }
 
-      // Everything was successful, write a backup
-      this.stateService.writeDataToCollectionFile(id, data, true);
+    // The app cannot load without a default scene.
+    if (this.scenesService.views.scenes.length === 0) {
+      console.error('Scene collection was loaded but there were no scenes.');
```
michelinewu, line 749: Should this be surfaced to the user somehow? Maybe a noty in the notifications area?

app/services/scenes/scene.ts
```diff
@@ -464,46 +464,47 @@ export class Scene {
...
         recordingVisible: sceneNode.recordingVisible!,
         scaleFilter: sceneNode.scaleFilter!,
         blendingMode: sceneNode.blendingMode!,
-        blendingMethod: sceneNode.blendingMethod,
-        display: sceneNode.display,
-      });
+        blendingMethod: sceneNode.blendingMethod!,
+      };
+
+      obsSceneItems[sceneNode.id] = this.getObsScene().add(source.getObsInput(), transform, video);
```
michelinewu: This adds additional calls to obs, which may have a performance impact for large collections. Not sure if there's a way to batch like `addItems` previously did?

michelinewu (changes requested):

app/services/scene-collections/nodes/root.ts
aleksandr-voitenko: baseResolutions was introduced in schema v4, so v4 collections already contain both resolutions. The < 4 block correctly constructs that field only for older schemas.

app/services/scene-collections/nodes/sources.ts
aleksandr-voitenko: The concern is valid, I'll rewrite this portion of code.

app/components-react/windows/settings/Video.tsx
aleksandr-voitenko: Agree.

app/components-react/windows/settings/Video.tsx
aleksandr-voitenko: Agree

app/services/scene-collections/scene-collections.ts
aleksandr-voitenko: Yes, NotificationsService will work well here.

app/services/scenes/scene.ts
aleksandr-voitenko: OSN’s `addItems` is not a native batch operation. It is a JavaScript loop that calls Input.fromName() and scene.add() for every item. The PR still performs the same number of scene-add calls and may perform fewer total calls because it assigns the canvas during creation instead of using a subsequent SetCanvas.

Caching `const obsScene = this.getObsScene()` outside the loop is a reasonable micro-optimization.

michelinewu (approved):

PR #6110, opened 2026-08-26
Fix stream shift.

michelinewu:
# Fix Stream Shift Losing Its Session Identifier and Switching the Wrong Device

## Issues

**The stream key went out without its `sid` when a switch was accepted.** `setupIngest` appends `&sid=<uuid>` to the restream key, and that identifier is how the server tells the two devices apart for the rest of the switch. It was gated on `isStreamShiftMode` alone, but the accept path runs `confirmStreamShift`, which clears the status *before* going live. By the time `setupIngest` ran, neither the setting nor a `pending` status necessarily said "stream shift", so the branch was skipped and the stream went live with an ordinary key. Everything downstream that matched on the id had nothing to match.

**`fetchTargetData` deleted targets it could not enrich.** It rebuilt `streamShiftTargets` with a `reduce` that only pushed a target when the response came back with data.

**`getPlatformMode` could return `portrait` during a switch.** Stream shift can only work in single output mode.

**Removing every target at once ended the stream.** In `runUpdateStreamSettings` the removal always ran before the add. When an update replaced *all* live targets, the restream session was momentarily empty, the server ended it, and the add that followed failed against a stream that no longer existed.

Also refactored socket event handling to the service layer.

## Fixes

**`setupIngest` now recognises a pending switch.**

**`fetchTargetData` maps instead of reducing**, keeping every target and merging data only when the endpoint returned data. Relay targets pass through untouched, and a missing status row no longer removes a platform from the switch.

**`getPlatformMode` always returns `landscape` for stream shift**

**Switch protocol moved into `RestreamService`.** Aside from expected handling of this kind of logic, a forced go live no longer opens an empty dialog.

**Removal is deferred when nothing survives the update.**

**`checkIsLive` is called from `GoLiveSettingsModule.init()`** which prevents a state update on unmounted component error.

**Files changed:** `app/services/restream.ts`, `app/services/streaming/streaming.ts`,
`app/services/streaming/streaming-view.ts`, `app/services/diagnostics.ts`,
`app/components-react/root/StartStreamingButton.tsx`,
`app/components-react/windows/go-live/GoLiveWindow.tsx`,
`app/components-react/windows/go-live/useGoLiveSettings.ts`,
`app/components-react/shared/DisplaySelector.tsx`,
`app/components-react/windows/go-live/GoLive.m.less`, `app/i18n/en-US/stream-shift.json`

10 files, +357/−167.

## Performance Implications

Net Reduction: Removes work from the main window by `StartStreamingButton` no longer subscribing to `streamShiftEvent` with a closure over four pieces of Vuex state. `checkIsLive` moving into `init()` means one status request per Go Live window open.

Cost: `fetchTargetData` now maps over the full target list rather than a filtered subset, which is a handful of extra iterations on an array that is at most a few entries long.

## Notes

**This is the base of a stack.** The remaining live output editing work from `mw_multistream_toggle_4` sits in eleven further PRs on top of this one. Nothing here depends on them.

### DIFF
--- a/app/components-react/windows/go-live/GoLiveWindow.tsx
+++ b/app/components-react/windows/go-live/GoLiveWindow.tsx
@@ -66,9 +66,7 @@ function ModalFooter() {
     isPrime,
     isStreamShiftMode,
     hasIncompatibleCodec,
-    streamShiftStatus,
     codec,
-    checkIsLive,
     forceStreamShiftGoLive,
     goLiveWithDefaultCodec,
     showSettings,
@@ -93,10 +91,6 @@ function ModalFooter() {
 
     get streamShiftStatus() {
       return module.streamShiftStatus;
-    },
-
-    async checkIsLive() {
-      return this.restreamService.actions.return.checkIsLive();
     },
 
     async forceStreamShiftGoLive() {
@@ -146,14 +140,6 @@ function ModalFooter() {
   const [isCoolingDown, setIsCoolingDown] = useState(false);
   const isStreamShiftPromptShown = useRef(false);
 
-  // Check stream shift status on mount for Prime users
-  useEffect(() => {
-    if (!isPrime) return;
-    checkIsLive().catch((e: unknown) => {
-      console.error('Error checking stream shift status on mount:', e);
-    });
-  }, []);
-
   const promptUseDefaultCodec = useCallback(async () => {
     // If the user is not live but has an incompatible codec, prompt to change codec
     let message = $t(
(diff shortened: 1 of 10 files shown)


app/components-react/windows/go-live/GoLiveWindow.tsx
```diff
@@ -246,10 +225,22 @@ function ModalFooter() {
...
+
+    const isLive = Services.RestreamService.isLive.subscribe(isLive => {
+      if (isLive && !isStreamShiftPromptShown.current) {
+        promptStreamShift();
+      }
+    });
+
+    return () => isLive.unsubscribe();
+    // Note : `promptStreamShift` is intentionally left out of the dependency array to avoid multiple prompts on window load
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
wesrupert: Looks like these lines got added to the wrong place

wesrupert (approved):

PR #6118, opened 2026-08-26
Update go live checklist.

michelinewu:
# Live Output Editing Service Layer and Go Live Checklist

**Stack:** `master` ← mw_fix_ss ← mw_ms_toggle_1 … mw_ms_toggle_6 ← **mw_ms_toggle_7**

## Summary

The mid-stream target update path: adding and removing platforms and custom destinations on a running stream, and the checklist that reports it.

**`runUpdateStreamSettings`** is the new mid-stream update path in `updateStreamSettings`, taken only when live output editing is on. It diffs enabled against active targets into start/stop/continue, runs `beforeGoLive` for started platforms, updates settings for continuing ones, and adds and removes restream targets in single batched requests.

**Removal is deferred when nothing survives the update.** Removing every target first empties the restream session and the server ends it, so the following add fails against a stream that no longer exists. When `willRemoveTargets && willAddTargets && !keepsAtLeastOneTarget`, the add runs first. Every other update keeps the existing remove-then-add order.

**Failure reconciliation.** `syncTargetsToLive` and `restoreFailedTargets` correct the saved settings against what the server actually has. The Edit Stream window persists a toggle as soon as it is switched, so a failed update otherwise leaves the settings claiming targets that never started, or claiming a target was removed while it is still streaming. `getLiveTargets` in `RestreamService` is the read side, matching by stream key rather than platform because relayed platforms all report as `relay`.

**Checklist reports target changes.** `GoLiveChecklist` renders start, stop and continue rows for platforms and destinations separately in update mode.

**Files changed:** `app/services/streaming/streaming.ts`, `app/services/restream.ts`, `app/services/streaming/streaming-view.ts`, `app/components-react/windows/go-live/GoLiveChecklist.tsx`, `app/components-react/windows/go-live/useGoLiveSettings.ts`, `app/services/streaming/streaming-api.ts`

## Performance Implications

No change to the existing go live or update paths. For live output editing, adds and removes are batched into one request each rather than per target, and `getDisplaysToSetup` limits the full go live setup to displays that actually lack a session. The reconciliation for the failures (`syncTargetsToLive`, `getLiveTargets`) would make one extra `fetchTargets` request.


app/services/restream.ts
```diff
@@ -664,6 +693,86 @@ export class RestreamService extends StatefulService<IRestreamState> {
...
+   * TODO: This is an unverified assumption about the shape of the backend's stream keys.
+   * Replace it with `fetchUserSettings('portrait').streamKey` once the stream shift flow,
+   * which depends on the modeless key, can also fetch per-mode keys.
+   * @param streamKey - The landscape stream key for the restream session
+   * @param orientation - The orientation to resolve the key for
+   */
+  private async getModeStreamKey(
+    orientation: TOutputOrientation,
+    streamKey?: string,
+  ): Promise<string> {
```
wesrupert, line 710: Nit: Since a provided `streamKey` trumps the computation, prefer param order more similar to e.g. [`Map.getOrInsert`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/getOrInsert):

```suggestion
  private async getModeStreamKey(
    streamKey: string | undefined,
    orientation: TOutputOrientation,
  ): Promise<string> {
```

app/services/restream.ts
```diff
@@ -811,36 +920,19 @@ export class RestreamService extends StatefulService<IRestreamState> {
...
-
-        this.streamSettingsService.setSettings(
-          {
-            streamType: 'rtmp_custom',
-          },
-          display,
-        );
+      // Await the settings for every display. Otherwise `beforeGoLive` resolves before the
+      // stream settings have been written and `createStreaming` reads stale values.
+      await Promise.all(
```
wesrupert: Should probably be `allSettled` if the issue is data corruption, in case a fetch fails:

```suggestion
      await Promise.allSettled(
```

app/services/restream.ts
```diff
@@ -897,17 +989,27 @@ export class RestreamService extends StatefulService<IRestreamState> {
...
-    return this.streamInfo.enabledPlatforms.reduce((platforms, platform) => {
+    const targetPlatforms = updatedPlatforms ?? this.streamInfo.enabledPlatforms;
+
+    return targetPlatforms.reduce((platforms, platform) => {
       // Enhanced broacasting when multistreaming uses its own video context and stream
       // so skip setting up Twitch as a target here
       if (isEnhancedBroadcasting && platform === 'twitch') {
+        if (updatedPlatforms) {
+          // Enhanced broadcasting is disabled while live output editing is enabled, so reaching
+          // this while adding targets means no Twitch target will be created for the display.
```
wesrupert, line 1006: Is this something we should toast for the user to know, or will it get resolved elsewhere in the flow?

app/components-react/windows/go-live/GoLiveChecklist.tsx
```diff
@@ -28,10 +29,68 @@ export default function GoLiveChecklist(p: HTMLAttributes<unknown>) {
...
+    get stopTargets() {
+      return module.activePlatforms
+        ? difference(module.activePlatforms, module.enabledPlatforms)
+        : [];
+    },
+
+    get stopDestinations() {
+      return module.activeDestinations
+        ? difference(
+            module.activeDestinations.map(dest => `${dest.url}/${dest.streamKey}`),
```
wesrupert: This syntax is present throughout the diff now. If desired, you can enforce this in the types to make sure it's not missed anywhere:

```typescript
// Somewhere
export type TDestinationId = `${string}/${string}`;

// Here, and elsewhere
get stopDestinations(): TDestinationId[] {
  // ...
}
```

Then any spots that are passing around destination IDs will have compile-time checks that they're being generated properly.

app/services/streaming/streaming-view.ts
```diff
@@ -287,9 +293,12 @@ export class StreamInfoView<T extends Object> extends ViewHandler<T> {
 
   /**
    * Returns if the user can or should use the restream service
+   * @remark Order matters here when checking for which features are enabled. Stream shift mode and live output editing
+   * take precedence over dual output mode.
    */
   get isMultiplatformMode(): boolean {
     if (this.isStreamShiftMode) return true;
+    if (this.isLiveOutputEditingEnabled) return true;
```
wesrupert: Nit: Move remarks to the code that it matters for. When it's in the docstring, it can make consumers think it's relevant for how it's called, when this is an internal detail.

```suggestion
  */
get isMultiplatformMode(): boolean {
    // Order matters here when checking for which features are enabled.
    // Stream shift mode and live output editing take precedence over
    // dual output mode.
    if (this.isStreamShiftMode) return true;
    if (this.isLiveOutputEditingEnabled) return true;
```

app/services/streaming/streaming-view.ts
```diff
@@ -334,9 +343,27 @@ export class StreamInfoView<T extends Object> extends ViewHandler<T> {
...
+   * wherever the persisted setting is read outside of `settings`, so that a setting persisted
+   * while the flag was granted cannot keep switching on live output editing behavior after it is
+   * revoked.
+   */
+  private get savedLiveOutputEditing(): boolean {
+    if (!this.incrementalRolloutView.featureIsEnabled(EAvailableFeatures.liveOutputEditing)) {
+      return false;
+    }
+    return this.streamSettingsView.state.goLiveSettings?.liveOutputEditing ?? false;
+  }
```
wesrupert: SAA

```suggestion
  /**
   * The persisted live output editing setting, gated by the feature flag
   * @remark Use this wherever the persisted setting is read outside of
   * `settings`, so that a setting persisted while the flag was granted
   * cannot keep switching on live output editing behavior after it is
   * revoked.
   */
  private get savedLiveOutputEditing(): boolean {
    if (!this.incrementalRolloutView.featureIsEnabled(EAvailableFeatures.liveOutputEditing)) {
      return false;
    }
    
    // Read `goLiveSettings` from state directly instead of `this.settings`
    // to avoid the circular dependency:
    // settings → savedSettings → getSavedPlatformSettings → settings.
    return this.streamSettingsView.state.goLiveSettings?.liveOutputEditing ?? false;
```

app/services/streaming/streaming-view.ts
```diff
@@ -464,6 +492,59 @@ export class StreamInfoView<T extends Object> extends ViewHandler<T> {
     );
   }
 
+  /**
+   * Returns the passed in targets according to their assigned display
+   * @remark Currently unused, but could be used for a future refactor to unify logic for filtering
```
wesrupert, line 498: Is this going to change soon? Otherwise this should probably be removed; it will be a current maintenance burden, without adding value.

app/services/streaming/streaming-view.ts
```diff
@@ -495,6 +576,25 @@ export class StreamInfoView<T extends Object> extends ViewHandler<T> {
     );
   }
 
+  /**
+   * Validate the display when live output editing is enabled
+   * @remark Used to ensure a platform with the `both` display, used for dual streaming, uses the
+   * default display instead. Reads `savedLiveOutputEditing` instead of `isLiveOutputEditingEnabled`
+   * to avoid the circular dependency: settings → savedSettings → getSavedPlatformSettings → settings
```
wesrupert: Doesn't look like this reads any of these values:

```suggestion
```

app/services/restream.ts
michelinewu: There will be calls that only pass in an orientation, which is why I chose this order

app/services/streaming/streaming-view.ts
michelinewu: This is added to simplify restrictions for different features working/not working with each other. It's to have a way to identify dual output outside the state variable

app/services/restream.ts
michelinewu: I'll make a note to address this with error handling in the final PR

app/services/restream.ts
```diff
@@ -948,33 +1050,51 @@ export class RestreamService extends StatefulService<IRestreamState> {
...
   }
 
-  setupCustomDestinations() {
+  setupCustomDestinations(customDestinations?: ICustomStreamDestination[], display?: TDisplayType) {
     const isDualOutputMode = this.streamingService.views.isDualOutputMode;
-    const modesToRestream = this.streamInfo.displaysToRestream.map(display =>
-      this.getMode(display),
-    );
+    const modesToRestream = this.streamInfo.displaysToRestream.map(d => this.getMode(d));
+
```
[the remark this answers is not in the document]
michelinewu: Addressed in a later PR

app/components-react/windows/go-live/GoLiveChecklist.tsx
michelinewu: Done

wesrupert (approved):

PR #6140, opened 2026-08-27
Fix HackerOne -3045235 by disabling the hyperlink

sandboxcoder:
- fix https://app.asana.com/1/1083097041131/project/1209632301705219/task/1210265116004123

The installer runs elevated (nsis.perMachine = true -> RequestExecutionLevel
admin), and NSIS's license page turns URLs in the licence text into live
hyperlinks -- including the GPLv3 header's <https://fsf.org/>. NSIS opens them
with a plain ShellExecute, so the browser inherits the installer's Administrator
token.

Remove the link rather than the URL: the GPL text is not ours to edit
("changing it is not allowed").

DisableLicenseLinks now clears ENM_LINK from the RichEdit's event mask. That is
the actual kill switch -- NSIS opens URLs in response to the EN_LINK
notification, so a control that never sends it can never trigger the
ShellExecute. It also turns off EM_AUTOURLDETECT and strips CFE_LINK across the
document so the text stops looking clickable.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

Below is image, user can no longer click on the URL:
image

### DIFF
--- a/installer.nsh
+++ b/installer.nsh
@@ -40,6 +40,52 @@ Function un.ModifyUnWelcome
   ${NSD_Check} $cleanupCheckbox
 FunctionEnd
 
+; H1-3045235: the installer runs elevated (nsis.perMachine = true -> RequestExecutionLevel
+; admin), and NSIS's license page turns URLs in the licence text -- e.g. the GPLv3 header's
+; <https://fsf.org/> -- into live hyperlinks. NSIS opens them with a plain ShellExecute, so the
+; browser inherits the installer's Administrator token. Kill the link instead of the URL: the
+; GPL text is not ours to edit ("changing it is not allowed").
+;
+; Installer pass only. electron-builder runs makensis twice, and on the BUILD_UNINSTALLER pass
+; assistedInstaller.nsh skips every installer page (it is all inside !ifndef BUILD_UNINSTALLER)
+; while still inserting MUI_UNPAGE_INSTFILES -- which would swallow the define below and fail
+; with "Call must be used with function names starting with un.".
+!ifndef BUILD_UNINSTALLER
+Function DisableLicenseLinks
+  Push $0
+  Push $1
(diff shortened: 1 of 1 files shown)


michelinewu (approved):

installer.nsh
```diff
@@ -40,6 +40,52 @@ Function un.ModifyUnWelcome
...
+; Installer pass only. electron-builder runs makensis twice, and on the BUILD_UNINSTALLER pass
+; assistedInstaller.nsh skips every installer page (it is all inside !ifndef BUILD_UNINSTALLER)
+; while still inserting MUI_UNPAGE_INSTFILES -- which would swallow the define below and fail
+; with "Call must be used with function names starting with un.".
+!ifndef BUILD_UNINSTALLER
+Function DisableLicenseLinks
+  Push $0
+  Push $1
+  Push $2
+  ; The licence RichEdit (id 1000) is on the inner dialog (class #32770), a child of
```
michelinewu: Nit, typo

installer.nsh
```diff
@@ -1,5 +1,22 @@
 !include MUI2.nsh
 
+; Disable auto-URL detection on the license page RichEdit control so that
+; URLs in the license text are displayed as plain text, not clickable links.
+; MUI_PAGE_CUSTOMFUNCTION_SHOW is scoped to the license page only: it is
+; undefined after the license page insertmacro (inserted by electron-builder)
+; to prevent it leaking into subsequent page definitions.
```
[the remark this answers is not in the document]
sandboxcoder: Updated PR description thanks

michelinewu (approved):

installer.nsh
sandboxcoder: good catch, thanks. Fixed commit 8e76a414f9e3c16c750b8af6b284818383c4f050

PR #6141, opened 2026-08-28
Multistream Toggle Fixes and Error Handling

michelinewu:


app/components-react/shared/StreamShiftToggle.tsx
```diff
@@ -16,53 +16,45 @@ interface IStreamShiftToggle {
   disabled?: boolean;
 }
 
+/**
+ * @deprecated This checkbox is currently no longer used but kept for legacy purposes.
```
gettinToasty, line 20: what legacy purposes?

app/services/streaming/stream-error.ts
```diff
@@ -51,9 +51,87 @@ export const errorTypes = {
...
   },
+  RESTREAM_INVALID_CONFIG: {
+    get message() {
+      return $t(
+        'Multistream settings are invalid, please check your platforms and destinations and try again',
+      );
+    },
+    get action() {
+      return $t(
+        'confirm the user has Ultra and confirm the settings for enabled platforms and destinations',
```
gettinToasty, line 62: these are things displayed to support in the diag right? in that case i don't think we want them translated in case a user has desktop running in chinese or something and the message becomes unusable for support

should be done in a different PR though

test/regular/filters.ts
```diff
@@ -215,7 +215,7 @@ test('Adding and removing a Gain filter', async t => {
   t.pass();
 });
 
-test.skip('Adding and removing a Noise Suppression filter', async t => {
+test('Adding and removing a Noise Suppression filter', async t => {
```
gettinToasty, line 218: kinda curious about why we changed which test we're skipping here

gettinToasty (approved):

app/components-react/shared/StreamShiftToggle.tsx
michelinewu: It's only for the feature flag right now. The flag will hide the cards and show this toggle, but eventually should be removed. I think it's worth keeping around for a month or so in case the UI changes

app/services/streaming/stream-error.ts
michelinewu: Makes sense. I'll make another PR to take the `actions` out of translations, and add to our CR instructions to leave them untranslated

wesrupert: For core, I added a no-op `$ti()` (Translate Ignore) function that we used to mark strings as intentionally untranslated, so e.g. our current claude i18n skills wouldn't flag them. Probably worth doing here too.

test/regular/filters.ts
michelinewu: Left over from when I was debugging tests, will remove the change. This test is fixed by Richard already

app/services/streaming/streaming.ts
```diff
@@ -1651,7 +1746,7 @@ export class StreamingService
       const status = this.state.info.error?.status ?? 400;
 
       const streamError = createStreamError(
-        type,
+        resolvedType,
         { status, statusText: $t('Multistream Error') + messages.join('. '), platform },
```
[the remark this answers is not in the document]
wesrupert: +1

app/services/streaming/stream-error.ts
michelinewu: I would love this!

PR #6158, opened 2026-09-04
Fix dual output source z order

aleksandr-voitenko:
### Description

This fixes the following issue:

```
Steps:
- Add a new source in a single output mode
- Enable dual output

Expected results:
The latest source is present at the top of the sources list and rendered on top of other
sources for both vertical and horizontal canvases.

Actual results:
The latest source is present at the top of the sources list and rendered on top of other
sources for the horizontal canvas, but it is rendered behind other sources on the vertical canvas (see screenshot)
```

image

### Details

When a source was added in single-output mode and Dual Output was subsequently enabled, the source remained top-most on the horizontal canvas but could render behind older sources on the vertical canvas.

Collections saved with this incorrect vertical order also retained the problem after restarting Desktop.

### Even more details

Horizontal and vertical scene nodes share one scene-node list and are connected through `sceneNodeMaps`. Partner creation and single-to-dual conversion did not consistently preserve the horizontal z-order when constructing the vertical node block. Nested-folder reparenting could also reverse sibling order.

### Changes

- Preserve source z-order when converting a single-output collection to Dual Output.
- Keep newly created source, copied-source, and folder partners correctly ordered for both active and inactive scenes.
- Preserve nested-folder sibling order during conversion.
- Validate mapped dual-output scenes when a collection is loaded.
- Rebuild the vertical node order from the authoritative horizontal order for valid existing scenes.
- Require a complete one-to-one node map with matching displays, node types, sources, and mirrored folder relationships before repairing.
- Leave malformed or incomplete scene graphs untouched and log a warning instead of guessing.
- Avoid changing node mappings while repairing persisted order.

### DIFF
--- a/app/services/dual-output/node-order.ts
+++ b/app/services/dual-output/node-order.ts
@@ -76,55 +76,40 @@ export function getValidatedDualOutputNodeOrder(
   }
 
   const horizontalNodes = nodes.filter(node => node.display === 'horizontal');
-  const horizontalNodeIds = new Set(horizontalNodes.map(node => node.id));
-  const ancestorsByNodeId = new Map<string, Set<string>>();
+  // A valid flattened folder tree is a depth-first preorder. Each folder is
+  // pushed and popped at most once, keeping hierarchy validation linear.
+  const openFolderIds: string[] = [];
 
   for (const node of horizontalNodes) {
-    const ancestors = new Set<string>();
-    let parentId = node.parentId;
+    const parentId = node.parentId || '';
 
-    while (parentId) {
-      if (ancestors.has(parentId) || parentId === node.id) {
-        return { isValid: false, reason: 'the horizontal folder hierarchy contains a cycle' };
-      }
-
+    if (parentId) {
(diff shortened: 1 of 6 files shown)


app/services/dual-output/node-order.ts
```diff
@@ -0,0 +1,120 @@
+import type { Scene, TSceneNode } from 'services/scenes';
+
+export type TValidatedNodeOrder =
+  | { isValid: true; nodeIds: string[] }
+  | { isValid: false; reason: string };
+
+/**
+ * Group dual-output nodes by display without changing either display's z-order.
+ * Scene nodes are stored from top to bottom, so the filters below are stable.
+ */
```
michelinewu, line 11: Nit to add to comment that the horizontal nodes should always be ordered before the vertical ones

michelinewu (approved):

app/services/dual-output/node-order.ts
aleksandr-voitenko: Fixed.

PR #6163, opened 2026-09-08
Fixed saving non-common encoder settings

aleksandr-voitenko:
⚠️This PR depends on:
https://github.com/streamlabs/obs-studio-node/pull/1770 

### Motivation

Users could save encoder settings in the UI without those settings reaching the encoder. This caused recordings to ignore the selected keyframe interval and Advanced audio tracks to use 160 Kbps despite a configured bitrate of 320 Kbps. Streaming also lost settings such as encoder presets and profiles.

The issue originated during the Factory API migration. Settings continued to persist through the existing settings API, while Desktop passed only selected video properties to Factory-created encoders and hardcoded audio track settings. A subsequent recording fix restored bitrate and rate control but left other properties missing.

### Description

This change depends on [OSN PR #1770](https://github.com/streamlabs/obs-studio-node/pull/1770), which exposes saved encoder settings and supports replacing user settings on inactive encoders. Desktop supplies these settings explicitly, preserving native encoder defaults and existing output policies.

Registered audio tracks are updated in place so outputs retain valid references. Retained video encoders receive current settings before reuse.

Tests inspect native objects and use `ffprobe` to verify keyframe spacing and audio bitrates in local recordings.

### Changes

- Pass complete saved video encoder settings to streaming and recording encoders.
- Refresh retained inactive encoders before reuse, clearing disabled options and recreating idle output contexts when the selected encoder changes.
- Apply saved Advanced audio track bitrates and names, Simple streaming audio bitrate, and the selected recording audio encoder.
- Add integration tests covering settings propagation, retained outputs, shared recording/replay encoders, and encoded media.

### DIFF
--- a/test/regular/settings/encoder-settings.ts
+++ b/test/regular/settings/encoder-settings.ts
@@ -2,24 +2,24 @@ import { execFileSync } from 'child_process';
 import { ensureDir, readdir, writeFile } from 'fs-extra';
 import * as path from 'path';
 import { platform } from 'os';
-import type { ServicesManager } from '../../app/services-manager';
-import type { StreamingService } from '../../app/services/streaming/streaming';
-import { SettingsService } from '../../app/services/settings';
-import { VideoSettingsService } from '../../app/services/settings-v2/video';
-import { ScenesService } from '../../app/services/api/external-api/scenes';
-import { IAudioServiceApi } from '../../app/services/audio';
-import { NotificationsService } from '../../app/services/notifications';
-import { IObsListInput } from '../../app/components/obs/inputs/ObsInput';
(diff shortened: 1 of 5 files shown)


test/regular/settings/encoder-settings.ts
```diff
@@ -0,0 +1,467 @@
+import { execFileSync } from 'child_process';
```
michelinewu, line 1: Consider moving this file into the `regular/settings` directory

michelinewu (approved):

test/regular/settings/encoder-settings.ts
aleksandr-voitenko: Fixed.

PR #6171, opened 2026-09-09
Add test failure screenshot option.

michelinewu:
# Capture a Screenshot of the App When an E2E Test Fails

## Issues

Speed up debugging by recording what the app actually looked like at the moment of failure.

## Fixes

Added `saveFailureScreenshot()` to `test/helpers/webdriver/index.ts`, called from `afterEach.always` right before the existing teardown stops the app — the last point at which the failing UI still exists. Opt-in via `SLOBS_FAIL_SCREENSHOT_DIR`, so a normal test run captures nothing and is unaffected. Files are named after the test title (kebab-cased, truncated to 80 chars). The capture is wrapped in try/catch and never throws, so a failed screenshot can't mask or replace the real test failure.

**File(s) changed:** `test/helpers/webdriver/index.ts`

## Performance Implications

None on a normal test run — the function returns immediately unless `SLOBS_FAIL_SCREENSHOT_DIR` is set. When enabled, adds one screenshot capture per failing test, on the already-failing path.

### DIFF
--- a/test/helpers/webdriver/index.ts
+++ b/test/helpers/webdriver/index.ts
@@ -215,6 +215,35 @@ export async function debugPause() {
   await new Promise(() => {});
 }
 
+/**
+ * Screenshot the app as it stands when a test fails.
+ *
+ * @remarks Must opt-in by setting the screenshot directory with `SLOBS_FAIL_SCREENSHOT_DIR`, so a normal test run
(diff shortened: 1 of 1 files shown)


wesrupert (commented):
+1 on current suggestion, otherwise LGTM

gettinToasty (approved):

PR #6173, opened 2026-09-10
Fix source selection after disabling dual output

aleksandr-voitenko:
The PR fixes the following issue:
```
Steps:
- Enable dual output mode
- Disable Horizontal canvas in the top right corner settings
- Disable dual output mode

Expected: When sources selected on canvas, they highlighted in the list an vice versa.
Actual: This does not happen. Selection link is lost.

Note: it works again if enable dual output and enable there horizontal canvas. So it must be a pure UI issue.
```
Reproduction video:

https://github.com/user-attachments/assets/ad1a4a80-f7a0-40f4-b1b2-5e91cd94cd14

### Description

Disabling dual output after hiding the horizontal canvas left the Sources list using vertical item IDs while the canvas displayed horizontal items, breaking selection highlighting in both directions.

Use horizontal items consistently when dual output is off, and restrict paired selection and highlight mapping to dual-output mode. Preserve saved display visibility preferences when dual output is re-enabled.

Added regression tests covering selection in both directions after disabling dual output with horizontal hidden or visible, plus restoration of display preferences.

### DIFF
--- a/test/regular/api/dual-output.ts
+++ b/test/regular/api/dual-output.ts
@@ -3,10 +3,92 @@ import { getApiClient } from '../../helpers/api-client';
 import { test, useWebdriver, TExecutionContext } from '../../helpers/webdriver';
 import { ScenesService, Scene, SceneItem } from 'services/scenes';
 import { VideoSettingsService } from 'services/settings-v2/video';
+import { SelectionService } from 'services/api/external-api/selection';
+import { click, focusMain, focusWindow, waitForDisplayed } from '../../helpers/modules/core';
 
 // not a react hook
 // eslint-disable-next-line react-hooks/rules-of-hooks
 useWebdriver();
+
+async function setDualOutputMode(t: TExecutionContext, status: boolean) {
+  try {
+    t.true(await focusWindow('worker'), 'Worker window is available');
+    // Exercise the normal mode transition without requiring a provider login.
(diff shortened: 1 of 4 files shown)


test/regular/api/dual-output.ts
```diff
@@ -3,11 +3,84 @@ import { getApiClient } from '../../helpers/api-client';
...
+  const dualOutput = client.getResource<{
+    setDualOutputMode(status: boolean, skipShowVideoSettings: boolean): Promise<void>;
+  }>('DualOutputService');
+
+  // Exercise the normal mode transition without requiring a provider login.
+  await dualOutput.setDualOutputMode(status, true);
+  await focusMain();
+}
+
+for (const horizontalVisible of [false, true]) {
```
michelinewu: Could running the variants of this test be formatted more like the [goals widgets tests](https://github.com/streamlabs/desktop/blob/9ad16c718c810f7144f9b1fc2760ab50ce850e2a/test/regular/widgets/goals.ts#L9)?

michelinewu (changes requested):

test/regular/api/dual-output.ts
aleksandr-voitenko: Fixed

michelinewu (approved):

PR #6174, opened 2026-09-10
Highlighter -> Replay rename

jankalthoefer:


app/components-react/highlighter/migration/PageInstallationFlow.tsx
```diff
@@ -32,9 +37,12 @@ export default function PageInstallationFlow(props: IPageInstallationFlowProps)
   const isStaging = Utils.getHighlighterEnvironment() !== 'production';
 
   return (
-    <>
+    <div
+      className={cx(styles.appDetection, installedApp !== null && styles.appDetectionReady)}
```
michelinewu: Nit, for conventional styling we generally use `{ [styles.propName]: conditional }`

app/components-react/highlighter/migration/PageInstallationFlow.tsx
```diff
@@ -32,9 +37,12 @@ export default function PageInstallationFlow(props: IPageInstallationFlowProps)
...
+      aria-hidden={installedApp === null}
+    >
       <FeatureCarousel
-        title={REPLAY_APP_NAME}
+        title={appName}
         description={$t(
           'Turn your streams into TikTok and YouTube videos with the click of a button.',
           { appName: REPLAY_APP_NAME },
```
michelinewu: Should this be updated to `appName`?

app/components-react/highlighter/ImportStream.tsx
```diff
@@ -44,8 +49,8 @@ export function ImportStreamModal({
   } | null>(null);
 
   useEffect(() => {
-    HighlighterService.isStreamlabsReplayInstalled().then(installed => {
-      setReplayInstalled(installed);
+    HighlighterService.getInstalledHighlighterApp().then(app => {
```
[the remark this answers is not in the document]
michelinewu: Seconding this

michelinewu (approved):

app/components-react/highlighter/migration/PageInstallationFlow.tsx
jankalthoefer: Yeah, we now have the case that people who have the old highlighter installed should be brought into highlighter (where they migrate to replay) and then if they have replay installed we bring them into replay. 
With that the naming now is dynamic.

PR #6176, opened 2026-09-10
Fix LOE bugs.

michelinewu:
# Fix Live Output Editing Regressions in Dual Stream Display, Enhanced Broadcasting, and Title Persistence

## Summary

Two commits on `mw_fix_loe`. The first (`1027a1b34`) fixes how the `both`-display value used for Twitch dual streaming interacts with Live Output Editing (LOE) and stream shift. The second (`8282ea831`) fixes `prepopulate()` wiping an in-progress title/description edit every time the user toggles a platform in the Go Live window.

## Issues

**`getValidatedDisplay` coerced `both` to `horizontal` at the wrong layer.** 

**Enhanced broadcasting was decided after the stream key was already sent.** 

**`createEnhancedBroadcastMultistream` inferred the Twitch display from the wrong signal.** 

**Enhanced broadcasting stayed enabled during stream shift.** 

**Stream shift's Twitch sync read back the OBS runtime flag it had just forced off.** 

**Stream shift showed a numeric category ID instead of its name.** 

**`prepopulate()` discarded unsaved title/description edits on every platform toggle.** 

## Fixes

`getValidatedDisplay` is removed. `shouldSetupDualOutput` now defaults an unset display to `horizontal` inline and separately treats any enabled platform with `both` as sufficient to force dual output mode. `getSavedPlatformSettings` now passes `both` through unchanged (`savedDestinations[platform]?.display ?? 'horizontal'`), leaving LOE's incompatibility with dual streaming to be enforced at the point the display is actually used rather than at the point it's stored.

`TwitchService.beforeGoLive` now resolves the enhanced-broadcasting/dual-stream/LOE branch (via a new `IPlatformService.setupLiveOutputStream` hook, implemented on `TwitchService` as `setEnhancedBroadcasting(false)`) before the stream-key write, so the decision applies to the stream that's actually starting.

`createEnhancedBroadcastMultistream` now derives the display from `this.views.getPlatformDisplayType('twitch')` when dual output mode is active, falling back to `horizontal` otherwise — reading the actual platform-to-display assignment instead of guessing from the ingest URL.

`streaming.ts`'s enhanced-broadcasting eligibility check adds `!this.views.isStreamShiftMode` alongside the existing LOE check.

The stream-shift Twitch sync now reads `this.state.settings.isEnhancedBroadcasting` (the persisted user preference) instead of the OBS runtime flag, and resolves `settings.game_id` to a name via `this.fetchGame()` before assigning it to `game`/`gameName`, falling back to the existing channel info on fetch failure.

`GoLiveSettingsModule.prepopulate` takes an optional `{ preserveCommonFields?: boolean }`. When set, it snapshots `commonFields` before the rebuild and restores the title/description afterward (`editedCommonFields.title || this.state.commonFields.title`, same for description). `switchPlatforms` is the only caller that passes it — every other `prepopulate()` call site is a window open, where the freshly fetched values should win.

**Files changed:** `app/components-react/windows/go-live/useGoLiveSettings.ts`, `app/services/platforms/index.ts`, `app/services/platforms/twitch.ts`, `app/services/streaming/streaming-view.ts`, `app/services/streaming/streaming.ts`

## Performance Implications

None materially. Most of this is reordering existing logic and replacing one indirection (`getValidatedDisplay`) with an inline default — no new work on the hot path. The one addition is a `fetchGame()` HTTP call during stream-shift settings sync, and only when the shifted stream reports a `game_id`; that sync already makes a Twitch API call in the same `Promise.all`, so this adds one more request to an already network-bound, infrequent (shift-triggered) path, not to Go Live or steady-state streaming.

### DIFF
--- a/app/components-react/windows/go-live/useGoLiveSettings.ts
+++ b/app/components-react/windows/go-live/useGoLiveSettings.ts
@@ -336,9 +336,12 @@ export class GoLiveSettingsModule {
   /**
    * Fetch settings for each platform
    */
-  async prepopulate() {
+  async prepopulate(options?: { preserveCommonFields?: boolean }) {
     const { StreamingService, RestreamService, DualOutputService } = Services;
     const { isMultiplatformMode } = StreamingService.views;
+
+    // Snapshot the common fields `updateSettings` below replaces every platform's settings
+    const editedCommonFields = options?.preserveCommonFields ? this.state.commonFields : undefined;
 
     this.state.setNeedPrepopulate(true);
     await StreamingService.actions.return.prepopulateInfo();
@@ -358,16 +361,18 @@ export class GoLiveSettingsModule {
     };
 
     if (this.state.isUpdateMode && !view.isMidStreamMode) {
-      (Object.keys(settings.platforms) as (keyof typeof settings.platforms)[]).forEach(
-        (platform: TPlatform) => {
-          // In multi-platform mode, allow deleting all platform settings, including primary
-          if (!isMultiplatformMode && this.state.isPrimaryPlatform(platform)) {
-            return;
-          }
+      if (!view.isMidStreamMode) {
+        (Object.keys(settings.platforms) as (keyof typeof settings.platforms)[]).forEach(
+          (platform: TPlatform) => {
+            // In multi-platform mode, allow deleting all platform settings, including primary
+            if (!isMultiplatformMode && this.state.isPrimaryPlatform(platform)) {
+              return;
+            }
 
-          delete settings.platforms[platform];
-        },
-      );
+            delete settings.platforms[platform];
+          },
+        );
+      }
     }
 
     // prefill the form if `prepopulateOptions` provided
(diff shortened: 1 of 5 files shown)


app/components-react/windows/go-live/useGoLiveSettings.ts
```diff
@@ -358,16 +361,18 @@ export class GoLiveSettingsModule {
...
-        (platform: TPlatform) => {
-          // In multi-platform mode, allow deleting all platform settings, including primary
-          if (!isMultiplatformMode && this.state.isPrimaryPlatform(platform)) {
-            return;
-          }
-
-          delete settings.platforms[platform];
-        },
-      );
+      if (!view.isMidStreamMode) {
```
blackxored: Isn't this covered by the conditional above it?

blackxored (approved):
I can only spot check, there's a lot of display logic here I don't have the context for.

wesrupert (approved):

PR #6198, opened 2026-09-18
Fix stream shift disable enhanced broadcasting and clarity refactor.

michelinewu:
# Disable Enhanced Broadcasting for Stream Shift 

## Issues

Stream shift requires enhanced broadcasting to be disabled, which wasn't being set. The call sites were ambiguous so a small distinction between tracking the enhanced broadcasting user setting vs. the actual setting to go live could be confused.

## Fixes

Adds an explicit `else if (goLiveSettings.streamShift)` branch in `beforeGoLive` that forces enhanced broadcasting off, matching what `setupStreamShiftStream()` already does on the path that does short-circuit.

Renames the state-only method to `setEnhancedBroadcastingState` and introduces `setEnhancedBroadcastingSetting` as the named wrapper around the OBS write.

**Files changed:** `app/services/platforms/twitch.ts`, `app/components-react/windows/go-live/useGoLiveSettings.ts`

## Performance Implications

None.


app/services/platforms/twitch.ts
```diff
@@ -248,11 +248,14 @@ export class TwitchService
     if (channelInfo) {
       if (goLiveSettings?.liveOutputEditing) {
```
wesrupert: Nit: `!!channelInfo` *implies* `!!goLiveSettings`, but that's pretty tenuous IMO. Adding `goLiveSettings` to the first if would make it safer if the logic changes in the future here, since line 251 isn't checking nullish values.

```suggestion
    if (goLiveSettings && channelInfo) {
      if (goLiveSettings.liveOutputEditing) {
```

wesrupert (approved):

